Issue #63, opened 2019-09-03
Create perceptual representation for Mitch's "Mom put the ball on the box" scenario


rgabbard:

rgabbard:
Done

PR #64, opened 2019-09-03
Start of developmental primitive perceptual representation

rgabbard:
Addresses but does not complete #63

### DIFF
--- a/adam/perception/developmental_primitive_perception.py
+++ b/adam/perception/developmental_primitive_perception.py
@@ -20,7 +20,8 @@ class DevelopmentalPrimitivePerceptionFrame(PerceptualRepresentationFrame):
 
     - a set of `ObjectPerception`\ s, with one corresponding to each
       object in the scene (e.g. a ball, Mom, Dad, etc.)
-    - a set of `ObjectPerception`\ s which associate a `ObjectPerception` with perceived properties
+    - a set of `PropertyPerception`\ s which associate a `ObjectPerception` with perceived
+    properties
       of various sort (e.g. color, sentience, etc.)
     - a set of `RelationPerception`\ s which describe the learner's perception of how two
       `ObjectPerception`\ s are related.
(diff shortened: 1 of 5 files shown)


tests/perception/developmental_primitive_perception.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+
+def test_relations():
+    # ball on a table
+    ball = ObjectPerception("ball")
+    table = ObjectPerception("table")
+
+    PerceptualRepresentation.single_frame(
+        ObjectPerception(
```
denizbeser: I'm not sure if this test is correct - If I understand the structures correctly, ObjectPerception is for objects (not frames). Should the argument be DevelopmentalPrimitivePerceptionFrame(...) instead?

adam/perception/developmental_primitive_perception.py
```diff
@@ -0,0 +1,219 @@
...
+    r"""
+    a set of `ObjectPerception`\ s, with one corresponding to each
+      object in the scene (e.g. a ball, Mom, Dad, etc.)
+    """
+    property_assertions: ImmutableSet["PropertyPerception"] = attrib(
+        converter=_to_immutableset, default=immutableset()
+    )
+    r"""
+    a set of `ObjectPerception`\ s which associate a `ObjectPerception` with perceived properties
+      of various sort (e.g. color, sentience, etc.)
```
denizbeser: Should the documentation here be "a set of 'PropertyPerception'" instead?

tests/perception/developmental_primitive_perception.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_color():
+    # create a situation with a red ball
+    red = RgbColorPerception(255, 0, 0)
+    ball = ObjectPerception("ball")
+
+    PerceptualRepresentation.single_frame(
+        DevelopmentalPrimitivePerceptionFrame(
+            perceived_objects=[ball], property_assertions=[HasColor(ball, red)]
```
denizbeser: As PyCharm warns I believe the arguments are supposed to be ImmutableSets, so I think we need to convert the lists with ImmutableSet(). The same is true for the arguments in the above test function.

tests/perception/developmental_primitive_perception.py
rgabbard: Good catch. This is fixed in 92745a2

adam/perception/developmental_primitive_perception.py
rgabbard: Good catch, fixed in c345a2d

tests/perception/developmental_primitive_perception.py
rgabbard: This is a PyCharm bug - it does not properly handle the `converter` applies by `attrs`.

PR #71, opened 2019-09-04
Representation of object shape and hierarchy

rgabbard:


LichMaster98:
@rgabbard With the addition of the docstrings this is ready for review.

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
...
+        converter=_to_immutableset, default=immutableset()
+    )
+    """
+    A set of `SubObject` which form parts of the whole. SubObjects themselves are a hierarchy
+    """
+    sub_object_relations: ImmutableSet["SubObjectRelation"] = attrib(
+        converter=_to_immutableset, default=immutableset()
+    )
+    """
+    A set of `SubObjectRelation` which define how the `SubObject` relate to one another. These
```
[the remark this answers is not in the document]
rgabbard: Do make an issue, though, that we need to distinguish "is above from the point of view of an object's primary axis" from "is above [absolutely]" in order to handle e.g. a person lying down.

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
...
+    can move their arm either above or below their head.
+    
+    Does this need to be addressed?
+    """
+
+
+@attrs(frozen=True, slots=True, repr=False)
+class SubObject:
+    """
+    A `HierarchicalObjectSchema` which is a part of a whole.
```
rgabbard: Well, this isn't itself a `HierarchicalObjectSchema`; rather it *has* a `HierarchicalObjectSchema`. Perhaps better would be:

```
A sub-component of a generic type of object.

This is for use only in constructing `HierarchicalObjectSchema`\ ta.
```

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
...
+
+    SubObjects should not exist outside of `HierarchicalObjectSchema`
+    """
+
+    schema: HierarchicalObjectSchema = attrib()
+    """
+    SubObjects themselves can be a HierarchicalObject.
```
rgabbard: Perhaps better:

```
The `HierarchicalObjectSchema` describing the internal structure of this sub-component.

For example, an ARM is a sub-component of a PERSON, but ARM itself has a complex structure
(e.g. it includes a hand)
```

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
...
+    This class defines the relationships between `SubObject` of a `HierarchicalObjectSchema`
+    """
+
+    relation_type: OntologyNode = attrib(validator=instance_of(OntologyNode))
+    """
+    An `OntologyNode` which gives the relationship type between the args
+    """
+    arg1: SubObject = attrib()
+    """
+    A `SubObject` which is the parent of the relation_type
```
rgabbard: I would say "first argument" and "second argument", where the interpretation of the arguments is relation-type-specific. For example, there isn't any clear parent or child in a `contacts` relationship.

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
 
     def __repr__(self) -> str:
         return f"+{self._handle}"
+
+
+@attrs(frozen=True, slots=True, repr=False)
+class HierarchicalObjectSchema:
+    """
+    A representation of objects made of `SubObject` via `SubObjectRelation`
```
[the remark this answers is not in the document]
rgabbard: Also mention here that it specifying `HierarchicalObjectSchema`\ ta can be verbose,
and direct the reader to `sub_object_relations` for some recommendations on
how to make it easier.

adam/ontology/__init__.py
```diff
@@ -179,3 +190,81 @@ class OntologyProperty:
...
+    arg2: SubObject = attrib()
+    """
+    A `SubObject` which is the child of the relation_type
+    """
+
+
+# DSL to make writing object hierarchies easier
+
+
+def sub_object_relations(
```
rgabbard: Something like:

```
Convenience method to enable writing sub-object relations in a `HierarchicalObjectSchema` more easily.

This method simply flattens collections of items in the input iterable.

This is useful because it allows you to write methods 
for your relations which produce collections of relations
as their output. 
This allows you to use such DSL-like methods to enforce constraints
between the relations.

Please see the  `adam.ontology.phase1_ontology.PERSON_SCHEMA` for an example of how this is useful.
```

adam/ontology/phase1_ontology.py
```diff
@@ -66,6 +92,120 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
...
+"""
+A relation indicating that  one object provides the force to counteract gravity and prevent another 
+object from falling.
+"""
+subtype(SUPPORTS, SPATIAL_RELATION)
+
+
+def supports(obj1: SubObject, obj2: SubObject) -> Tuple[SubObjectRelation, ...]:
+    """
+    Define a support `SubObjectRelation` for `HierarchicalObjectSchema`
```
rgabbard:
```
Convenience method for indicating that one `SubObject` in a `HierarchicalObjectSchema` has a `SUPPORTS` relation with another.

For us with `sub_object_relations`.
```

Note the return is not accurate - it does not return a `SubObjectRelation` but a *tuple* wrapping a single `SubObjectRelation`. But I think if we point the user to `sub_object_relations` we can omit the documentation of the returns.

rgabbard: (please alter the others in a similar fashion)

adam/ontology/phase1_ontology.py
```diff
@@ -66,6 +92,120 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
...
+subtype(CONTACTS, SPATIAL_RELATION)
+
+
+def contacts(obj1: SubObject, obj2: SubObject) -> Tuple[SubObjectRelation, ...]:
+    """
+    DSL for constructing a 'contacts' spatial_relationship
+
+    Args:
+        obj1: The `SubObject` which has a reciprocal contacts relationship with
+        obj2: The `SubObject` which contacts obj1
```
rgabbard: When referencing arguments in docstrings, the convention is to surround them with asterisks

adam/perception/developmental_primitive_perception.py
```diff
@@ -83,68 +83,33 @@ class PropertyPerception(ABC):
...
 
 @attrs(slots=True, frozen=True, repr=False)
-class HasFlagProperty(PropertyPerception):
+class HasBinaryProperty(PropertyPerception):
     """
     A learner's perception that *perceived_object* possesses the given *flag_property*.
     """
 
-    flag_property = attrib(validator=instance_of(FlagProperty))
+    flag_property = attrib(validator=instance_of(OntologyProperty))
```
rgabbard: On further reflection, we should just call this `binary_property`

docs/conf.py
```diff
@@ -204,7 +204,10 @@
 
                   # pending https://github.com/isi-vista/immutablecollections/issues/60
                   ("py:class", "immutablecollections._immutableset.ImmutableSet"),
-                  ("py:class", "random.Random")]
+                  ("py:class", "random.Random"),
+                  ("py:class", "HierarchicalObjectSchema.SubObject"),
```
rgabbard: This line can be removed since this class no longer exists

docs/conf.py
```diff
@@ -204,7 +204,10 @@
...
+                  ("py:class", "HierarchicalObjectSchema"),
```
rgabbard: check whether this one is still necessary

PR #76, opened 2019-09-04
Class to compute differences between perception frames

denizbeser:
The class computes 6 different sets for added and removed objects, relations, and assertions. Would be great if you could check whether the class hierarchy is as needed.

### DIFF
--- a/adam/perception/perception_frame_difference.py
+++ b/adam/perception/perception_frame_difference.py
@@ -1,51 +1,50 @@
-from abc import ABC
-from typing import Generic, TypeVar
-
 from attr import attrib, attrs
 from attr.validators import in_, instance_of
-from immutablecollections import ImmutableSet
+from immutablecollections import ImmutableSet, immutableset
+from immutablecollections.converter_utils import _to_immutableset
+
 from adam.perception.developmental_primitive_perception import DevelopmentalPrimitivePerceptionFrame, ObjectPerception, \
     PropertyPerception, RelationPerception
 
 
-class PerceptionFrameDiff(ABC):
-    r"""
-    Represents the difference between a `LanguageLearner`\ 's perception
-    of two different moments.
-    """
-
-
 @attrs(slots=True, frozen=True)
-class DevelopmentalPrimitivePerceptionFrameDiff(PerceptionFrameDiff):
+class DevelopmentalPrimitivePerceptionFrameDiff:
     r"""
     Represents the difference between a `LanguageLearner`\ 's perception
     of two different moments as sets of added and removed relations, perceived_objects, and property assertions.
     """
 
-    added_objects: ImmutableSet["ObjectPerception"] = attrib(validator=instance_of(ImmutableSet))
+    added_objects: ImmutableSet["ObjectPerception"] = attrib(kw_only=True, converter=_to_immutableset,
+                                                             default=immutableset())
     r"""
     the set of `ObjectPerception`\ s, that were present on the second perception frame, but not the first.
     """
-    removed_objects: ImmutableSet["ObjectPerception"] = attrib(validator=instance_of(ImmutableSet))
+    removed_objects: ImmutableSet["ObjectPerception"] = attrib(kw_only=True, converter=_to_immutableset,
+                                                               default=immutableset())
     r"""
     the set of `ObjectPerception`\ s, that were present on the first perception frame, but not the second.
     """
-    added_property_assertions: ImmutableSet["PropertyPerception"] = attrib(validator=instance_of(ImmutableSet))
+    added_property_assertions: ImmutableSet["PropertyPerception"] = attrib(kw_only=True, converter=_to_immutableset,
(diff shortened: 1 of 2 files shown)


adam/perception/perception_frame_difference.py
```diff
@@ -0,0 +1,73 @@
...
+    Represents the difference between a `LanguageLearner`\ 's perception
+    of two different moments.
+    """
+
+
+PerceptionDiffT = TypeVar("PerceptionDiffT", bound="PerceptionFrameDiff")
+
+
+@attrs(slots=True, frozen=True)
+class DevelopmentalPrimitivePerceptionFrameDiff(PerceptionFrameDiff):
```
rgabbard: It's often useful to make the distinction between "classes which compute things" and "classes which represent data".  I'd suggest restructuring this so that `DevelopmentalPrimitivePerceptionFrameDiff` holds not the before frame and after frame, but rather six fields `add_objects`, `removed_objects`, `added_property_assertions`, `removed_property_assertions`, `added_relations`, and `removed_relations`.  Then have a method `def diff_primitive_perception_frames(before: DevelopmentalPrimitivePerceptionFrame, after: DevelopmentalPrimitivePerceptionFrame) -> DevelopmentalPrimitivePerceptionFrameDiff`. 

Contrary to my earlier guidance, I think we can hold off on generics for now.

tests/perception/developmental_primitive_perception_test.py
```diff
@@ -63,3 +64,31 @@ def test_relations():
             ],
         )
     )
+
+
+def test_difference():
```
rgabbard: 👍 for the test

adam/perception/perception_frame_difference.py
```diff
@@ -0,0 +1,72 @@
...
+from typing import Generic, TypeVar
+
+from attr import attrib, attrs
+from attr.validators import in_, instance_of
+from immutablecollections import ImmutableSet
+from adam.perception.developmental_primitive_perception import DevelopmentalPrimitivePerceptionFrame, ObjectPerception, \
+    PropertyPerception, RelationPerception
+
+
+class PerceptionFrameDiff(ABC):
```
rgabbard: I think we can delete this class for now and add it in the future if we decide it will provide value then.

adam/perception/perception_frame_difference.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@attrs(slots=True, frozen=True)
+class DevelopmentalPrimitivePerceptionFrameDiff(PerceptionFrameDiff):
+    r"""
+    Represents the difference between a `LanguageLearner`\ 's perception
+    of two different moments as sets of added and removed relations, perceived_objects, and property assertions.
+    """
+
+    added_objects: ImmutableSet["ObjectPerception"] = attrib(validator=instance_of(ImmutableSet))
```
rgabbard: Prefer"
```
 added_objects: ImmutableSet["ObjectPerception"] = attrib(converter=_to_immutableset, default=immutableset())
```

This allows the user to omit this parameter if empty and also to specify it via a list, tuple, or generator.

adam/perception/perception_frame_difference.py
```diff
@@ -0,0 +1,72 @@
...
+    r"""
+    the set of `ObjectPerception`\ s, that were present on the second perception frame, but not the first.
+    """
+    removed_objects: ImmutableSet["ObjectPerception"] = attrib(validator=instance_of(ImmutableSet))
```
rgabbard: Consider making all these fields `kw_only=True`, otherwise specifying them as arguments can be error prone

adam/perception/perception_frame_difference.py
```diff
@@ -0,0 +1,72 @@
...
+    added_property_assertions: ImmutableSet["PropertyPerception"] = \
+        after.property_assertions.difference(before.property_assertions)
+    removed_property_assertions: ImmutableSet["PropertyPerception"] = \
+        before.property_assertions.difference(after.property_assertions)
+    added_relations: ImmutableSet["RelationPerception"] = \
+        after.relations.difference(before.relations)
+    removed_relations: ImmutableSet["RelationPerception"] = \
+        before.relations.difference(after.relations)
+
+    return DevelopmentalPrimitivePerceptionFrameDiff(added_objects, removed_objects, added_property_assertions,
```
rgabbard: If you do switch to `kw_only=True`, note you can just do:


```python
return DevelopmentalPrimitivePerceptionFrameDiff(
    added_objects = after.perceived_objects.difference(before.perceived_objects),
    removed_objects = before.perceived_objects.difference(after.perceived_objects), 
....)

```

rgabbard:
@denizbeser : Looks like CI is still failing on some unused imports.

denizbeser:
@rgabbard Yes, but if I remove those imports on line 5 (ObjectPerception etc) then mypy fails because the imports are actually used on line 20 onwards.

PR #78, opened 2019-09-05
Introduce Hierarchy of Table to the Ontology

LichMaster98:


adam/ontology/phase1_ontology.py
```diff
@@ -67,11 +67,12 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
 
 # head, arm, torso, and leg are not part of the phase 1
 # vocabulary but are useful for describing the structure
-# of PERSON
+# of PERSON. Additional terms are for hierarchical objects
```
rgabbard: The addition to this comment is unclear to me.  Maybe we should rename all the objects we don't intend to be top-level with `_` prefixes and leave a general comment that these are only accessible as parts of other objects?

LichMaster98: I like that idea.

PR #84, opened 2019-09-06
Indicator for perceiveability of ontology properties

denizbeser:


adam/ontology/__init__.py
```diff
@@ -186,9 +186,14 @@ class OntologyProperty:
...
     """
+    perceivable: bool = attrib(default=False, kw_only=True)
+    """
+    A boolean indicator whether an ontological property is perceivable for the learner.
+    E.g. sentience and animacy, which are innately known. 
+    """
 
     def __repr__(self) -> str:
-        return f"+{self._handle}"
+        return f"+{self._handle},{self.perceivable}"
```
rgabbard: hmm - this is going to come out as something like `animate,true` which is both verbose and opaque.  Since this is a debug string for use by project developers, I think going for brief and opaque is okay. :-) . You could consider appending `%` to the handle for perceivable properties (it looks kind of like a pair of eyes....), like this: `animate%`.

Issue #100, opened 2019-09-11
HTML Visualizer for Curricula


rgabbard:
In order to give our curriculum to DARPA in an interpretable form, we'd like to be able to dump our learning examples to HTML documents.

Write a class which can take an `InstanceGroup` and render it to HTML.  Have a representation of the situation, the linguistic input, and the perceptual representation which is nicely human readable.  You can require that the situations be `HighLevelSemanticsSituations`, the linguistic forms be linearized dependency trees, and the perceptual forms be `PerceptualRepresentation[DevelopmentalPrimitivePerceptionFrame]`.

Start off with just simple situations containing a single object.

PR #115, opened 2019-09-11
Create HTML page from Curricula

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/100


tests/curriculum_to_html_test.py
```diff
@@ -0,0 +1,7 @@
+from adam.curriculum.phase1_curriculum import EACH_OBJECT_BY_ITSELF_SUB_CURRICULUM
+from adam.curriculum_to_html import CurriculumToHtml
+
+
+def test_simple_curriculum_html():
```
rgabbard: For the test, write your output to a temporary directory: http://doc.pytest.org/en/latest/tmpdir.html

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+
+from attr import attrs
+
+from adam.experiment import InstanceGroup
+from adam.language.dependency import LinearizedDependencyTree
+from adam.perception import PerceptualRepresentation
+from adam.perception.developmental_primitive_perception import (
+    DevelopmentalPrimitivePerceptionFrame,
+)
+from adam.situation.high_level_semantics_situation import HighLevelSemanticsSituation
```
rgabbard: Add a `__main__` to this class which applies it to the `GAILA_PHASE_1_CURRICULUM`. It should take one argument, a parameter file (using [`Parameters`](https://github.com/isi-vista/vistautils/blob/master/vistautils/parameters.py) from `vistautils`) which specified the output directory.

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@attrs(frozen=True, slots=True)
+class CurriculumToHtml:
+    """
+    Class to turn an `InstanceGroup` into an html document
+    """
+
+    @staticmethod
+    def generate(
```
rgabbard: `dump_to_html` is a clearer name than `generate`

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        instances: List[
+            InstanceGroup[
+                HighLevelSemanticsSituation,
+                LinearizedDependencyTree,
+                DevelopmentalPrimitivePerceptionFrame,
+            ]
+        ],
+        outputdestination: str,
```
rgabbard: Use `Path` rather than str for filesystem paths

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        title: str = "Instance Group",
```
rgabbard: I recommend making these kw-only

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+from adam.situation.high_level_semantics_situation import HighLevelSemanticsSituation
+
+
+@attrs(frozen=True, slots=True)
+class CurriculumToHtml:
+    """
+    Class to turn an `InstanceGroup` into an html document
+    """
+
+    @staticmethod
```
rgabbard: Why static?  Maybe rename the class to `CurriculumToHtmlDumper` and have the user instantiate one when they want to use it?

(the advantage of this is that we may want to add configuration options to the dumper in the future)

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+    def generate(
+        instances: List[
```
rgabbard: Prefer `Sequence` over `List` as an argument type - that way the use has more flexibility. Even better here is just plain `Iterable`.

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        outputdestination: str,
+        title: str = "Instance Group",
+        overwrite: bool = False,
+    ):
+        r"""
+        Static method to take a list of `InstanceGroup`\ s and turns each one into an indivual page
+        
+        Given a list of InstanceGroups and an output directory of *outputdestination* along with 
+        a *title* for the pages the generator loops through each group and calls the internal 
+        method to create HTML pages. *overwrite* indicates is previously existing HTML files should
```
rgabbard: In my experience, for things which are inexpensive to reproduce, it is less error prone to just nuke the output directory's current contents.

adam/curriculum_to_html.py
rgabbard: Also it looks like this should be called `instance_groups` instead of `instances`?

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        be overwritten or not
+        """
+        for x in range(len(instances)):  # pylint:disable=consider-using-enumerate
```
rgabbard:
```python
for (idx, instance_group) in enumerate(instance_groups):
```

let's you relax your input requirements to `Iterable`

adam/curriculum_to_html.py
rgabbard: Python style is to separate words in variable names with underscores.

adam/curriculum_to_html.py
rgabbard: Pylint was right. :-P

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+            CurriculumToHtml._generate(
+                instance=instances[x],
+                outputdestination=f"{outputdestination}{title}{x}.html",
+                title=title + f" {x}",
```
rgabbard: Why not `f"{title} {x}"`?

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+                overwrite=overwrite,
+            )
+
+    @staticmethod
+    def _generate(
```
rgabbard: `_dump_instance_group` would be clearer

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Internal generation method for individual instance groups into HTML pages
+
+        Given an `InstanceGroup` with a `HighLevelSemanticsSituation`,
+        `LinearizedDependencyTree`, and `DevelopmentalPrimitivePerceptionFrame` this function
+        creates an html page at the given *outputdestination* and *title*. If the file already
+        exists and *overwrite* is set to False an error is raised in execution. Each page turns an
+        instance group with each "instance" as an indiviudal section on the page.
+
+        No returns
```
rgabbard: No need to specify this - the type signature will put it into the docs automatically

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+
+        """
+        if os.path.isfile(outputdestination) and not overwrite:
+            raise RuntimeError(f"Not able to create new HTML file in {outputdestination}")
+        html = open(outputdestination, "w")
```
rgabbard: Use `with`:

```python
with open(output_destination, "w") as html_out:
   # do stuff, automatically closed on block exit
```

rgabbard: (this will ensure file closure even if the block is exited by an exception)

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        html.write(f"<h1>{title} - {instance.name()}</h1>\n")
+        instance_number = 0
+        for inst in instance.instances():
```
rgabbard: Use `enumerate`

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+            if not isinstance(inst[0], HighLevelSemanticsSituation):
+                raise RuntimeError(
+                    f"Expected the Situation to be HighLevelSemanticsSituation got {type(inst[0])}"
+                )
+            if not isinstance(inst[1], LinearizedDependencyTree):
+                raise RuntimeError(
+                    f"Expected the Lingustics to be LinearizedDependencyTree got {type(inst[1])}"
+                )
+            if not isinstance(inst[2], PerceptualRepresentation) and not isinstance(
```
rgabbard: I think you want `or` here (or else factor out the `not` over the whole expression).

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+                inst[2].frames, DevelopmentalPrimitivePerceptionFrame
```
rgabbard: `inst[2].frames[0]`, I think?

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+            ):
+                raise RuntimeError(
+                    f"Expected the Perceptual Representation to contain DevelopmentalPrimitivePerceptionFrame got {type(inst[2].frames)}"
+                )
+            html.write(
+                f'<table>\n<thead><tr><th colspan="3"><h2>Scene {instance_number}</h2>\n</th></tr></thead>'
+                f'<tbody><tr><td><h3 id="situation{instance_number}">Situation Description</h3></td>'
+                f'<td><h3 id="lingustics{instance_number}">Linguistic Descrption</h3></td>'
```
rgabbard: separating the instance number from `linguistics` by a dash will make the HTML more readable

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+                    f"Expected the Lingustics to be LinearizedDependencyTree got {type(inst[1])}"
+                )
+            if not isinstance(inst[2], PerceptualRepresentation) and not isinstance(
+                inst[2].frames, DevelopmentalPrimitivePerceptionFrame
+            ):
+                raise RuntimeError(
+                    f"Expected the Perceptual Representation to contain DevelopmentalPrimitivePerceptionFrame got {type(inst[2].frames)}"
+                )
+            html.write(
+                f'<table>\n<thead><tr><th colspan="3"><h2>Scene {instance_number}</h2>\n</th></tr></thead>'
```
rgabbard: Using newlines and tabs in your generated HTML makes it easier to debug problems with it

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+            html.writelines(CurriculumToHtml._situationtext(inst[0]))
+            html.writelines(CurriculumToHtml._linguistictext(inst[1]))
+            html.writelines(CurriculumToHtml._perceptiontext(inst[2]))
+            html.write("</tr></tbody>")
+            instance_number = instance_number + 1
+        html.close()
+
+    @staticmethod
+    def _situationtext(situation: HighLevelSemanticsSituation):
```
rgabbard: use `_` to separate words

rgabbard: Return type on this method and its siblings

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Converts a situation description into its sub-parts as a table entry
+
+        Receiving a `HighLevelSemanticsSituation` the objects, actions, and relations are displayed
```
rgabbard: The second paragraph can be removed, I think

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        outputtext = [f"<td>\n", "<h4>Objects</h4>\n<ul>"]
+        for obj in situation.objects:
+            outputtext.append(f"<li>{obj.ontology_node.handle}</li>")
+            # Reintroduce Properties as [] around the object
+            # outputtext.append(f"<li>Properties:\n<ul>")
+            # for prop in obj.properties:
+            #    outputtext.append(f"<li>{prop.handle}</li>")
+            # outputtext.append("</ul></li>")
+        outputtext.append("</ul>")
+        if not situation.actions:
```
rgabbard: Shouldn't this be `if situation.actions`?

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+            outputtext.append("<h4>Actions</h4>\n<ul>")
+            for acts in situation.actions:
+                outputtext.append(f"<li>{acts.action_type.handle}</li>")
+            outputtext.append("</ul>")
+        if not situation.relations:
```
rgabbard: ditto on the negation here

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+        return outputtext
+
+    @staticmethod
+    def _perceptiontext(
+        perception: PerceptualRepresentation[DevelopmentalPrimitivePerceptionFrame]
+    ):
+        """
+        Turns a perception into a list of items in the perceptions frames.
+
+        Given a `PerceptualRepresentation[DevelopmentalPrimitivePerceptionFrame]` the information
```
rgabbard: The summary line is good enough here as well, I think.

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,184 @@
...
+    @staticmethod
+    def _linguistictext(linguistic: LinearizedDependencyTree):
+        """
+        Parses the Linguistic Description of a Linearized Dependency Tree into a table entry
+
+        Takes a `LinearizedDependencyTree` which is turned into a token sequence and
+        phrased as a sentence for display. Returns a List[str]
+        """
+        outputtext = [f"<td>", " ".join(linguistic.as_token_sequence()), "</td>"]
+        return outputtext
```
rgabbard: In this case giving the variable name `outputtext` doesn't help the reader any, so you can just

```python
return [f"<td>", " ".join(linguistic.as_token_sequence()), "</td>"]
```

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        Parses the Linguistic Description of a Linearized Dependency Tree into a table entry
+
+        Takes a `LinearizedDependencyTree` which is turned into a token sequence and
+        phrased as a sentence for display. Returns a List[str]
+        """
+        return [f"\t\t<td>", " ".join(linguistic.as_token_sequence()), "\t\t</td>"]
+
+
+def main(params: Parameters):
```
rgabbard: You'll also need to call this from a

```
if __name__ == "__main__":
```

block for Python to run it.  You may want to use `parameters_only_entry_point` from vistautils.

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,177 @@
...
+        method to create HTML pages.
+        """
+        for (idx, instance_group) in enumerate(instance_groups):
+            CurriculumToHtmlDumper._dump_instance_group(
+                instance=instance_group,
+                output_destination=output_destination.joinpath(f"{title}{idx}.html"),
+                title=f"{title} {idx}",
+            )
+
+    @staticmethod
```
rgabbard: Un-static-ify all the methods. Apart from static factory methods, use static methods sparingly.

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,177 @@
...
+    def _dump_instance_group(
+        instance: InstanceGroup[
```
rgabbard: `instance_group` rather than `instance`?

adam/curriculum_to_html.py
```diff
@@ -0,0 +1,188 @@
...
+    curriculum_dumper = CurriculumToHtmlDumper()
+    curriculum_dumper.dump_to_html(
+        GAILA_PHASE_1_CURRICULUM,
+        output_destination=params.creatable_directory("output_directory"),
+        title="GAILA PHASE 1 CURRICULUM",
+    )
+
+
+USAGE_MESSAGE = (
+    "This main function only expects an output_directory location. Please provide it."
```
rgabbard:
```suggestion
    """
    curriculum_to_html.py param_file
     \twhere param_file has the following parameter:
     \t\toutput_directory: where to write the HTML output
   """
```

PR #119, opened 2019-09-12
add plural forms to lexicon

elizlee:


adam/language/lexicon.py
```diff
@@ -27,6 +27,7 @@ class LexiconEntry:
     properties: ImmutableSet["LexiconProperty"] = attrib(
         converter=_to_immutableset, default=immutableset()
     )
+    plural_form: str = attrib(validator=instance_of(str), default="", kw_only=True)
```
rgabbard: Prefer to make the potential absence of something explicit rather than using special values.

In this case, make the type of this `Optional[str]` with a default of `None`. While you are at it, change `verb_form_3SG_PRS` to be the same.

adam/language_specific/english/english_language_generator.py
```diff
@@ -96,8 +100,25 @@ class _Generation:
         ] = dict()
 
         def generate(self) -> ImmutableSet[LinearizedDependencyTree]:
-            for _object in self.situation.objects:
-                self._translate_object_to_noun(_object)
+            # For now, only apply quantifiers to object-only situations
+            if self.situation.actions == ImmutableSet.empty():
```
rgabbard: In Python, empty collections evaluate to `False`, so prefer:

```
if not self.situation.actions:
 ...
else:
  ...
```

You could also consider reversing the branches to eliminate the negation (negation tends to confuse readers)

adam/language_specific/english/english_language_generator.py
```diff
@@ -128,9 +149,14 @@ def _translate_object_to_noun(
             lexicon_entry = self._unique_lexicon_entry(
                 _object.ontology_node  # pylint:disable=protected-access
             )
-            dependency_node = DependencyTreeToken(
-                lexicon_entry.base_form, lexicon_entry.part_of_speech
-            )
+            if count > 1:
```
rgabbard: Note that a lot of the code on the two branches is shared.  Better would be:

```python
if count > 1:
   word_form = lexicon_entry.plural_form
else:
   word_form = lexicon_entry.base_form

dependency_node = DependencyTreeToken(
                    word_form, lexicon_entry.part_of_speech
                )
```

adam/language_specific/english/english_language_generator.py
```diff
@@ -142,10 +168,22 @@ def _translate_object_to_noun(
             if (dependency_node.part_of_speech != PROPER_NOUN) and (
                 MASS_NOUN not in lexicon_entry.properties
             ):
-                determiner_node = DependencyTreeToken("a", DETERMINER)
-                self.dependency_graph.add_edge(
-                    determiner_node, dependency_node, role=DETERMINER_ROLE
-                )
+                if count == 1:
+                    determiner_node = DependencyTreeToken("a", DETERMINER)
+                    self.dependency_graph.add_edge(
```
rgabbard: Note this part is the same for all branches, so you can pull the same trick as above, calling it `determiner_node` or something.

adam/language_specific/english/english_language_generator.py
```diff
@@ -142,10 +168,22 @@ def _translate_object_to_noun(
...
+                        determiner_node, dependency_node, role=DETERMINER_ROLE
+                    )
+                elif count == 2:
+                    numeral_node = DependencyTreeToken("two", NUMERAL)
+                    self.dependency_graph.add_edge(
+                        numeral_node, dependency_node, role=NUMERIC_MODIFIER
+                    )
+                # Currently, any number of objects greater than two is considered "many"
+                else:
+                    adjective_node = DependencyTreeToken("many", ADJECTIVE)
```
rgabbard: Does universal dependencies analyze this as `ADJECTIVE` or `DETERMINER`?

tests/language_specific/english/test_english_language_generator.py
```diff
@@ -70,6 +71,29 @@ def test_simple_verb():
     ).as_token_sequence() == ("Mom", "pushes", "a", "table")
 
 
+def test_simple_quantifier():
```
rgabbard: Also add a test with just one box. Also maybe rename this test to `test_two_objects` since it isn't clear to the reader what "simple" means in this context"

adam/language_specific/english/english_language_generator.py
```diff
@@ -116,7 +137,7 @@ def generate(self) -> ImmutableSet[LinearizedDependencyTree]:
             )
 
         def _translate_object_to_noun(
-            self, _object: SituationObject
+            self, _object: SituationObject, count: int = 1
```
rgabbard: It is not necessary on this pull request, but make an issue (and reference it from a comment in the code) to eventually enforce that when using plurals we need to ensure that all relevant relations and actions involving any of the pluralized things involve them all (e.g. we can't say _The two boxes on the table_ if one box is under the table)

adam/language_specific/english/english_language_generator.py
```diff
@@ -96,8 +100,25 @@ class _Generation:
         ] = dict()
 
         def generate(self) -> ImmutableSet[LinearizedDependencyTree]:
-            for _object in self.situation.objects:
-                self._translate_object_to_noun(_object)
+            # For now, only apply quantifiers to object-only situations
+            if self.situation.actions == ImmutableSet.empty():
+                # Get number of objects of each type
+                node_counts: Dict[OntologyNode, int] = dict()
```
rgabbard: Try `Counter` here: https://docs.python.org/2/library/collections.html#collections.Counter

adam/language_specific/english/english_language_generator.py
```diff
@@ -96,8 +100,25 @@ class _Generation:
...
+                for _object in self.situation.objects:
+                    try:
+                        node_counts[_object.ontology_node] += 1
+                    except KeyError:
+                        node_counts.update({_object.ontology_node: 1})
+                object_counts: Dict[SituationObject, int] = {
```
rgabbard: I think you can skip `object_counts` and just look below at the count of the node's `ontology_node`

adam/language/lexicon.py
elizlee: Done

adam/language_specific/english/english_language_generator.py
elizlee: Done

adam/language_specific/english/english_language_generator.py
elizlee: Done

adam/language_specific/english/english_language_generator.py
elizlee: Done

adam/language_specific/english/english_language_generator.py
elizlee: By definition it seems like it could be a determiner, but CoreNLP tags it as an adjective. Is there a better way to confirm this?

tests/language_specific/english/test_english_language_generator.py
elizlee: I thought adding a test with one box would have been repetitive since we already have `test_common_noun`.

adam/language_specific/english/english_language_generator.py
elizlee: Described issue in a comment

adam/language_specific/english/english_language_generator.py
elizlee: I totally forgot this existed...

elizlee: Done

adam/language_specific/english/english_language_generator.py
elizlee: Done

adam/language_specific/english/english_language_generator.py
rgabbard: Thanks - can you make a match issue on GitHub and include its URL in the comment also?

adam/language_specific/english/english_language_generator.py
```diff
@@ -156,7 +191,7 @@ def _translate_action_to_verb(
             # TODO: we don't currently handle verbal morphology.
             # https://github.com/isi-vista/adam/issues/60
             verb_dependency_node = DependencyTreeToken(
-                lexicon_entry.verb_form_3SG_PRS, lexicon_entry.part_of_speech
+                str(lexicon_entry.verb_form_3SG_PRS), lexicon_entry.part_of_speech
```
rgabbard: Do you need `str` here because it was complaining about it being `Optional[str]`? What you want to do instea dis to explicitly check it for being `None` and throw an exception if it is

adam/language_specific/english/english_language_generator.py
elizlee: Changed POS to `DETERMINER`

adam/language_specific/english/english_language_generator.py
elizlee: Done

PR #138, opened 2019-09-13
Bind action variables which are not bound by semantic roles

### LINKED ISSUE #122: Finding manipulator in action perception

denizbeser:
Closes #122

### DIFF
--- a/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
+++ b/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
@@ -21,6 +21,8 @@ from adam.ontology.phase1_ontology import (
     IS_SPEAKER,
     PART_OF,
     PERCEIVABLE,
+    AGENT,
+    CAN_MANIPULATE_OBJECTS,
 )
 from adam.perception import PerceptualRepresentation, PerceptualRepresentationGenerator
 from adam.perception.developmental_primitive_perception import (
@@ -251,21 +253,38 @@ class _PerceptionGeneration:
                     raise RuntimeError(
                         "Don't yet handle multiple fillers for the same semantic role"
                     )
-            first_relation_slot_filler_role_in_action = only(
-                entities_to_roles[condition.first_slot]
-            )
-            situation_object_1 = only(
-                action_roles_to_fillers[first_relation_slot_filler_role_in_action]
-            )
-            second_relation_slot_filler_role_in_action = only(
-                entities_to_roles[condition.second_slot]
-            )
-            situation_object_2 = only(
-                action_roles_to_fillers[second_relation_slot_filler_role_in_action]
-            )
 
-            perception_1 = self._objects_to_perceptions[situation_object_1]
-            perception_2 = self._objects_to_perceptions[situation_object_2]
+            # Check for manipulator in first slot
+            if CAN_MANIPULATE_OBJECTS in condition.first_slot.properties:
+                manipulator_ontology_node = self._get_manipulator_of_agent(
+                    only(action_roles_to_fillers[AGENT])
+                )
+                perception_1 = ObjectPerception(manipulator_ontology_node.handle)
+            else:
+                first_relation_slot_filler_role_in_action = only(
+                    entities_to_roles[condition.first_slot]
+                )
+
+                situation_object_1 = only(
+                    action_roles_to_fillers[first_relation_slot_filler_role_in_action]
+                )
+                perception_1 = self._objects_to_perceptions[situation_object_1]
+
+            # Check for manipulator in first slot
+            if CAN_MANIPULATE_OBJECTS in condition.second_slot.properties:
+                manipulator_ontology_node = self._get_manipulator_of_agent(
+                    only(action_roles_to_fillers[AGENT])
+                )
+                perception_2 = ObjectPerception(manipulator_ontology_node.handle)
+            else:
+                second_relation_slot_filler_role_in_action = only(
+                    entities_to_roles[condition.second_slot]
+                )
+                situation_object_2 = only(
+                    action_roles_to_fillers[second_relation_slot_filler_role_in_action]
+                )
+                perception_2 = self._objects_to_perceptions[situation_object_2]
+
             relation_perception = RelationPerception(
                 relation_type=condition.relation_type,
                 arg1=perception_1,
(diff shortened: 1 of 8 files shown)


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -251,21 +253,38 @@ def _get_relations_from_condition(
...
-                entities_to_roles[condition.second_slot]
-            )
-            situation_object_2 = only(
-                action_roles_to_fillers[second_relation_slot_filler_role_in_action]
-            )
 
-            perception_1 = self._objects_to_perceptions[situation_object_1]
-            perception_2 = self._objects_to_perceptions[situation_object_2]
+            # Check for manipulator in first slot
+            if CAN_MANIPULATE_OBJECTS in condition.first_slot.properties:
```
rgabbard: I think we want to check first whether the slot corresponds to one of the known argument role fillers. If it does, we should use it. We should only go searching for the manipulator if the binding isn't given explicitly.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +297,28 @@ def _get_relations_from_condition(
 
         return relations
 
+    def _get_manipulator_of_agent(self, agent_object: SituationObject) -> OntologyNode:
```
rgabbard: I think we want something a bit more general here.  For example, if we fail to find the explicit binding for a slot filler to an argument role, we want to then search over all unbound objects in the scene to find one with the required properties.

rgabbard: I think we already add sub-objects to the scene before we deal with relations, right?

denizbeser: Didn't realize that, working on it now. How should we handle cases where there are multiple objects that fit the property? For example, for "give", we want to associate the manipulator in the precondition with the agent and manipulator in the postcondition with the goal.

rgabbard: For now, throw an exception if we can't unambiguously resolve a variable to an object.

I am not sure I understand "manipulator in the postcondition with the goal." Can you explain?

denizbeser: That works. For example A person can give a ball to another person. In the post condition, the second person would hold the ball with their manipulator, while there are multiple manipulators in the situation.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
 
         return relations
 
+    def _find_perception_object_for_situation_object(
+        self, slot: SituationObject, entities_to_roles, action_roles_to_fillers
```
rgabbard: Missing type annotations

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+    ) -> ObjectPerception:
+        if slot in entities_to_roles.keys():
```
rgabbard:
```suggestion
        if slot in entities_to_roles:
```

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+            filler_role_in_action = only(entities_to_roles[slot])
+            situation_object = only(action_roles_to_fillers[filler_role_in_action])
+            perception = self._objects_to_perceptions[situation_object]
+        else:
+            found_objects = []
```
rgabbard: perhaps `objects_matching_constraints`.
Also add a comment above the `if` that if we already know what the situation object is, we just look it up. Otherwise, we need to search through the other situation objects to see if any matches the constraints we know to hold for the situation object.
Note also we don't currently handle constraints due to relations and link to the relevant issue

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+            for (
+                object_perception,
+                ontology_node,
+            ) in self._object_perceptions_to_ontology_nodes.items():
+                # check all the sub objects in the situation
+                if all(
+                    necessary_property in ontology_node.inheritable_properties
```
rgabbard: check for properties using the methods on the `Ontology` class; they will handle inheritance for you.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+                                self._object_perceptions_to_ontology_nodes[
+                                    obj
+                                ].inheritable_properties
+                                for obj in found_objects
+                            ]
+                        )
+                    )
+                    == 1
+                ):
+                    # if the found objects have identical properties
```
rgabbard: cite your example of hands in the comment here

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+                    return found_objects[0]
```
rgabbard: Sometimes you `return`, sometimes you assign to `perception`. I prefer to just `return`

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+            if not found_objects:
+                raise RuntimeError(f"Can not find object with properties {slot}")
+            elif len(found_objects) > 1:
+                if (
+                    len(
+                        set(
+                            [
+                                self._object_perceptions_to_ontology_nodes[
+                                    obj
+                                ].inheritable_properties
```
rgabbard: get properties via `Ontology`; this is not getting inherited properties

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -278,6 +282,49 @@ def _get_relations_from_condition(
...
+                    for necessary_property in slot.properties
+                ):
+                    found_objects.append(object_perception)
+            if not found_objects:
+                raise RuntimeError(f"Can not find object with properties {slot}")
+            elif len(found_objects) > 1:
+                if (
+                    len(
+                        set(
+                            [
```
rgabbard: Don't need wrapping list because `set` can take a generator expression directly

PR #145, opened 2019-09-16
Add regions for the representation of spatial relations

rgabbard:


adam/ontology/phase1_ontology.py
```diff
@@ -51,8 +54,8 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
 subtype(PATIENT, SEMANTIC_ROLE)
 THEME = OntologyNode("theme")
 subtype(THEME, SEMANTIC_ROLE)
-DESTINATION = OntologyNode("destination")
-subtype(DESTINATION, SEMANTIC_ROLE)
+GOAL = OntologyNode("destination")
```
elizlee: Should "destination" here be "goal"?

rgabbard: Nice catch!

Issue #109, opened 2019-09-11
Annotate ontology objects for their size relative to the learner


rgabbard:

LichMaster98:
@rgabbard, if `OntologyNodeRelations` are defined as having a `relation_type` and `agr1` is it fine to assume for "bigger_than" that the representation is `BIGGER_THAN(self, arg1)` or should an `OntologyNodeRelation` store `relation_type`, `arg1`, and `arg2` like the SubObjectRelations do? (Where the result would be `BIGGER_THAN(arg1, arg2)`)

rgabbard:
@LichMaster98 : the latter

PR #148, opened 2019-09-17
Introduce OntologyNode relative size to learner

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/109


adam/ontology/ontology.py
```diff
@@ -61,6 +62,9 @@ class Ontology:
     action_to_description: ImmutableDict[OntologyNode, ActionDescription] = attrib(
         converter=_to_immutabledict, default=immutabledict(), kw_only=True
     )
+    _node_to_relations: ImmutableDict[
```
rgabbard: Would it be easier just to take a set of relations and then do the indexing within the `Ontology` class?

LichMaster98: I believe this has been resolved with the newer commits.

adam/ontology/ontology.py
```diff
@@ -61,6 +62,11 @@ class Ontology:
     action_to_description: ImmutableDict[OntologyNode, ActionDescription] = attrib(
         converter=_to_immutabledict, default=immutabledict(), kw_only=True
     )
+    _node_to_relations: ImmutableSetMultiDict[
```
rgabbard: I would suggest just taking an `ImmutableSet` of relations; we can always index them later if we need to.

PR #167, opened 2019-09-18
Represent paths

rgabbard:


rgabbard:
Will merge this without review due to time constraints, but review after the fact would be appreciated.

PR #170, opened 2019-09-18
my your dads

denizbeser:


tests/language_specific/english/test_english_language_generator.py
```diff
@@ -245,6 +246,132 @@ def test_dad_put_a_cookie_in_a_box_using_i():
     ).as_token_sequence() == ("I", "put", "a", "cookie", "in", "a", "box")
 
 
+def test_dad_put_a_cookie_in_a_box_using_you():
```
rgabbard: We also want an test for the simple case of "Dad has a cookie" with no other action going on.

adam/language_specific/english/english_language_generator.py
```diff
@@ -193,7 +194,38 @@ def _translate_object_to_noun(
             if (dependency_node.part_of_speech != PROPER_NOUN) and (
                 MASS_NOUN not in lexicon_entry.properties
             ):
-                if count == 1:
+
+                # Objects that might possess the noun
+                try:
```
rgabbard: I know it is supposedly "Pythonic", but I am not a fan of using exceptions for flow control.

I would suggest rewriting this section in the following way:
* first check if anything has a `HAS` relation with the object in question.
* if it does, then examine that particular object to see how to render it as a word.

adam/language_specific/english/english_language_generator.py
```diff
@@ -193,7 +194,34 @@ def _translate_object_to_noun(
...
+                    # If speaker possesses the noun
+                    if IS_SPEAKER in possessor.properties:
+                        quantifier_node = DependencyTreeToken("my", DETERMINER)
+                        quantifier_role = DETERMINER_ROLE
+                    # If addressee possess the noun
+                    elif IS_ADDRESSEE in possessor.properties:
+                        quantifier_node = DependencyTreeToken("your", DETERMINER)
+                        quantifier_role = DETERMINER_ROLE
+                    # otherwise, if the agent possess the noun
+                    elif possessor == only(
```
rgabbard: Actually, I think the desired behavior is to use _'s_ if a non-agent is the possessor; if the agent is a possessor, just fall back on the regular determiner behavior.  

Also this is getting complex enough we should pull it out into its own `add_determiner` method.

adam/language_specific/english/english_language_generator.py
```diff
@@ -220,6 +209,63 @@ def _translate_object_to_noun(
 
             return dependency_node
 
+        def add_determiner(
+            self,
+            _object: SituationObject,
+            count: int,
+            dependency_node: DependencyTreeToken,
```
rgabbard: Maybe rename to `noun_dependency_node` so it is clear which node you are talking about

adam/language_specific/english/english_language_generator.py
```diff
@@ -220,6 +209,63 @@ def _translate_object_to_noun(
...
+                elif (
+                    len(possession_relations) == 1
+                    and possession_relations[0].first_slot
+                    not in only(self.situation.actions).argument_roles_to_fillers[AGENT]
+                ):
+                    token = (
+                        self.objects_to_dependency_nodes[
+                            possession_relations[0].first_slot
+                        ].token
+                        + "'s"
```
rgabbard: I think most dependency trees analyze `'s` as its own token, attached to the possessor by a `CASE` role.

adam/language_specific/english/english_language_generator.py
```diff
@@ -220,6 +209,63 @@ def _translate_object_to_noun(
...
+            if len(possession_relations) > 1:
+                raise RuntimeError("Can not handle multiple possession relations")
+            else:
+                # If speaker possesses the noun
+                if (
+                    len(possession_relations) == 1
+                    and IS_SPEAKER in possession_relations[0].first_slot.properties
+                ):
+                    quantifier_node = DependencyTreeToken("my", DETERMINER)
+                    quantifier_role = DETERMINER_ROLE
```
rgabbard: for possessives the role should be `nmod:poss`

tests/language_specific/english/test_english_language_generator.py
```diff
@@ -270,6 +273,97 @@ def test_dad_put_a_cookie_in_a_box_using_you():
...
+
+
+def test_dad_has_a_cookie():
+    dad = SituationObject(DAD)
+    cookie = SituationObject(COOKIE)
+    situation = HighLevelSemanticsSituation(
+        ontology=GAILA_PHASE_1_ONTOLOGY,
+        objects=[dad, cookie],
+        relations=[],
+        actions=[SituationAction(HAVE, ((AGENT, dad), (THEME, cookie)))],
```
rgabbard: `HAVE` should not be an available action type (remove it from the `Ontology` if it is there). Rather, if there is a `has` relation and no action, then translate to English using _has_

adam/language_specific/english/english_language_generator.py
```diff
@@ -220,6 +219,66 @@ def _translate_object_to_noun(
...
+                    determiner_node = DependencyTreeToken("your", DETERMINER)
+                    determiner_role = DETERMINER_ROLE
+                # add 's if a non-agent possess the noun
+                elif (
+                    len(possession_relations) == 1
+                    and self.situation.is_dynamic
+                    and possession_relations[0].first_slot
+                    not in only(self.situation.actions).argument_roles_to_fillers[AGENT]
+                ):
+                    token = self.objects_to_dependency_nodes[
```
rgabbard: Needs test covered (which, as you note, requires translating the possessing object)

adam/language_specific/english/english_language_generator.py
```diff
@@ -331,6 +390,54 @@ def _preposition_for_region_as_goal(self, region: Region[SituationObject]) -> st
                     f"Don't know how to translate {region} to a preposition yet"
                 )
 
+        def _translate_relation_to_verb(
+            self, relation: Relation[SituationObject]
+        ) -> DependencyTreeToken:
+            lexicon_entry = self._unique_lexicon_entry(relation.relation_type)
+            if any(
```
rgabbard: Is there similar logic for other verbs that we can refactor into a method?

adam/language_specific/english/english_language_generator.py
```diff
@@ -331,6 +390,54 @@ def _preposition_for_region_as_goal(self, region: Region[SituationObject]) -> st
...
+            )
+            self.dependency_graph.add_node(verb_dependency_node)
+
+            first_slot_dependency_node = self.objects_to_dependency_nodes[
+                relation.first_slot
+            ]
+            self.dependency_graph.add_edge(
+                first_slot_dependency_node,
+                verb_dependency_node,
+                role=self._translate_argument_role(AGENT),
```
rgabbard: I would just specify the roles to be subject and direct object directly; no need to do the translation step.

adam/language_specific/english/english_language_generator.py
denizbeser: That's slightly different because we access those through argument roles.

rgabbard: Got it.

PR #172, opened 2019-09-18
103 represent the ground

elizlee:


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -592,6 +596,12 @@ def _perceive_objects(self) -> None:
             self._instantiate_object_schema(
                 only(object_schemata), situation_object=situation_object
             )
+        # Add ground to _object_perceptions here
```
rgabbard: Rather than tracking if you saw it on the main loop through, prefer instead to do:
```python
if not any(situation_object.ontology_node == GROUND 
                for situation_object in self._situation.objects):
    ...
```

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -592,6 +596,12 @@ def _perceive_objects(self) -> None:
...
+        if not has_ground:
+            object_schemata = self._generator.ontology.structural_schemata(GROUND)
```
rgabbard: Better style would be:

```python
ground_schema = only(self._generator.ontology.structural_schemata(GROUND))
```

tests/perception/high_level_semantics_situation_to_developmental_primitive_perception_test.py
```diff
@@ -246,3 +248,18 @@ def test_learner_perceivable():
     ).frames[0]
 
     assert _some_object_has_binary_property(learner_perception, IS_LEARNER)
+
+
+def test_ground():
```
rgabbard: Also add a test to ensure that if the ground is explicitly specified in a situation, it is not added a second time

PR #197, opened 2019-09-19
add colors to sub-objects in part-of relations

### LINKED ISSUE #113: Add color perceptions even for objects which are not highlighted in the situation
#94  will add color perceptions for objects whose color is explicitly called out in the situation description. It would be more realistic if some color was always specified for an object, even when it isn't salient.

### LINKED ISSUE #176: Perception: inherit colors across part-of relations
If prototypical colors are not specified for your node, though, then that blocks inheritance.

### LINKED ISSUE #249: Give identical RgbColorPerceptions to sub-objects of the same type
e.g. currently if we have a car/truck object, each tire will get its own random "black" color (#000000, #343434, #0c020f...)

### LINKED ISSUE #288: the blocks of color in the perceptual rep. should probably be color:"block", ideally with the name of the color in the block (joe olive is color blind).

### LINKED ISSUE #293: we should block color for people
white Dad is not a great example.

elizlee:
Closes #176 
Closes #288 
Closes #293 
Closes #249 
Closes #113


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -666,6 +676,18 @@ def _instantiate_object_schema(
             self._relation_perceptions.append(
                 Relation(PART_OF, root_object_perception, sub_object_perception)
             )
+            # If the sub-object does not already have a color,
+            # use the parent's prototypical color if one exists
+            sub_object_color = re.findall(
```
rgabbard: If there is a structured way to access data, use it rather than processing things as strings - it's much more robust.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -604,13 +604,13 @@ def _determine_color(
...
         )
 
-        prototypical_colors_for_object_type = immutableset(
-            property_
-            for property_ in self._generator.ontology.properties_for_node(
-                situation_object.ontology_node
-            )
-            if self._generator.ontology.is_subtype_of(property_, COLOR)
-        )
+        # prototypical_colors_for_object_type = immutableset(
```
elizlee: Commenting out these blocks of code prevented the root object from getting re-assigned a random color. Would it be safe to remove these (and let the "prototypical" colors be determined within `_perceive_objects`)?

Issue #157, opened 2019-09-17
Add templates for possession relations


rgabbard:
Also add templates to curiculum of people possessing objects.
Also add language generation to handle generating "has"

rgabbard:
The language conversion is handled, but the template (commented out in `phase1_curriculum.py`):
```python

_PERSON_0 = object_variable("person", PERSON)
_INANIMATE_OBJECT_0 = object_variable("inanimate-object", INANIMATE_OBJECT)
PERSON_HAS_OBJECT_TEMPLATE = Phase1SituationTemplate(
    object_variables=[_PERSON_0, _INANIMATE_OBJECT_0, _LEARNER_OBJECT],
    asserted_always_relations=[Relation(HAS, _PERSON_0, _INANIMATE_OBJECT_0)],
    constraining_relations=[bigger_than(_PERSON_0, _INANIMATE_OBJECT_0)],
)

PERSON_HAS_OBJECT_SUB_CURRICULUM = _phase1_instances(
     "person has object",
     situations=sampled(
         PERSON_HAS_OBJECT_TEMPLATE,
         chooser=_CHOOSER,
         ontology=GAILA_PHASE_1_ONTOLOGY,
         max_to_sample=100,
     ),
 )
```

doesn't work because we want to constrain the thing possessed to be smaller than the holder, but right now we only add size relationships from the ontology in perception generation, not situation generation.

A short-term solution for the M3 deadline is just to add a non-perceivable ontology property `PERSON_CAN_HAVE` and use property filtering rather than size filtering.

LichMaster98:
@rgabbard Does the following sound like a solution to both the issue here and https://github.com/isi-vista/adam/issues/83 ?

In the `Ontology` we add a `Relation[OntologyNode]` to the implicit `relations` in the ontology where the relation `HAS` relates `SMALLER_THAN` and `BIGGER_THAN` then in perception generation when we process `situation.always_relations` we check if the `relation` we are adding has a mapping in the relations of the `Ontology` and if so apply the stored relations to the nodes in the `HAS` relation?

[...]

PR #224, opened 2019-09-20
Add implicit relationships as defined by other relations

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/157


LichMaster98:
This issue: https://github.com/isi-vista/adam/issues/157 was solved by https://github.com/isi-vista/adam/pull/226. Therefore this PR (which was originally for issue 157) is no longer needed

PR #226, opened 2019-09-20
Add PERSON_CAN_HAVE property has a hack for Posession in M3 Deadline.

LichMaster98:
Completes https://github.com/isi-vista/adam/issues/157 with regards to the M3 Deadline -- However, the issue will need a more robust answer


adam/ontology/phase1_ontology.py
```diff
@@ -213,7 +215,11 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
 # Information about the hierarchical structure of objects
 # is given at the end of this module because it is so bulky.
 
-INANIMATE_OBJECT = OntologyNode("inanimate-object", inheritable_properties=[INANIMATE])
+INANIMATE_OBJECT = OntologyNode(
+    "inanimate-object",
+    inheritable_properties=[INANIMATE],
+    non_inheritable_properties=[PERSON_CAN_HAVE],
```
rgabbard: Why here, since it is not inherited and we never instantiate a raw `INANIMATE_OBJECT` anyway?

adam/situation/templates/phase1_templates.py
```diff
@@ -189,13 +197,31 @@ def generate_situations(
                 for obj_var in template.object_variables
             )
 
+            always_relations = []
```
rgabbard: @LichMaster98 : We don't want to special case this logic in for the deadline hack; you can copy it somewhere for when we address this for real.  We are ok with just having the size relations missing for now.

adam/curriculum/phase1_curriculum.py
```diff
@@ -163,15 +163,15 @@ def build_object_multiples_situations(
...
-# PERSON_HAS_OBJECT_SUB_CURRICULUM = _phase1_instances(
-#     "person has object",
-#     situations=sampled(
-#         PERSON_HAS_OBJECT_TEMPLATE,
-#         chooser=_CHOOSER,
-#         ontology=GAILA_PHASE_1_ONTOLOGY,
-#         max_to_sample=100,
-#     ),
-# )
+PERSON_HAS_OBJECT_SUB_CURRICULUM = _phase1_instances(
```
rgabbard: Don't we need a property restriction on the object variable for `PERSON_HAS_OBJECT_TEMPLATE`?

adam/ontology/phase1_ontology.py
LichMaster98: The curriculum for PERSON_HAS_OBJECT instantiates a raw `INANIMATE_OBJECT`. You can find this in `phase1_curriculum.py` line 166. I can change this to a `BALL` and therefore could remove this addition.

rgabbard: @LichMaster98 : The interpretation of

```
object_variable("object_0", INANIMATE_OBJECT)
```

is that the variable ahs to be bound to a sub-type of `INANIMATE_OBJECT` which has the `CAN_FILL_TEMPLATE_SLOT` (IIRC) property, so `INANIMATE_OBJECT` itself will never be instantiated.

We will want to change this to
```
object_variable("object_0", INANIMATE_OBJECT,
     required_properties=[PERSON_CAN_HAVE])
```

(from memory, may not be exactly right)

PR #253, opened 2019-09-20
Add drink to the Curriculum

LichMaster98:


adam/curriculum/phase1_curriculum.py
```diff
@@ -582,6 +583,38 @@ def _make_roll_curriculum():
...
+def _make_drink_curriculum():
+    object_0 = object_variable(
+        "object_0", required_properties=[HOLLOW], banned_properties=[IS_BODY_PART]
+    )
+    liquid_0 = object_variable("liquid_0", required_properties=[LIQUID])
+    person_0 = object_variable("person_0", PERSON)
+
+    drink_liquid = Phase1SituationTemplate(
+        "drink",
+        salient_object_variables=[object_0, liquid_0, person_0],
```
rgabbard: `object_0` should go on `other_objects` (or whatever it is called) rather than `salient_objects`,

adam/curriculum/phase1_curriculum.py
```diff
@@ -582,6 +583,38 @@ def _make_roll_curriculum():
     )
 
 
+def _make_drink_curriculum():
```
rgabbard: Can you add a test for the language generation, which ensures that the container is not mentioned? (e.g. `a baby drinks juice`)

adam/curriculum/phase1_curriculum.py
LichMaster98: @rgabbard I see an `all_object_variables` field on the situation template but I don't see an `other_objects` (or equivelant).

It looks like the situation infers the auxiliary objects from the `auxiliary_variable_bindings` so I'll just remove `object_0` from the list there.

rgabbard: @LichMaster98 : Yes, I remember now - you are correct, and that is the correct thing to do

PR #254, opened 2019-09-20
add paragraph to HTML dump

elizlee:
WIP


adam/curriculum_to_html.py
```diff
@@ -127,6 +127,42 @@ def _dump_instance_group(
         with open(output_destination, "w") as html_out:
             html_out.write(f"<head>\n\t<style>{CSS}\n\t</style>\n</head>")
             html_out.write(f"\n<body>\n\t<h1>{title} - {instance_group.name()}</h1>")
+            html_out.write("\n<h4>How to read</h4>")
```
LichMaster98: NITPICK: This should probably be `<h2>` rather than `<h4>`

adam/curriculum_to_html.py
```diff
@@ -127,6 +127,42 @@ def _dump_instance_group(
...
+                "semantic roles.</li> "
+                "\n\t<li>Meta-properties provide information about a property, e.g. whether or not a property can "
+                "be perceived by the learner.</li> "
+                "\n\t<li>OBJECT[PROPERTY_0[META_PROPERTY], PROPERTY_1, PROPERTY_2]</li> "
+                "\n</ul>\n<ul>\n<p>Relations</p> "
+                "\n\t<li>These define relations between two objects; e.g. smallerThan(A, B) indicates that object A is "
+                "smaller than object B.</li> "
+                "\n\t<li>IN_REGION is a special relation that holds between a SituationObject and a Region."
+                "\n\t<li>RELATION(OBJECT_0, OBJECT_1)</li>"
+                "\n</ul>\n<ul>\n<p>Part-of nesting</p>"
```
LichMaster98: I think this section could use a clarification of why `IN_REGION` is also a part of a nested relationship rather than being in the `Other Relations` section. (My understanding is that it is entirely for readability but explaining that could be beneficial)

adam/curriculum_to_html.py
```diff
@@ -127,6 +127,43 @@ def _dump_instance_group(
         with open(output_destination, "w") as html_out:
             html_out.write(f"<head>\n\t<style>{CSS}\n\t</style>\n</head>")
             html_out.write(f"\n<body>\n\t<h1>{title} - {instance_group.name()}</h1>")
+            html_out.write("\n<h2>How to read</h2>")
+            html_out.write(
+                "\n<p>A Situation is curriculum-designer-facing; it is not accessible to the learner. "
+                "Names on Perceptions are handles for debugging; they are not accessible to the "
```
LichMaster98: @elizlee, from my discussion with Marjorie we might want to add a sentence here describing how to do the mapping from the Situation Objects to the perceived objects. (Example: Baby becomes `person_0` which might not be intuitive to follow)

Issue #104, opened 2019-09-11
Add Dowty's proto-roles to perceptual generation


rgabbard:

rgabbard:
@denizbeser : Dowty's paper is http://linguistics.berkeley.edu/~syntax-circle/syntax-group/dowty91.pdf ; we are interested in (27) and (28) in p. 572.

@charles-yang-upenn  @mitchmarcus : I think we don't want to include the proto-roles in the perception directly, but rather than finer-grained properties (a)-(d) under each of (27) and (28) (and assume the learner module uses these to form proto-roles).  Do you agree?

@denizbeser : Pending comment from Mitch and Charles, proceed with making perceivable properties for all of (a) - (d) under (27) and (28).  Then add an `asserted_properties`  field to `ActionDescription` which should be a map from `SituationObject` to a collection of `OntologyNode`s for properties we want to say it has if this action takes place.  Finally, alter perception generation to apply these properties.

rgabbard:
@denizbeser : example for `PUT`:

```python
_PUT_ACTION_DESCRIPTION = ActionDescription(
    frame=ActionDescriptionFrame({AGENT: _PUT_AGENT, THEME: _PUT_THEME, GOAL: _PUT_GOAL}),

    ....
    asserted_properties = [
    (_PUT_AGENT, VOLITIONAL_INVOLVMENT),
    (_PUT_AGENT, CAUSES_CHANGE_OF_STATE),
    (_PUT_THEME, CHANGE_OF_STATE),
)
```

I think there are more we could perhaps add, but that gives you the idea.  I am actually a bit unsure how the `GOAL` of `put` should be handled; I see [here](http://web.stanford.edu/~bclevin/dowty19fin.pdf) that: 
> Some researchers recast additional traditional semantic roles as proto-roles. Primus
(1999:54–55, 137–138) introduces a Recipient Proto-role to deal with phenomena involving ditransitive verbs, benefactives, and addressees. This proto-role combines Proto-Agent
and Proto-Patient properties.

@charles-yang-upenn  @mitchmarcus : If either of you has time today and would like to pitch in, I'm sure @denizbeser would appreciate help in assigning the proto-role sub-properties to semantic roles of the various actions in the vocabulary.

mitchmarcus:
[...]

PR #262, opened 2019-09-23
Dowty's proto-roles

denizbeser:
Linked to #104

### DIFF
--- a/adam/ontology/phase1_ontology.py
+++ b/adam/ontology/phase1_ontology.py
@@ -147,6 +147,33 @@ implicity generated in the perception step if not explicit in a situation.
 """
 subtype(GAZED_AT, PERCEIVABLE_PROPERTY)
 
+
+# Dowty's Proto-Roles: Issue 104; Dotwy, 91, page 572.
+# Agent Proto-Roles properties:
+AGENT_PROTO_ROLE = OntologyNode("agent-proto-role")
+subtype(AGENT_PROTO_ROLE, PERCEIVABLE_PROPERTY)
+VOLITIONALLY_INVOLVED = OntologyNode("volitionally-involved", [BINARY])
+subtype(VOLITIONALLY_INVOLVED, AGENT_PROTO_ROLE)
+SENTIENT_OR_PERCEIVES = OntologyNode("sentient-or-perceives", [BINARY])
+subtype(SENTIENT_OR_PERCEIVES, AGENT_PROTO_ROLE)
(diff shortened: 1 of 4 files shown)


adam/ontology/phase1_ontology.py
```diff
@@ -147,6 +147,33 @@ def subtype(sub: OntologyNode, _super: OntologyNode) -> None:
 """
 subtype(GAZED_AT, PERCEIVABLE_PROPERTY)
 
+
+# Dowty's Proto-Roles: Issue 104; Dotwy, 91, page 572.
+# Agent Proto-Roles properties:
+AGENT_PROTO_ROLE = OntologyNode("agent-proto-role")
```
rgabbard: We don't want to include the agent and patient proto-roles directly; rather the learner will figure them out from the atomic properties like `VOLITIONALLY_INVOLVED`, etc.

adam/ontology/phase1_ontology.py
```diff
@@ -939,6 +966,11 @@ def _below_region_factory(reference_object: ObjectT) -> Region[ObjectT]:
         Relation(IN_REGION, _PUT_THEME, _CONTACTING_MANIPULATOR, negated=True),
         Relation(IN_REGION, _PUT_THEME, _PUT_GOAL),
     ],
+    asserted_properties=[
+        (_PUT_AGENT, VOLITIONALLY_INVOLVED),
+        (_PUT_AGENT, CAUSES_CHANGE),
+        (_PUT_THEME, CAUSALLY_AFFECTED),
```
rgabbard: `_PUT_GOAL` could also be specified as `STATIONARY`, I think

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -581,6 +581,32 @@ def _perceive_property_assertions(self) -> None:
...
+                    role
+                ]  # e.g. _PUT_AGENT
+                fillers = action.argument_roles_to_fillers[role]  # e.g. {Mom}
+                for property_ in action_description.asserted_properties[variable]:
+                    for situation_or_region in fillers:
+                        if isinstance(situation_or_region, SituationObject):
+                            perception_of_object = self._objects_to_perceptions[
+                                situation_or_region
+                            ]
+                        else:
```
rgabbard: Drop a comment on the `else` that we are here propagating properties asserted on regions to their reference objects; make an issue for post-M3 for us to think carefully about whether this is really the right thing to do.

PR #271, opened 2019-09-23
Reduce pairs of size relations to one line

### LINKED ISSUE #219: More compact representation of size relations
In the HTML dumps, if the relations `smallerThan(X, Y)` and `biggerThan(Y, X)` both hold in the perceptual representation, don't print either of those relations normally. Instead, include one line of the form `X > Y`.

elizlee:
Closes #219

### DIFF
--- a/adam/curriculum_to_html.py
+++ b/adam/curriculum_to_html.py
@@ -19,7 +19,7 @@ from adam.experiment import InstanceGroup
 from adam.language.dependency import LinearizedDependencyTree
 from adam.ontology import IN_REGION
 from adam.ontology.during import DuringAction
-from adam.ontology.phase1_ontology import PART_OF
+from adam.ontology.phase1_ontology import PART_OF, SMALLER_THAN, BIGGER_THAN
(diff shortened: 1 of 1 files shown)


adam/curriculum_to_html.py
```diff
@@ -398,9 +398,35 @@ def dfs_walk(node):
...
-                        f"\t\t\t\t\t\t<li>{relation_prefix}{relation}{relation_suffix}</li>"
-                    )
+                    # if matching smallerThan/biggerThan relations exist, give as single relation
+                    opposite_relations = {
+                        SMALLER_THAN: BIGGER_THAN,
+                        BIGGER_THAN: SMALLER_THAN,
+                    }
+                    single_size_relation = None
+                    if relation.relation_type in opposite_relations:
+                        if any(
```
rgabbard: Note this will be quadratic time in the total number of relations, which can sometimes be large.  You could index the relations by tuples of their slot pairs to provide O(1) lookup

PR #277, opened 2019-09-23
wip, put template

denizbeser:


rgabbard:
@denizbeser : The problem was the ontology didn't know dogs could manipulate objects with their heads. :-)

denizbeser:
Oh I see, reviewing the code now. Thanks!

Issue #240, opened 2019-09-20
Unless otherwise specified, objects should be on the ground


mitchmarcus:
Looking at the simplest one object contexts, ground_0 exists as does the object, but the relation between them seems undefined.  Does this matter?

rgabbard:
Yes, things ought to have some relationship to the ground.  The trick is determining automatically exactly when to assert it.  I think we can get 90% of the way there by asserting that an object rests on the ground unless some other region is specified for it.  This isn't quite right, because it could be specified as e.g. _beside_ another object, but it is probably good enough for the M3 checkpoint.

PR #285, opened 2019-09-23
Objects should Contact the Ground unless otherwise stated

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/240


denizbeser:
We still have to handle relations between objects and ground for objects that already have relations, #309

PR #286, opened 2019-09-23
Update curriculum_to_html.py

marjorief:
@marjorief first attempt at updating instructions .  1st to paragraphs are new and mine.  properties still needs to be revised

### DIFF
--- a/adam/curriculum_to_html.py
+++ b/adam/curriculum_to_html.py
@@ -129,14 +129,21 @@ class CurriculumToHtmlDumper:
             html_out.write(f"\n<body>\n\t<h1>{title} - {instance_group.name()}</h1>")
             html_out.write("\n<h2>How to read</h2>")
             html_out.write(
-                "\n<p>A Situation is curriculum-designer-facing; it is not accessible to the learner. "
-                "The name of a perceived object is derived from the type of the corresponding SituationObject, and the "
-                "number indicates which object of its type it is - e.g. if there are two babies in a situation, "
-                "the perceived objects will be 'person_0' and 'person_1'. "
-                "Names on Perceptions are handles for debugging; they are not accessible to the "
-                "learner. 'GAZED-AT' refers the object of focus of the speaker.</p> "
-            )
+                "\n<p>Our curiculum is designed in two levels: The objective, here - {instance_group.name},"
+                "\n and a set of scenes for each objective. The scenes are in the shaded table below."
+                "\n For each scene, we display: "
+                "\n\t<li>Situation (column 1), which is used by the curriculum developer to specify key objects and actions,"
+                "\n but is unseen by the learner</li>."  
+                "\n\t<li>Language (column 2), the text presented to the learner</li>."  
+                "\n\t<li>Learner Perception (column 3), a textual representaiton of the perceptual features"
+                "\n presented to the learner</li>"             )
             html_out.write(
+                "<p> The perceptual features described in column 3 are presented to the learner in a"
+                "\n machine interpretable fashion (e.g. as attributes, variables, spatial properties). Here "
+                "\n to assist interpretation of the curriculum we serialize them as debugging text strings." 
+                "\n The following are descriptions of some of the perceptual inputs (and their serialization). "
+            )
+            html_out.write(   
                 "\n<p>Properties</p>\n<ul>"
                 "\n\t<li>These define characteristics of an object that may influence perception and fulfillment of "
                 "semantic roles.</li> "
--- a/adam/curriculum_to_html.py
+++ b/adam/curriculum_to_html.py
@@ -142,7 +142,7 @@ class CurriculumToHtmlDumper:
             )
             html_out.write(
                 "<p> The perceptual features described in column 3 are presented to the learner in a"
-                "\n machine interpretable fashion (e.g. as attributes, variables, spatial properties). Here "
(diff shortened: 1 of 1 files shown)


adam/curriculum_to_html.py
```diff
@@ -128,48 +128,33 @@ def _dump_instance_group(
             html_out.write(f"<head>\n\t<style>{CSS}\n\t</style>\n</head>")
             html_out.write(f"\n<body>\n\t<h1>{title} - {instance_group.name()}</h1>")
             html_out.write("\n<h2>How to read</h2>")
+            html_out.write(f"\n<p>Our curiculum is designed in two levels: The objective, here - {instance_group.name()}")
```
rgabbard: What the interpolation of `instance_group.name()` here?

adam/curriculum_to_html.py
```diff
@@ -128,48 +128,33 @@ def _dump_instance_group(
...
             html_out.write(
-                "\n<p>A Situation is curriculum-designer-facing; it is not accessible to the learner. "
-                "The name of a perceived object is derived from the type of the corresponding SituationObject, and the "
-                "number indicates which object of its type it is - e.g. if there are two babies in a situation, "
-                "the perceived objects will be 'person_0' and 'person_1'. "
-                "Names on Perceptions are handles for debugging; they are not accessible to the "
-                "learner. 'GAZED-AT' refers the object of focus of the speaker.</p> "
+                "\n and a set of scenes for each objective. The scenes are in the shaded table below."
```
rgabbard: Something seems to be missing? This starts with "and a set of scenes..."?

adam/curriculum_to_html.py
```diff
@@ -128,48 +128,33 @@ def _dump_instance_group(
...
-                "defined in terms of a PathOperator and whether or not the orientation changed, with respect to a "
-                "reference object and an optional reference axis.</li>"
-                "\n\t<li>e.g. the path of a falling object may be represented as SpatialPath(operator=TOWARD, "
-                "reference_object=GROUND, reference_axis=None, orientation_changed=False)</li>"
-                "\n\t<li>SpatialPath(OPERATOR, REFERENCE_OBJECT, REFERENCE_AXIS, ORIENTATION_CHANGED)</li>"
+                "\n\t<li>Part of relations (e.g. a head being a part of a person) are represented with the parts 
+                "\n\t appearing as sub-bullets of an object (e.g. head as a sub-bullet under person).</li>"
+                "\n\t<li>Numeric indices (e.g. person_0 person_1 leg_0, leg_1) distinguish between objects of the"
+                "\n\t same class. </li>"
+                "\n\t<li>Region and in-region relations describe the spatial relations between objects</li>"
```
rgabbard: Just `IN_REGION`, I think?

adam/curriculum_to_html.py
marjorief: @rgabbard -- copied from above to explain what the objective is (it could be replaced with "e.g. learning objects, learning roll", but it seemed more clear to me to state what the current objective is

adam/curriculum_to_html.py
marjorief: @rgabbard what does Region do then?

adam/curriculum_to_html.py
marjorief: i suspect this is bad merging--- should I edit in the file?

marjorief: actually, not bad merging... perhaps bad use of python. 
```
            html_out.write(f"\n<p>Our curiculum is designed in two levels: The objective, here - {instance_group.name()}")
            html_out.write(
                "\n<p>Our curiculum is designed in two levels: The objective, here - {instance_group.name},"
                "\n and a set of scenes for each objective. The scenes are in the shaded table below."
                "\n For each scene, we display: "
                "\n\t<li>Situation (column 1), which is used by the curriculum developer to specify key objects and actions,"
                "\n but is unseen by the learner</li>."  
                "\n\t<li>Language (column 2), the text presented to the learner</li>."  
                "\n\t<li>Learner Perception (column 3), a textual representaiton of the perceptual features"
                "\n presented to the learner</li>"             )
                "\n presented to the learner</li>")
```

marjorief: and the repeated line I had thought was deleted.

adam/curriculum_to_html.py
LichMaster98: @marjorief `Region` is a class which holds an `object` and `distance` indicators for use in mapping relationships. This allows us to define that a cup is in a proximal relationship to a book. Or example being `in-region(cup_0, REGION(table_0,distance=exterior-but-in-contact))` so that we know the cup is on the exterior region of the table.

LichMaster98:
These updates will also need to explain that `mom > cup` implies a size relationship of mom being bigger than a cup and the inverse. See https://github.com/isi-vista/adam/pull/271 for the implementation and 
image for the display on an HTML page.

adam/curriculum_to_html.py
rgabbard: @marjorief : basically, `IN_REGION` is the relation type; a `Region` object is the second argument of an `IN_REGION` relation.  So if there is a ball on a table, there is the following relation:
```
IN_REGION(ball, Region(reference=table, distance=EXTERIOR_BUT_IN_CONTACT, direction=GRAVITATIONAL_UP)
```

gabbard:
I incorporated these changes manually.

Issue #189, opened 2019-09-19
Add information about the speaker to the HTML


denizbeser:
To represent the speaker information, when we have a speaker we should render it as something like the following:
- "Mom says: Dad puts the cookie in my box"
so we can resolve to whom "my" refers.

PR #322, opened 2019-09-24
Display speaker information

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/189


LichMaster98:
Waiting on https://github.com/isi-vista/adam/pull/256 for testing

LichMaster98:
image

marjorief:
@rgabbard , @LichMaster98 does this change require a a change to the instructions is "Dad says..." the language that the learners hears, or is the fact that someone is saying something a part of the perception of the scene.

LichMaster98:
@marjorief The phrasing of `Dad says "Mom gave me the ball"` is designated to give the reader an easier understanding of who "me" is. It takes the language from being a floating voice to a specific speaker. The learner does perceive who `IS_SPEAKER`.

rgabbard:
@marjorief : "Dad says" is not part of the language the learner hears; rather, the learner knows it is Dad who utters the language.  It would be good to add something on this to the instructions.

marjorief:
IMO, we would want the `Dad Says` to appear as a feature in the perception column. @rgabbard lmk if given where we are, I should just make a change to the instructions that says. 

"Language is typically the language presented to learner. When the speaker is a part of the scene  , this column also includes an attribution statement (e.g. Dad says..) which is not presented to the learner (but rather a part of what the learner perceives.) ."

rgabbard:
We do represent it in the perception, but there is so much stuff there it is difficult to pick out, so I think adding to the instructions as you suggest is good

marjorief:
OK. I've changed the instructions.

Issue #283, opened 2019-09-23
jump -- Dad jumps over a hand


marjorief:
This could be strange perceptually. Is it a disembodied hand?

marjorief:
@rgabbard -- this issue seems to be in  part do we want to allow disembodied body parts.  E.g. "a red hand" we have hand_0 that does not belong to any person.

rgabbard:
Closed by #324

PR #324, opened 2019-09-24
Don't jump over body parts or liquids

rgabbard:


LichMaster98:
Closes https://github.com/isi-vista/adam/issues/283

Issue #200, opened 2019-09-19
Handle core vocabulary verb: Come


denizbeser:
Come (~1500): almost always with directional/locational PP/ADVP (~1200; more including idioms e.g., “come on”), occasionally taking VP 

Come: almost all directional

rgabbard:
@LichMaster98 : for the initial templates, just do a bare _X comes_ without the PP/ADVP because we don't have the right prepositions yet.  Make a separate issue for handling those modifiers.

For now, treat the distinction between _go_ and _come_ as that for _come_, the path must be towards the speaker (or if no speaker, towards the learner, let's say; make two separate templates)

Issue #239, opened 2019-09-20
Language generation does not merge in "during" semantics of actions with those explicitly specified in the situation


gabbard:

rgabbard:
Once fixed we can remove the `FALL` hack in `_collect_action_modifiers`

rgabbard:
We have to hack around this in the `push` template

gabbard:
@LichMaster98 @denizbeser : Is this still true?

gabbard:
Closed by #704

PR #330, opened 2019-09-24
Introduce come curriculum

LichMaster98:
closes https://github.com/isi-vista/adam/issues/200


rgabbard:
This is not picking up the "closer to the speaker" semantics on the perceptual side, probably due to https://github.com/isi-vista/adam/issues/239. Don't pursue this at the moment.

rgabbard:
This is because it was marked as nehated.

PR #358, opened 2019-10-10
Collapse geon axes relationships

### LINKED ISSUE #354: Geon Axes Relationships don't collapse `bigger_than`, etc
image (1)

Much bigger than will be indicated via ">>"

This may be most easily accomplished by turning the section of code which processes relationships in the HTML process into its own function. Additionally, the explanation section will need to include explaining the ">>" notation.

elizlee:
Closes #354


adam/curriculum_to_html.py
```diff
@@ -85,7 +92,8 @@
     "\n\t during the scene are preceded by 'Ø --->'. Those that cease to exist during a scene are"
     "\n\t followed by '---> Ø'  </li>"
     "\n\t<li>Size relationships are indicated by the '>' symbol. We only display the bigger than size"
```
LichMaster98: Explanation of ">" is duplicated.

gabbard:
@elizlee : can you post a screenshot of the relevant part of the output?

elizlee:
@gabbard
Screen Shot 2019-10-16 at 12 38 39 PM

adam/curriculum_to_html.py
```diff
@@ -660,31 +718,12 @@ def dfs_walk(node):
...
-                    ):
-                        if relation.relation_type == SMALLER_THAN:
-                            single_size_relation = (
-                                f"{relation.second_slot} > {relation.first_slot}"
-                            )
-                        else:
-                            single_size_relation = (
-                                f"{relation.first_slot} > {relation.second_slot}"
-                            )
+                single_size_relation: Optional[
```
gabbard: nit: if you annotate the return type of the function you won't need the variable type annotation here.

PR #359, opened 2019-10-10
Subset learning algorithm

### LINKED ISSUE #348: Implement subset learning based approach for single object detection
For our first try, assume the objects are the only thing in the scene and that the linguistic representation is a single word (or single word plus an article)

denizbeser:
closes #348

### DIFF
--- a/adam/learner.py
+++ b/adam/learner.py
@@ -3,7 +3,7 @@ Interfaces for language learning code.
 """
 
 from abc import ABC, abstractmethod
-from typing import Dict, Generic, Mapping
+from typing import Dict, Generic, Mapping, Set
 
 from attr import Factory, attrib, attrs
 from attr.validators import instance_of
@@ -11,6 +11,7 @@ from immutablecollections import immutabledict
 
 from adam.language import LinguisticDescription, LinguisticDescriptionT
 from adam.perception import PerceptionT, PerceptualRepresentation
+from adam.perception.developmental_primitive_perception import PropertyPerception
 
 
 @attrs(frozen=True)
@@ -46,7 +47,7 @@ class LanguageLearner(ABC, Generic[PerceptionT, LinguisticDescriptionT]):
 
     @abstractmethod
     def observe(
-        self, learning_example: LearningExample[PerceptionT, LinguisticDescriptionT]
+            self, learning_example: LearningExample[PerceptionT, LinguisticDescriptionT]
     ) -> None:
         """
         Observe a `LearningExample`, possibly updating internal state.
@@ -54,7 +55,7 @@ class LanguageLearner(ABC, Generic[PerceptionT, LinguisticDescriptionT]):
 
     @abstractmethod
     def describe(
-        self, perception: PerceptualRepresentation[PerceptionT]
+            self, perception: PerceptualRepresentation[PerceptionT]
     ) -> Mapping[LinguisticDescriptionT, float]:
         r"""
         Given a `PerceptualRepresentation` of a situation, produce one or more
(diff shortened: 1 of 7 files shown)


adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+        is_new_description = True
+        observed_linguistic_description = learning_example.linguistic_description
+        if isinstance(
+            learning_example.perception.frames[0], DevelopmentalPrimitivePerceptionFrame
+        ):
+            observed_property_assertions = learning_example.perception.frames[
+                0
+            ].property_assertions
+        else:
+            raise RuntimeError("Cannot process perception type.")
```
gabbard: Consider instead:

```suggestion
            raise RuntimeError(f"Cannot process perception of type {type(perception)}")
```

If you habitually include helpful information like that in your exception messages, future-you (and future-others) will thank you often.

adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+    def observe(
+        self, learning_example: LearningExample[PerceptionT, LinguisticDescriptionT]
+    ) -> None:
+        if len(learning_example.perception.frames) != 1:
+            raise RuntimeError("Subset learner can only handle single frames for now")
+
+        is_new_description = True
+        observed_linguistic_description = learning_example.linguistic_description
+        if isinstance(
+            learning_example.perception.frames[0], DevelopmentalPrimitivePerceptionFrame
```
gabbard: If you're going to refer to something more than once, pull it out into a variable.

adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+        ):
+            observed_property_assertions = learning_example.perception.frames[
+                0
+            ].property_assertions
+        else:
+            raise RuntimeError("Cannot process perception type.")
+
+        for description in self._descriptions_to_properties.keys():
```
gabbard:
```suggestion
        for description in self._descriptions_to_properties:
```

(if you iterate over a dict in Python, it iterates over the keys)

adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+            # If already observed, reduce the properties for that description to the common subset of
+            # the new observation and the previous observations
+            if (
+                observed_linguistic_description.as_token_sequence()
```
gabbard: Make an issue that we should relax this requirement in the future (e.g. to apply the learner to _the blue truck_ and _the red truck_)

adam/learner.py
gabbard: A bigger problem is that this approach requires iterating over every description you've ever seen for each training example you encounter - that's quadratic time. :-( . Why not just a simple map lookup?

adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+                == description.as_token_sequence()
+            ):
+                self._descriptions_to_properties[
+                    description
+                ] = self._descriptions_to_properties[description].intersection(
```
gabbard: Does doing the simple intersection here work? I thought perceived objects were never equal to one another?

adam/learner.py
```diff
@@ -102,3 +106,77 @@ def describe(
...
+        # get the learned description for which there are the maximum number of matching properties (i.e. most specific)
+        max_matching_properties = 0
+        learned_description = None
+        for description, properties in self._descriptions_to_properties.items():
+            if all(prop in observed_property_assertions for prop in properties) and (
+                len(properties) > max_matching_properties
+            ):
+                learned_description = description
+                max_matching_properties = len(properties)
+        if learned_description and max_matching_properties > 0:
```
gabbard: If `learned_description` is non-None than `max_matching_properties` is already known to be >0

tests/learner/learner_test.py
```diff
@@ -0,0 +1,58 @@
...
+        learner_factory=SubsetLanguageLearner,
+        pre_example_training_observers=[TopChoiceExactMatchObserver("pre")],
+        post_example_training_observers=[TopChoiceExactMatchObserver("post")],
+        # warm_up_test_instance_groups=[ball_curriculum],
+        test_instance_groups=[test_ball_curriculum],
+        test_observers=[TopChoiceExactMatchObserver("test")],
+        sequence_chooser=RandomChooser.for_seed(0),
+    )
+
+    execute_experiment(experiment)
```
gabbard: If something is going to be a unit test, it needs something `assert`able at the end (unless you are just testing it doesn't crash).  The `Experiment` setup is good for rentry points but not great for unit tests. I would also suggest including multiple object types in the unit test to demonstrate they can be distinguished.

adam/learner.py
denizbeser: With this implementation we ensure we get the object that has the matching description, and intersect the properties, which can be equal. Here is an example for the 
```
First set: {hasProperty(learner, is-learner), hasProperty(ball_0, inanimate[binary]), hasColor(ball_0, #6e5f13), hasProperty(learner, self-moving[binary]), hasColor(ball_0, #c40233), hasProperty(learner, animate[binary])}
Second set: {hasColor(ball_0, #f2003c), hasProperty(ball_0, inanimate[binary]), hasProperty(learner, is-learner), hasProperty(learner, animate[binary]), hasProperty(learner, self-moving[binary])}
```

adam/learner/subset.py
```diff
@@ -0,0 +1,89 @@
...
+@attrs
+class SubsetLanguageLearner(
+    Generic[PerceptionT, LinguisticDescriptionT],
+    LanguageLearner[PerceptionT, LinguisticDescriptionT],
+):
+    """
+    An implementation of `LanguageLearner` for subset learning based approach for single object detection.
+    """
+
+    _descriptions_to_properties: Dict[
```
gabbard: Why not just map directly from the tokens to the properties without the `LinguisticDescriptionT` as middle-man?

PR #360, opened 2019-10-11
Assign the same color to sub-objects of the same type

elizlee:
Addresses #249

### DIFF
--- a/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
+++ b/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
@@ -1,5 +1,15 @@
 from itertools import chain
-from typing import AbstractSet, Any, Dict, List, Mapping, Optional, Union, cast
+from typing import (
+    AbstractSet,
+    Any,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Union,
+    cast,
+    MutableMapping,
+)
 
 from attr import Factory, attrib, attrs
 from attr.validators import instance_of
@@ -768,7 +778,6 @@ class _PerceptionGeneration:
                 # which we will recursively instantiate.
                 perceived_object = self._instantiate_object_schema(
                     only(object_schemata),
-                    {},
                     situation_object=situation_object,
                     inherited_color=color,
                 )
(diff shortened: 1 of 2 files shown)


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -798,6 +799,8 @@ def _perceive_object(self, situation_object: SituationObject) -> None:
     def _instantiate_object_schema(
         self,
         schema: ObjectStructuralSchema,
+        # This will track what types of objects have which RgbColorPerceptions
+        sub_object_nodes_to_rgb: Dict[OntologyNode, RgbColorPerception],
```
gabbard: Prefer more general types like `Mapping` as function arguments (it gives the caller more flexibility)

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -768,6 +768,7 @@ def _perceive_object(self, situation_object: SituationObject) -> None:
                 # which we will recursively instantiate.
                 perceived_object = self._instantiate_object_schema(
                     only(object_schemata),
+                    {},
```
gabbard: Use the keyword argument format here; otherwise the reader has no clue what this empty dictionary is for

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -876,13 +879,20 @@ def _instantiate_object_schema(
                 if self._generator.ontology.is_subtype_of(property_, COLOR)
             )
 
-            if sub_object_colors:
-                sub_object_color: Any = self._chooser.choice(sub_object_colors)
+            if sub_object.schema.ontology_node in sub_object_nodes_to_rgb:
```
gabbard: Would it suffice to handle this at the level of this for loop without needing to pass anything as a function argument?  that is, before the `for sub_object in schema.sub_objects` loop, gather the ontology node types of the sub-objects and assign one color per ontology node type, then use those colors within the loop?

elizlee: I think it might!

PR #371, opened 2019-10-18
Add Curriculum for Training and Testing Prepositions

LichMaster98:


LichMaster98:
Find attached a sample of the curriculum output for this training & testing set. The training set always uses the same objects with all the variations of the use of the preposition while the training set varies the objects over the prepositions to see if we can get the expected prepositional result. 
[prepositions.zip](https://github.com/isi-vista/adam/files/3742188/prepositions.zip)

adam/curriculum_to_html.py
```diff
@@ -97,30 +98,35 @@
...
+    "phase1": GAILA_PHASE_1_CURRICULUM,
+    "prepositions": make_prepositions_curriculum(),
+}
 
 
 def main(params: Parameters) -> None:
     root_output_directory = params.creatable_directory("output_directory")
-    phase1_curriculum_dir = root_output_directory / "gaila-phase-1"
+    curriculum_string = params.optional_string(
+        "curriculum", valid_options=["phase1", "prepositions"], default="phase1"
```
gabbard: nit: you could do `valid_options=STR_TO_CURRICULUM.keys()` and then you don't need to worry about keeping them in sync.

adam/curriculum/phase1_curriculum.py
```diff
@@ -113,25 +113,30 @@
...
     HighLevelSemanticsSituation,
     LinearizedDependencyTree,
     DevelopmentalPrimitivePerceptionFrame,
 ]
 
-_GROUND_OBJECT = object_variable("ground", GROUND)
-
 
-def _standard_object(
+def standard_object(
```
gabbard: Can you shift this and other code shared with the curriculum to a `curriculum_utils` module?

adam/curriculum/preposition_curriculum.py
```diff
@@ -0,0 +1,726 @@
...
+        asserted_always_relations=[
+            Relation(
+                IN_REGION,
+                figure,
+                Region(
+                    ground,
+                    distance=PROXIMAL,
+                    direction=Direction(
+                        positive=is_right,
+                        relative_to_axis=HorizontalAxisOfObject(ground, index=0),
```
gabbard: Should this be relative to the horizontal axis of the object itself, rather than the ground?

adam/curriculum/preposition_curriculum.py
```diff
@@ -0,0 +1,726 @@
...
+def _under_template(
+    figure: TemplateObjectVariable,
+    ground: TemplateObjectVariable,
+    background: Iterable[TemplateObjectVariable],
+) -> Phase1SituationTemplate:
+    return Phase1SituationTemplate(
+        f"preposition-training-{figure.handle}-under-{ground.handle}",
+        salient_object_variables=[figure],
+        background_object_variables=background,
+        asserted_always_relations=[strictly_above(figure, ground)],
```
gabbard: Here and in `_over` - can you check how we have been using `strictly_above`? I thought the upper thing came first and the lower second.

adam/curriculum/preposition_curriculum.py
```diff
@@ -0,0 +1,726 @@
...
+    )
+
+
+def _in_template(
+    figure: TemplateObjectVariable,
+    ground: TemplateObjectVariable,
+    background: Iterable[TemplateObjectVariable],
+) -> Phase1SituationTemplate:
+    return Phase1SituationTemplate(
+        f"preposition-training-{figure.handle}-in-{ground.handle}",
```
gabbard: Do we need to assert the ground is `HOLLOW`?

adam/curriculum/preposition_curriculum.py
LichMaster98: This mirrors the way `beside` was constructed in the original curriculum -- That said, I'm open to the change suggested.

adam/curriculum/preposition_curriculum.py
LichMaster98: That's done inside the curriculum constructors as a `required_property` of ground.

adam/curriculum/preposition_curriculum.py
gabbard: Make a separate issue for us to think about it later, on the M3 cleanup milestone

LichMaster98: https://github.com/isi-vista/adam/issues/377

adam/curriculum/preposition_curriculum.py
LichMaster98: That's correct, I've adjusted the inputs. I needed to provide the `salient_object_variable` as the `ground` to generate the correct output.

PR #382, opened 2019-10-21
Graph-based language for semantic hypotheses

gabbard:


gabbard:
There are some outstanding issues with this (#399, #400, #401) but I think it is complete enough people can start playing with it.

adam/perception/_matcher.py
```diff
@@ -0,0 +1,862 @@
+# -*- coding: utf-8 -*-
```
gabbard: @denizbeser and @LichMaster98 : Since the code in this module is taken largely from NetworkX and is quite complex, you do not need to review it.

adam/perception/perception_graph.py
```diff
@@ -0,0 +1,1005 @@
+r"""
```
gabbard: This file needs close attention.

tests/perception/perception_graph_test.py
```diff
@@ -0,0 +1,164 @@
+from itertools import chain
```
gabbard: and also this one.

PR #391, opened 2019-10-22
Pursuit object recognition curriculum + with noise

### LINKED ISSUE #381: Make test curriculum for Pursuit - similar to the one described in the paper
Also add a parametrized noise to the test, which would vary the precision of test examples (e.g. 
hearing "cat" without a cat in the scene)

denizbeser:
closes #381

### DIFF
--- a/adam/curriculum/phase1_curriculum.py
+++ b/adam/curriculum/phase1_curriculum.py
@@ -8,7 +8,6 @@ from typing import Iterable
 from more_itertools import flatten
 
 from adam.axes import HorizontalAxisOfObject, FacingAddresseeAxis, AxesInfo
-from adam.curriculum import ExplicitWithSituationInstanceGroup
 from adam.curriculum.curriculum_utils import (
     phase1_instances,
     PHASE1_CHOOSER,
@@ -16,6 +15,7 @@ from adam.curriculum.curriculum_utils import (
     standard_object,
     GROUND_OBJECT_TEMPLATE,
 )
+from adam.curriculum.pursuit_curriculum import _make_simple_pursuit_curriculum
 from adam.language_specific.english.english_language_generator import (
     PREFER_DITRANSITIVE,
     USE_ADVERBIAL_PATH_MODIFIER,
@@ -86,10 +86,6 @@ from adam.ontology.phase1_ontology import (
     COLOR,
     PHASE_1_CURRICULUM_OBJECTS,
     is_recognized_particular,
-    BALL,
-    CHAIR,
-    TABLE,
-    DOG,
 )
 from adam.ontology.phase1_spatial_relations import (
     AWAY_FROM,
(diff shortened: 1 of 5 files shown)


tests/curriculum/phase1_curriculum_test.py
```diff
@@ -149,3 +150,7 @@ def test_come_curriculum():
 
 def test_behind_in_front_curriculum():
     curriculum_test(_make_behind_in_front_curriculum())
+
+
+def test_simple_pursuit_curriculum():
+    curriculum_test(_make_simple_pursuit_curriculum())
```
LichMaster98: Should we test curriculum generation with noise as well?

adam/curriculum/phase1_curriculum.py
```diff
@@ -1559,35 +1565,110 @@ def make_behind_in_front_templates() -> Iterable[Phase1SituationTemplate]:
...
+) -> Phase1InstanceGroup:
+    if n_noise > n_exemplar:
+        raise RuntimeError("Cannot have more noise than regular exemplars")
+
+    target_objects = [BALL, CHAIR, PERSON, TABLE, DOG, BIRD, BOX]
+    target_object_variables = [
+        object_variable(target.handle + "-target", target) for target in target_objects
+    ]
+    other_objects = [
+        object_variable("obj-" + str(idx)) for idx in range(n_objects_in_scene - 1)
```
LichMaster98: This notation will pull liquids and body parts as possible other objects. Prefer `standard_object` if we want to exclude those object types.

adam/curriculum/phase1_curriculum.py
```diff
@@ -1559,35 +1565,107 @@ def make_behind_in_front_templates() -> Iterable[Phase1SituationTemplate]:
...
+                )
+
+    description = f"simple_pursuit_curriculum_examples-{n_exemplar}_objects-{n_objects_in_scene}_noise-{n_noise}"
+    final_instance_group: Phase1InstanceGroup = ExplicitWithSituationInstanceGroup(
+        description, all_instances
+    )
+    return final_instance_group
+
+
 GAILA_PHASE_1_CURRICULUM = [
```
gabbard: Am I misreading this, or does this *replace* the normal phase 1 curriculum with the pursuit curriculum?

adam/curriculum/phase1_curriculum.py
```diff
@@ -1559,35 +1565,107 @@ def make_behind_in_front_templates() -> Iterable[Phase1SituationTemplate]:
     )
 
 
+# Simple Pursuit curriculum
+def _make_simple_pursuit_curriculum(
+    n_exemplar: int = 10, n_noise: int = 0, n_objects_in_scene: int = 3
```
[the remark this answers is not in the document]
gabbard: I see form further reading its actually the number of noisy instances, I think?

adam/curriculum/phase1_curriculum.py
denizbeser: I'd commented those out for faster testing - thank you for catching!

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def make_simple_pursuit_curriculum(
+    *,
+    num_instances: int = 10,
+    num_noise_instances: int = 0,
+    num_objects_in_instance: int = 3,
+) -> Phase1InstanceGroup:
+    """
+    Creates a Pursuit-learning curriculum with for a set of standard objects.
```
gabbard: Maybe add something like:

```
Each instance in the curriculum is a set of *num_objects_in_instance* objects paired with a word.  
We say an instance is non-noisy if the word refers to one of the objects in the set.
An instance is noisy if none of the objects correspond to the word.
For each type of object of interest, we will generate *num_instances_per_object_type* instances,
of which *num_noise_instances_per_object_type* will be noisy.
```

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+        num_objects_in_instance: number of objects in each instance
+    """
+    if num_noise_instances > num_instances:
+        raise RuntimeError("Cannot have more noise than regular exemplars")
+
+    target_objects = [BALL, CHAIR, PERSON, TABLE, DOG, BIRD, BOX]
+    target_object_variables = [
+        object_variable(target.handle + "-target", target) for target in target_objects
+    ]
+    other_objects = [
```
gabbard: Perhaps call this `noise_object_variables`. Also note that nothing restricts the relationship between noise object variables and target object variables, which could result in a noise instance accidentally being non-noisy by a `noise_object_variable` accidentally selecting the right thing, but that we are not going to worry about that case.

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+            "simple_pursuit",
+            salient_object_variables=[target_object_variable],
+            background_object_variables=other_objects[:-1],
+        )
+        for target_object_variable in target_object_variables
+    ]
+
+    # A generic template that is used to replace situations and perceptions (not dependency trees) in noise instances
+    noise_template = Phase1SituationTemplate(
+        "simple_pursuit",
```
gabbard:
```suggestion
        "simple_pursuit-noise",
```

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+        salient_object_variables=[other_objects[0]],
+        background_object_variables=other_objects[1:],
+    )
+
+    all_instances = []
+    # Generate phase_1 instance groups for each template (i.e each target word)
+    for object_is_present_template in object_is_present_templates:
```
gabbard: I think it would be clearer to just loop over the `target_objects` and instantiate the `target_object_variable` and `object_is_present` template there, rather than generating lots of templates and then iterating over them.

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+        new_instances = list(
```
gabbard:
```suggestion
        non_noise_instances = list(
```

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+                    object_is_present_template,
+                    max_to_sample=num_instances - num_noise_instances,
+                    chooser=PHASE1_CHOOSER,
+                    ontology=GAILA_PHASE_1_ONTOLOGY,
+                ),
+            ).instances()
+        )
+        all_instances.extend(new_instances)
+        # Create instances for noise
+        if num_noise_instances > 0:
```
gabbard: I think you can skip this check and let the `max_to_sample` parameter to `phase1_instances` take care of it?

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+            noise_instances = phase1_instances(
+                "simple_pursuit_curriculum",
+                sampled(
+                    noise_template,
+                    max_to_sample=num_noise_instances,
+                    chooser=PHASE1_CHOOSER,
+                    ontology=GAILA_PHASE_1_ONTOLOGY,
+                ),
+            ).instances()
+            for noise_instance in noise_instances:
```
gabbard: Python's tuple de-structuring can make this clearer:

```
for (situation, linguistic_description, perception) in noise_instances:
    ...
```

Also comment in this section describing why this is necessary.

adam/curriculum/pursuit_curriculum.py
denizbeser: `max_to_sample` throws an error if the number of samples is 0, so I had to check for this.

gabbard: @denizbeser : I think that's too restrictive; can you change it to allow 0?

denizbeser: @gabbard Done!

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+    num_objects_in_instance: int = 3,
+) -> Phase1InstanceGroup:
+    """
+    Creates a Pursuit-learning curriculum with for a set of standard objects. Each instance in the curriculum is a set
+    of *num_objects_in_instance* objects paired with a word.
+    We say an instance is non-noisy if the word refers to one of the objects in the set.
+    An instance is noisy if none of the objects correspond to the word.
+    For each type of object of interest, we will generate *num_instances_per_object_type* instances,
+    of which *num_noise_instances_per_object_type* will be noisy.
+    Args:
```
gabbard: Given the above, you can probably omit the `Args` list.

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+                noise_template,
+                max_to_sample=num_noise_instances,
+                chooser=PHASE1_CHOOSER,
+                ontology=GAILA_PHASE_1_ONTOLOGY,
+            ),
+        ).instances()
+        for (situation, _, perception) in noise_instances:
+            # For each instance to be replaced by a noisy instance, we keep the correct utterance, but
+            # replace the situation and perception. We do this to test the model's tolerance to varying degrees of
+            # noise that reflects noisy instances in the real-world.
```
gabbard: This is a bit confusing, because what is being kept is the situation and perception, and what is not being kept is the utterance.  Perhaps better would be:

```
# A noise instance needs to have the word for our target object
# while not actually having our target object be present.
# However, our language generator can't generate irrelevant language for a situation.
# Therefore, we generate the instance as normal above,
# but here we replace its linguistic description with the word for the target object.
```

adam/curriculum/pursuit_curriculum.py
```diff
@@ -0,0 +1,118 @@
...
+            linguistic_description = PHASE1_CHOOSER.choice(non_noise_instances)[1]
```
gabbard: If I am reading this correctly, the linguistic descriptions for the non-noise instances for a given object type are always the same (the name of the object), so choosing randomly is a bit of a red herring.  I think you could put this outside this `for` loop:

```
# [1] is the index of the linguistic description in an instance
# It doesn't matter which non-noise instance is chosen
# because they all have the object type name as their linguistic description.
target_object_name = first(non_noise_instances)[1]
```

and then here you can do

```
linguistic_description = target_object_name
```

adam/situation/templates/phase1_templates.py
```diff
@@ -273,7 +273,6 @@ def sampled(
     """
     Gets *max_to_sample* instantiations of *situation_template* with *ontology*
     """
```
gabbard: We still want to check it is non-negative

Issue #395, opened 2019-10-23
Fix link to index.html to be relative


LichMaster98:
I'm doing this really quickly to make the exports to send out to Charles & Mitch.

PR #396, opened 2019-10-23
395 index link fix

LichMaster98:
Closes https://github.com/isi-vista/adam/issues/395
The following zip can be used to verify the links are now relative.
[html.zip](https://github.com/isi-vista/adam/files/3764688/html.zip)

### DIFF
--- a/parameters/html/curriculum_to_html.sorted-by-utterance-length.params
+++ b/parameters/html/curriculum_to_html.sorted-by-utterance-length.params
@@ -1,0 +1,6 @@
+_includes:
+    - ../root.params
+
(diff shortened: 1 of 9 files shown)


parameters/html/curriculum_to_html.sorted-by-utterance-length.params
```diff
@@ -0,0 +1,6 @@
+_includes:
+    - ../root.params
+
+sort_by_utterance: True
+random_seed: 1
+curriculum: "phase1"
```
elizlee: Shouldn't the `output_directory` parameter also be here?

PR #397, opened 2019-10-24
392 nested geons

### LINKED ISSUE #392: Allow for rendering of nested geons
Previously, more complicated representations (i.e. people) weren't being picked up on while naively iterating over all `ObjectPerception`s within a scene.

Fixing this is a precondition for handling positioning and scaling of these kinds of nested geons.

chris-jenkins-isi:
Closes #392 original issue was due to some of the 'people' in scenes not having any properties attached to their arms/legs/etc (no color, no animacy, etc), not due to any nesting issues.

The new method of traversing all objects in a scene to find objects with geons gets around the issue of some objects not having any properties.

At any rate, the goal with this PR is to come up with a nested representation for scene objects that are `partOf` others so that later on they can be positioned/oriented/scaled relative to their parent object.

### DIFF
--- a/adam/visualization/__init__.py
+++ b/adam/visualization/__init__.py
@@ -4,7 +4,9 @@
         make_scenes.py:
             Converts perceptual representations of scenes into 3d rendered scenes.
 
-            Requires a parameters file to be run.
+            Currently this visualizes only a hardcoded set of scenes
+
+            Requires a parameters file to be run. See parameters/make_scenes.params for an example
 
     Other modules:
         panda3d_interface.py:
--- a/adam/visualization/__init__.py
+++ b/adam/visualization/__init__.py
@@ -1,0 +1,21 @@
+""" Visualization module: Code for producing more human-friendly representations of scenes from curricula.
+
+    Main executable:
+        make_scenes.py:
+            Converts perceptual representations of scenes into 3d rendered scenes.
+
+            Requires a parameters file to be run.
+
+    Other modules:
+        panda3d_interface.py:
+            Interface for the Panda3D rendering engine. Responsible for creating, destroying, positioning, coloring, etc
+            objects that are arranged in 3D space to represent perceptions of scenes.
+
+        positioning.py:
+            Module defining a Stochastic Gradient Descent model for arranging objects in a scene. Responsible for
+            translating relative layout requirements into a concrete layout.
+
+        utils.py:
+            Miscellaneous type definitions used by multiple visualization modules.
+
+"""
--- a/adam/visualization/panda3d_interface.py
+++ b/adam/visualization/panda3d_interface.py
@@ -129,7 +129,9 @@ class SituationVisualizer(ShowBase):
         # top level:
         if parent is None:
             if name in self.geo_nodes:
-                raise RuntimeError("Model names need to be unique")
+                raise RuntimeError(
+                    f"Error using name {name}: Model names need to be unique"
+                )
             self.geo_nodes[new_model.name] = new_model
             new_model.reparentTo(self.render)
         # nested
@@ -150,7 +152,9 @@ class SituationVisualizer(ShowBase):
         if parent is None:
             new_node.reparentTo(self.render)
             if name in self.geo_nodes:
-                raise RuntimeError("Model names need to be unique")
+                raise RuntimeError(
+                    f"Error using name {name}: Model names need to be unique"
+                )
             self.geo_nodes[new_node.name] = new_node
         else:
             new_node.reparentTo(parent)
--- a/adam/visualization/panda3d_interface.py
+++ b/adam/visualization/panda3d_interface.py
@@ -27,7 +27,7 @@ from panda3d.core import NodePath  # pylint: disable=no-name-in-module
 from panda3d.core import TextNode  # pylint: disable=no-name-in-module
 
 from direct.gui.OnscreenText import OnscreenText  # pylint: disable=no-name-in-module
-
+from adam.visualization.positioning import PositionsList
 from adam.visualization.utils import Shape
 
 from adam.perception.developmental_primitive_perception import RgbColorPerception
@@ -52,14 +52,14 @@ class SituationVisualizer(ShowBase):
 
         plight = PointLight("pointLight")
         plight_node = self.render.attachNewNode(plight)
-        plight_node.setPos(10, 20, 0)
+        plight_node.setPos(10, 20, 1)
         self.render.setLight(plight_node)
 
         # self.render is the top node of the default scene graph
 
         self.ground_plane = self._load_model("ground.egg")
         self.ground_plane.reparentTo(self.render)
-        self.ground_plane.setPos(0, 0, -2)
+        self.ground_plane.setPos(0, 0, 0)
         m = Material()
         m.setDiffuse((255, 255, 255, 255))
         # the "1" argument to setMaterial is crucial to have it override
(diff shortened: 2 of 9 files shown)


adam/visualization/positioning.py
```diff
@@ -20,50 +20,62 @@
...
-    ball = AdamObject(name="ball")
-    box = AdamObject(name="box")
+def run_model(
+    objs: List[AdamObject], num_iterations: int = 200, yield_steps: Optional[int] = None
+) -> Generator[List[torch.Tensor], None, List[torch.Tensor]]:
+    """
+    Construct a positioning model given a list of objects to position, return their position values.
+    Args:
+        objs: list of AdamObjects requested to be positioned
+        num_iterations: total number of SGD iterations.
```
gabbard:
```suggestion
        *num_iterations*: total number of SGD iterations.
```

(and similarly elsewhere)

adam/visualization/positioning.py
```diff
@@ -20,50 +20,62 @@
...
+        yield_steps: If provided, the current positions of all objects will be returned after this many steps
 
-    cardboard_box = AdamObject(name="cardboardBox")
-    aardvark = AdamObject(name="aardvark")
-    flamingo = AdamObject(name="flamingo")
+    Returns: List of (3,) tensors corresponding to the positions of the objs list
```
gabbard: Would a map from `AdamObjects` to their positions be simpler and less error-prone for the caller?

adam/visualization/positioning.py
```diff
@@ -80,13 +92,33 @@ class AxisAlignedBoundingBox:
     corners at (-1, -1, -1) and (1, 1, 1), giving the box a volume of 2(^3)
     """
 
-    center: torch.Tensor = attrib()  # tensor shape: (3,)
+    center: Parameter = attrib()  # tensor shape: (3,)
```
gabbard: a `Parameter` is a `torch.Tensor` - why change it to be more restricted? Couldn't you conceivable have a static bounding box you are not optimizing over?

adam/visualization/positioning.py
```diff
@@ -80,13 +92,33 @@ class AxisAlignedBoundingBox:
...
     scale: torch.Tensor = attrib()  # tensor shape: (3, 3) - diagonal matrix
     # rotation: torch.Tensor = attrib()
 
     def center_distance_from_point(self, point: torch.Tensor) -> torch.Tensor:
         return torch.dist(self.center, point, 2)
 
+    def distance_from_lowest_corner_to_ground(self) -> torch.Tensor:
+        """
+        Returns distance of z value of minimum (w/r/t Z coordinate) corner of the box from origin's Z value.
```
gabbard: I find this a little hard to read. How about:

"""
Get the position of the lowest corner of the box.

The third coordinate is interpreted as the height relative to the ground at z=0.
"""

Now that I look at this, this method should be called something like `z_coordinate_of_lowest_corner` because it doesn't return the distance to the ground if the lowest corner is below the ground.

adam/visualization/positioning.py
```diff
@@ -80,13 +92,33 @@ class AxisAlignedBoundingBox:
...
+        Returns: (1,) tensor
+
+        """
+        corners = self.get_corners()
+        min_corner_z = torch.min(
```
gabbard: This complex expression needs more comments to be clear to a reader (or pull out sub-expressions with descriptive names)

adam/visualization/positioning.py
```diff
@@ -272,34 +320,76 @@ def for_objects(
...
+        else:
+            return -distance_above_ground
+
+
+class WeakGravityPenalty(nn.Module):  # type: ignore
+    """
+    Model that penalizes boxes that are not resting on the ground.
+    """
+
+    # TODO: exempt birds from this constraint
```
gabbard: Make an issue and reference it in the comment.

adam/visualization/positioning.py
```diff
@@ -272,34 +320,76 @@ def for_objects(
...
+    # TODO: exempt things resting on top of other objects from this constraint
```
gabbard: I don't think they need to be exempted - we'll let the non-collision constraint hold them up (just like real-life).

adam/visualization/panda3d_interface.py
```diff
@@ -60,7 +59,7 @@ def __init__(self) -> None:
 
         self.ground_plane = self._load_model("ground.egg")
         self.ground_plane.reparentTo(self.render)
-        self.ground_plane.setPos(0, 0, -1)
+        self.ground_plane.setPos(0, 0, -2)
```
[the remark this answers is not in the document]
chris-jenkins-isi: Screen Shot 2019-12-11 at 15 51 56It should be -1 not -2, but it is offset like that so that a 2x2x2 cube positioned with its center point at (0,0,0) rests on the ground. Maybe that isn't the best assumption to make, but that's the idea.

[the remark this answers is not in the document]
gabbard: Can you alter it to have the ground plane at `z=0`? It is much less confusing that way, and also fits the "gravity" in our positioning model.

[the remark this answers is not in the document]
chris-jenkins-isi: Yes. I might have moved it there however long ago not understanding that the plane was changing color based on whether it was above a point light that was (until the upcoming changes) sitting at 0 in the Z dimension. In the screenshot above, the plane is unlit because it is blocking the light source!

adam/visualization/panda3d_interface.py
```diff
@@ -102,27 +101,66 @@ def add_model(
...
     def clear_scene(self) -> None:
         """Clears out all added objects (other than ground plane, camera, lights)"""
         for node in self.geo_nodes:
             node.remove_node()
         self.geo_nodes = []
 
+    def top_level_positions(self) -> List[Tuple[float, float, float]]:
+        """Returns a list of all positions of top level nodes of geometry objects
+           (so not cameras and lights). """
+        # TODO: need to ensure this is deterministic
```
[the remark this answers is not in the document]
chris-jenkins-isi: It's retrieving values from a list managed in this file, so determinism isn't a worry. The only design caveat would be that modifying the scene in between calling `top_level_positions()` and `set_positions(new_positions: List[torch.Tensor]` could invalidate the correspondence between these positions and objects in the scene.

adam/visualization/positioning.py
chris-jenkins-isi: This was one of those mypy-motivated changes. It complains about calling the model's register_parameter function with a Tensor instead of a Parameter

gabbard: Okay. The really correct way to handle this would be to make `AxisAlignedBoundingBox` a generic class with the center point type as a type parameter upper-bounded by `Tensor`.  But you can leave it as is until we need it to be more flexible.

adam/visualization/panda3d_interface.py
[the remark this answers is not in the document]
gabbard: A good reason to work with maps and not aligned lists. :-)

adam/visualization/panda3d_interface.py
```diff
@@ -100,51 +97,83 @@ def __init__(self) -> None:
...
             print(f"No geometry found for {model_type}")
             raise
-        self.geo_nodes.append(new_model)
-        new_model.reparentTo(self.render)
-        new_model.setPos(pos[0], pos[1], pos[2])
-        new_model.setColor((col.red / 255, col.green / 255, col.blue / 255, 1.0))
+        # top level:
+        if parent is None:
+            if name in self.geo_nodes:
+                raise RuntimeError("Model names need to be unique")
```
gabbard: Include the duplicated name. You will appreciate it when debugging. :-)

adam/visualization/__init__.py
```diff
@@ -0,0 +1,21 @@
+""" Visualization module: Code for producing more human-friendly representations of scenes from curricula.
+
+    Main executable:
+        make_scenes.py:
+            Converts perceptual representations of scenes into 3d rendered scenes.
```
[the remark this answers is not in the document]
chris-jenkins-isi: I removed that positioning test as an entry point. Maybe it would be better to copy it over to a test file?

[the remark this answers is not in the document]
gabbard: I think maintaining that entry point is useful for testing.

PR #410, opened 2019-10-30
Generate perception patterns from perception graphs

### LINKED ISSUE #407: Create PerceptionGraphPatterns directly from PerceptionGraphs
It occurs to me that to support #405  we will need the capability to go from a `PerceptionGraph` to a minimal matching `PerceptionGraphPattern`. This capability does not currently exist; instead we can go from an `ObjectStructuralSchema` to a perception graph pattern.  

I suggest adapting the current `ObjectStructuralSchema` translation code to instead translate from any `PerceptionGraph`.  Then the current `from_schema` can just instantiate an object for that schema and translate the resulting perception.

I suggest setting up individual unit tests for matching each object which currently matches correctly before you start working on this. That way you can test the change doesn't break anything.

denizbeser:
In this implementation we remove unwanted nodes from graphs created from schemas before creating patterns from those graphs (colors, property assertions, regions, ground, and learner) - it's worth double checking that's the intended behavior, and that we don't remove too much.

closes #407

### DIFF
--- a/adam/perception/perception_graph.py
+++ b/adam/perception/perception_graph.py
@@ -18,7 +18,7 @@ from attr.validators import instance_of, optional
 from immutablecollections import immutabledict, immutableset
 from immutablecollections.converter_utils import _to_immutabledict, _to_tuple
 from more_itertools import first
-from networkx import DiGraph, isolates
+from networkx import DiGraph
 
 from adam.axes import AxesInfo, HasAxes
 from adam.axis import GeonAxis
@@ -27,7 +27,7 @@ from adam.ontology import OntologyNode
 from adam.ontology.phase1_ontology import PART_OF, GAILA_PHASE_1_ONTOLOGY
 from adam.ontology.phase1_spatial_relations import Direction, Distance, Region
 from adam.ontology.structural_schema import ObjectStructuralSchema
-from adam.perception import ObjectPerception, GROUND_PERCEPTION, LEARNER_PERCEPTION
+from adam.perception import ObjectPerception
 from adam.perception._matcher import GraphMatching
 from adam.perception.developmental_primitive_perception import (
     DevelopmentalPrimitivePerceptionFrame,
@@ -37,6 +37,8 @@ from adam.perception.developmental_primitive_perception import (
 )
 from adam.perception.high_level_semantics_situation_to_developmental_primitive_perception import (
     HighLevelSemanticsSituationToDevelopmentalPrimitivePerceptionGenerator,
+    EXCLUDE_LEARNER,
+    EXCLUDE_GROUND,
 )
 from adam.random_utils import RandomChooser
 from adam.situation import SituationObject
@@ -341,44 +343,23 @@ class PerceptionGraphPattern:
         perception_generator = HighLevelSemanticsSituationToDevelopmentalPrimitivePerceptionGenerator(
             GAILA_PHASE_1_ONTOLOGY
         )
+        # We explicitly exclude ground and learner in perception generation, which were not specified in the schema
         perception = perception_generator.generate_perception(
-            situation, chooser=RandomChooser.for_seed(0)
+            situation,
+            chooser=RandomChooser.for_seed(0),
+            flags=[EXCLUDE_GROUND, EXCLUDE_LEARNER],
         )
         perception_graph = PerceptionGraph.from_frame(
             first(perception.frames)
         ).copy_as_digraph()
 
-        # We remove certain information that is added by situation generation, but is not in schema
-        # Removed in order: colors, ontology nodes (properties), regions related to ground, ground, learner and
-        # finally any leftover islands that dont belong to the schema's graph component
+        # We remove color and properties that is added by situation generation, but is not in schema
         nodes_to_remove = [
             node
-            for node in perception_graph.nodes
-            if isinstance(node, RgbColorPerception)
-            or (isinstance(node, OntologyNode))
-            or (
-                isinstance(node, tuple)
-                and isinstance(node[0], Region)
-                and any(
-                    isinstance(n, ObjectPerception)
-                    and n.debug_handle == GROUND_PERCEPTION.debug_handle
-                    for n in perception_graph.neighbors(node)
-                )
-            )
-            or (
-                isinstance(node, ObjectPerception)
-                and node.debug_handle == GROUND_PERCEPTION.debug_handle
-            )
-            or (
-                isinstance(node, ObjectPerception)
-                and node.debug_handle == LEARNER_PERCEPTION.debug_handle
-            )
+            for node in perception_graph
+            if isinstance(node, RgbColorPerception) or isinstance(node, OntologyNode)
         ]
         perception_graph.remove_nodes_from(nodes_to_remove)
-        # After removing ground and learner, we are still left with some nodes that belonged to their components.
-        # We remove these islands, effectively keeping the connected component containing the root object perception.
-        islands = list(isolates(perception_graph))
-        perception_graph.remove_nodes_from(islands)
 
         # Finally, we convert the PerceptionGraph DiGraph representation to a PerceptionGraphPattern
         return PerceptionGraphPattern.from_graph(perception_graph=perception_graph)
--- a/adam/perception/perception_graph.py
+++ b/adam/perception/perception_graph.py
@@ -349,7 +349,8 @@ class PerceptionGraphPattern:
         ).copy_as_digraph()
 
         # We remove certain information that is added by situation generation, but is not in schema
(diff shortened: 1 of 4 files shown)


adam/perception/perception_graph.py
```diff
@@ -285,6 +296,10 @@ def _to_dot_node(
 
         return node_id
 
+    @property
+    def graph(self):
```
gabbard: I would shift this towards the top of the class so a reader will spot it before the long methods below.  Also, I suggest renaming it to `copy_as_digraph()` and returning a copy so that the immutability of the object itself is not compromised.

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
-
-        PerceptionGraphPattern._translate_schema(
-            # since the root object does not itself have a `SubObject` object,
-            # we wrap it in one.
-            object_=SubObject(object_schema),
-            object_schema=object_schema,
-            graph=graph,
-            schema_node_to_pattern_node=object_to_node,
+
+        # Create a PerceptionGraph from schema
```
gabbard: You might spell out your strategy a bit more, e.g.

```
# First, we generate a PerceptionGraph corresponding to this schema
# by instantiating a situation which contains a single object
# of the desired type.
```

gabbard: Then add similar comments for the other major steps

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+            if isinstance(node, RgbColorPerception)
+            or (
+                isinstance(node, ObjectPerception)
+                and node.debug_handle == GROUND_PERCEPTION.debug_handle
+            )
+            or (
+                isinstance(node, ObjectPerception)
+                and node.debug_handle == LEARNER_PERCEPTION.debug_handle
+            )
+            or (isinstance(node, tuple) and isinstance(node[0], Region))
```
gabbard: We don't want to remove all `Regions` - just regions which reference the ground or the learner. `Regions` can describe the relationships of the parts of an object to one another

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+            or (isinstance(node, OntologyNode))
+        ]
+
+        perception_graph.graph.remove_nodes_from(nodes_to_remove)
+        islands = list(isolates(perception_graph.graph))
```
gabbard: See if we still have islands after holding on to `Region`s

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+            )
+            or (
+                isinstance(node, ObjectPerception)
+                and node.debug_handle == LEARNER_PERCEPTION.debug_handle
+            )
+            or (isinstance(node, tuple) and isinstance(node[0], Region))
+            or (isinstance(node, OntologyNode))
+        ]
+
+        perception_graph.graph.remove_nodes_from(nodes_to_remove)
```
gabbard: I think we want to keep `PerceptionGraph` as immutable, but you can get a copy of the underlying NetworkX `DiGraph`, modify it, and make a new `PerceptionGraph`

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+        pattern_graph = DiGraph()
+        PerceptionGraphPattern._translate_graph(
+            perception_graph=perception_graph.graph, pattern_graph=pattern_graph
+        )
+        return PerceptionGraphPattern(pattern_graph)
+
+    @staticmethod
+    def _translate_graph(perception_graph: DiGraph, pattern_graph: DiGraph) -> None:
+        perception_node_to_pattern_node: Dict[Any, "NodePredicate"] = {}
+        # perception graph -> perception graph pattern
```
gabbard: Stray comment?

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+                            key
+                        ] = RegionPredicate.matching_distance(node)
+                elif isinstance(node, GeonAxis):
+                    perception_node_to_pattern_node[key] = AxisPredicate.from_axis(node)
+                elif isinstance(node, ObjectPerception):
+                    perception_node_to_pattern_node[key] = AnyObjectPerception(
                         debug_handle=node.debug_handle
                     )
+                elif isinstance(node, OntologyNode):
+                    perception_node_to_pattern_node[key] = IsPropertyNodePredicate(node)
```
gabbard: Let's rename this predicate to `IsOntologyNode`.

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
+
+            key = node
+            if key not in perception_node_to_pattern_node:
+                if isinstance(node, tuple):
+                    node = node[0]
+                    if isinstance(node, Geon):
+                        perception_node_to_pattern_node[
+                            key
+                        ] = GeonPredicate.exactly_matching(node)
+                    elif isinstance(node, Region):
```
gabbard: Add an `else` which throws an exception. That way, if someone adds something we don't expect, we will find out about it quickly.

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,113 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
-                    "Did not expect a negated relation in a sub-object schema"
+        for original_node in perception_graph.nodes:
+            # Add each node
+            pattern_node = map_node(original_node)
+            pattern_graph.add_node(pattern_node)
+
+        # Once all nodes are translated, add all edges from each node
+        for original_node in perception_graph.nodes:
+            edges_from_node = perception_graph.edges(original_node, data=True)
+            for original_edge in edges_from_node:
```
gabbard: nit/style:  Integer indexing into tuples is hard to read, so prefer to iterate as

```python
for (source_node, dest_node, edge_data) in edges_from_node:
```

tests/perception/perception_graph_test.py
```diff
@@ -37,6 +40,17 @@ def test_house_on_table():
     assert do_object_on_table_test(HOUSE, _HOUSE_SCHEMA, BIRD)
 
 
+def test_objects_individually():
+    for object_ in GAILA_PHASE_1_ONTOLOGY.nodes_with_properties(
+        INANIMATE_OBJECT, banned_properties=[LIQUID, IS_BODY_PART]
+    ):
+        if object_ not in [BIRD, TABLE, CHAIR, TRUCK, CAR]:
```
gabbard: Comment that these are known failures

gabbard: Why not stick `GROUND` in this list?

adam/perception/perception_graph.py
denizbeser: We get islands related to axes, and it somehow causes matches() code to go into an infinite loop, within `subgraph_isomorphisms_iter()`

gabbard: Can you show me the "filtered" and "unfiltered" versions of a graph with axis islands?  It has to be happening due to over-aggressive filtering.

adam/perception/perception_graph.py
```diff
@@ -313,111 +328,132 @@ def from_schema(object_schema: ObjectStructuralSchema) -> "PerceptionGraphPatter
...
         )
-        return PerceptionGraphPattern(graph)
+        perception_graph = PerceptionGraph.from_frame(
+            first(perception.frames)
+        ).copy_as_digraph()
+
+        # We remove certain information that is added by situation generation, but is not in schema
+        # Removed in order: colors, ontology nodes (properties), regions related to ground, ground, learner and
+        # finally any leftover islands that dont belong to the schema's graph component
+        nodes_to_remove = [
```
gabbard: Actually, can we simplify this whole things to:

* delete everything not reachable (taking into account edge direction) from the root object perception
* From that, delete colors.

adam/perception/perception_graph.py
gabbard: I don't think this comment has been addressed?

PR #416, opened 2019-11-05
Introduce Subset Learning of Static Prepositions

### LINKED ISSUE #406: Apply the subset learner to learn preposition semantics
I propose the following:

- for now, assume the learner already knows all the objects. You can do this using `PerceptionGraphPattern.from_schema` to build patterns from the object schemata.
- when you encounter a situation, try to match the object patterns against it.  If they match, add new nodes to the perception graph associated with the object name.  Make an edge from each node in the matched graph to that node (of some type like "matchedPattern"). Then have that node inherit all relations any portion of the pattern match had with something outside the pattern match,
- take the candidate semantics of a preposition between two objects to be the subgraph of the perception graph on the path between the two "matched object pattern" nodes.  
- try applying the sub-set algorithm to narrow down the semantics.

A few things to bear in mind:
- the semantics of some prepositions involve the relationship of various objects axes to the learner or to the speaker or addressee, so you might need to incorporate some representation of those even when it doesn't like on the path between the object nodes.
- I also think the current perception graph translation does not incorporate that information, which is stored in the `axes_info` field of the perception. If you confirm this is not being added, make a new issue and go ahead and add it before proceeding with this.
- you might encounter the need to "intersect" the semantics of particular pattern predicates; you could do this by adding an `intersect` method to them.  Note that this may require changes in the implementation (e.g. if it is currently implemented as wrapping a reference object)

LichMaster98:
Closes #406


adam/perception/perception_graph.py
```diff
@@ -312,14 +334,12 @@ class PerceptionGraphPattern:
 
     _graph: DiGraph = attrib(validator=instance_of(DiGraph))
 
-    def matcher(
-        self, graph_to_match_against: PerceptionGraph
-    ) -> "PerceptionGraphPatternMatching":
+    def matcher(self, graph_to_match_against: PerceptionGraph) -> "PatternMatching":
```
gabbard:
```suggestion
    def matcher(self, graph_to_match_against: _TargetGraphT) -> "PatternMatching[_TargetGraphT]":
```

adam/perception/perception_graph.py
```diff
@@ -510,9 +544,7 @@ class PerceptionGraphPatternMatching:
     pattern: PerceptionGraphPattern = attrib(
         validator=instance_of(PerceptionGraphPattern)
     )
-    graph_to_match_against: PerceptionGraph = attrib(
-        validator=instance_of(PerceptionGraph)
-    )
+    graph_to_match_against: TargetGraphT = attrib(validator=instance_of(TargetGraphT))
```
gabbard: You can't provide type variables to `instance_of`; the best you can do is test it is one of the two possible types.

gabbard:
@LichMaster98 : has this been rebased on top of `master`? (that will probably be easier if you disentangle what you have done from the current subset learner and make a separate "preposition subset learner")

LichMaster98:
@gabbard I am rebased on top of master, I have also separated out the `PrepositionSubSetLearner` into its own file now. Currently tracking down why the `intersection` of two `PrepositionPattern`s is trying to return a `PrepositionPattern` without the mappings to `MODIFIED` and `GROUND`.

adam/perception/perception_graph.py
```diff
@@ -996,6 +1244,20 @@ def dot_label(self) -> str:
     def exactly_matching(direction: Direction[Any]) -> "DirectionPredicate":
         return DirectionPredicate(direction)
 
+    def matches_predicate(
+        self,
+        source_predicate_node: NodePredicate,
+        edge_label: "EdgePredicate",
+        dest_predicate_node: NodePredicate,
+    ) -> bool:
+        if isinstance(edge_label, DirectionPredicate):
```
gabbard: style/nit: you could do
```
return isinstance(edge_label, DirectionPredicate) and (edge_label.reference_direction.positive
                == self.reference_direction.positive)
```

adam/perception/perception_graph.py
```diff
@@ -92,10 +122,36 @@ def increment(self, amount=1) -> None:
 """
 Edge label in a `PerceptionGraph` linking an `ObjectPerception` to its associated `Property`.
 """
+GOVERNED = OntologyNode("governed")
```
gabbard: Can you shift these to immediately before `PrepositionPattern`?

adam/perception/perception_graph.py
```diff
@@ -927,6 +1148,17 @@ def reverse_in_dot_graph(self) -> bool:
         """
         return False
 
+    @abstractmethod
+    def matches_predicate(
```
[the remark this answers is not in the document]
gabbard: If it is for matching pattern nodes against each other, some of the implementations below seem wrong. Please double-check them.

adam/perception/perception_graph.py
```diff
@@ -729,6 +872,9 @@ def __call__(self, graph_node: PerceptionGraphNode) -> bool:
     def dot_label(self) -> str:
         return "*"
 
+    def matches_predicate(self, predicate_node: "NodePredicate") -> bool:
+        return True
```
gabbard: Shouldn't this only match another `AnyNodePredicate`?

adam/utils/networkx_utils.py
```diff
@@ -27,3 +30,25 @@ def digraph_with_nodes_sorted_by(
         new_graph.add_edge(source, dest, **data)
 
     return new_graph
+
+
+def subgraph(graph: DiGraph, nodes: Iterable[Any]) -> DiGraph:
+    """
+    Get a Subgraph view of a a Digraph with node iteration order in a deterministic fashion
```
gabbard: Expand this to explain why it is necessary, and cite your source for the code you modified to make this.

adam/perception/perception_graph.py
```diff
@@ -60,8 +74,24 @@ def increment(self, amount=1) -> None:
 
 DebugCallableType = Callable[[DiGraph, Dict[Any, Any]], None]
 
+
+@attrs(frozen=True, eq=False, slots=True)
+class MatchedObjectNode:
```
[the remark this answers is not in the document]
LichMaster98: It was at the top of the module so that the type of `PerceptionGraphNode` could include it as a possible type. This type is defined immediately under the node.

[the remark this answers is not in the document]
gabbard: Got it. Can you add a module-level comment letting the reader know what are the core classes they should read first?

adam/perception/perception_graph.py
```diff
@@ -896,6 +1095,28 @@ def __call__(self, graph_node: PerceptionGraphNode) -> bool:
     def dot_label(self) -> str:
         return " & ".join(sub_pred.dot_label() for sub_pred in self.sub_predicates)
 
+    def matches_predicate(self, predicate_node: "NodePredicate") -> bool:
+        return all(
```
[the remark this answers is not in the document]
LichMaster98: We don't use this anywhere so I'll just raise a `NotImplementedError` for now

PR #419, opened 2019-11-06
405 subset learner with graphs

### LINKED ISSUE #405: Subset learner needs to be ported to use graph matching
The "subset learning" algorithm was only previously working because of a bug which caused axes objects to be duplicated (so, for example, all tires would have the same axes).  This bug was fixed in https://github.com/isi-vista/adam/pull/382 , so I have disable the subset learning test.

We should port the subset learning algorithm to use the new graph pattern matching code.  A challenge here is that while before when finding what is in common between two perceptual representations we just took an intersection of properties, now we need to find the maximum common sub-graph of two perception graphs.  This is a challenging NP-complete problem and I am not aware of any existing Python implementations.  

In practice, our node and edge predicates let us sharply prune the search space, and our graphs are not huge or pathological (our patterns are "almost trees"), so hopefully finding the common sub-graph can be efficient enough in practice at our scale.

Let's establish some terminology.  We will call an _object pattern_ a `PerceptionGraphPattern` which is rooted by an `AnyObjectPredicate` and includes other predicates for other object perceptions, axes, and properties which are reachable by following edges from the root object perception.  We want to keep relation edges between parts of this object but don't want to follow any edge which involved us in the sub-graph of another object.  We will term a union of object patterns which the learner currently believes to be the meaning of a description as a "hypothesis". Note that although the learner will initially have multiple object patterns in its hypothesis (if the initial scene it encounters for an object has multiple descriptions), we hope that it will rapidly reduce that to a single object pattern with more observations.

When the subset learner encounters a new instance with linguistic description `L` and perception graph `P`, it wants to grab the current hypothesis `H` associated with `L`.  Then it should try to find the maximal match of each object pattern in `H` against `P`.  It should then reduce that pattern to only the portion of it which can be found successfully; if a pattern has no match or matches only in the root object node it should be removed from the hypothesis altogether.

The trickiness comes with finding he partial match.  There is support built-in to the current graph matcher to provide a partial match on match failure using the "debug mode" of the matcher.  Try experimenting with this and if it works well we can promote it to a top-level, non-debug method.  I am pretty sure for the partial match to be useful you will need to disable look-ahead pruning.  It's not clear to me that the partial match returned is necessarily maximal, so you may need to determine what part of the pattern needs pruning, remove it, and then repeat the process until the pattern is either empty or successfully matches.

denizbeser:
Closes #405


gabbard:
@denizbeser : Can you rebase this onto `master` (you probably want to collapse your WIP commits down to one commit first; you could also just merge `master` in, if easier) and run the matching benchmark both on this branch and from `master` to see what the speed impact, if any, of your changes are?

tests/learner/learner_test.py
```diff
@@ -39,30 +36,57 @@ def test_subset_learner_ball():
...
+            test_instance_perception,
+        ) in test_instance_group.instances():
+            descriptions_from_learner = learner.describe(test_instance_perception)
+            gold = test_instance_language.as_token_sequence()
+            assert [desc.as_token_sequence() for desc in descriptions_from_learner][
+                0
+            ] == gold
+
+
+def test_subset_learner_dog_infinite_loop():
```
gabbard: Does `dog` still produce an infinite loop?  (or take excessively long?) If not we should remove this test.

denizbeser:
The updates in this PR didn't improve the benchmark results of matching runtime.

adam/learner/subset.py
```diff
@@ -1,19 +1,58 @@
...
+from adam.perception.perception_graph import PerceptionGraph, PerceptionGraphPattern
+
+
+def get_largest_matching_pattern(
+    pattern: PerceptionGraphPattern, graph: PerceptionGraph
+) -> PerceptionGraphPattern:
+    # Helper function to return the largest matching pattern for learner from a perception pattern and graph pair.
+    # Initialize matcher in debug version to keep largest subgraph
+    matching = pattern.matcher(graph)
+    debug_sink = {"a": 1}
```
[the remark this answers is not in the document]
denizbeser: Because line 557 in perception_graph matcher checks if a debug_sink is empty/none. I will change that part of the code.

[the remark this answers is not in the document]
gabbard: @denizbeser : fix it to just only for `None`.

adam/perception/perception_graph.py
```diff
@@ -538,7 +543,7 @@ def matches(
         )
         got_a_match = False
         for mapping in matching.subgraph_isomorphisms_iter(
-            debug=debug_mapping_sink is not None
+            debug=(debug_mapping_sink is not None) is not None
```
gabbard: I am confused by this line?

denizbeser: Thanks for catching! Must be the Pycharm autocorrect. I'm pushing the fix now.

PR #421, opened 2019-11-07
Add test for object axis facing addressee

elizlee:
WIP

### DIFF
--- a/tests/axes_test.py
+++ b/tests/axes_test.py
@@ -1,0 +1,59 @@
+from immutablecollections import immutableset
+
+from adam.axes import HorizontalAxisOfObject, AxesInfo, FacingAddresseeAxis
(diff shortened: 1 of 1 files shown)


gabbard:
@elizlee : can you remind me the status of this? Is it ready to be reviewed and merged?

elizlee:
@gabbard : I pushed this when I was working on finding a situation that would fail when we had that issue with the axis IDs. In its current state, this test would not be able to catch it if the issue came up again, but otherwise the code is fine...

gabbard:
@elizlee : can you elaborate on " In its current state, this test would not be able to catch it if the issue came up again"?  If the test wouldn't catch anything, should we close the PR?

elizlee:
@gabbard : I think we can close the PR.

PR #425, opened 2019-11-12
Graph match renderer for debugging purposes

### LINKED ISSUE #399: Trucks cannot be matched by their schemas using PerceptionGraphPattern
Run `test_inanimate_objects` to replicate the failure.

### LINKED ISSUE #418: Improve speed of perception graph pattern matching

denizbeser:
Closes #399 
Closes #418

### DIFF
--- a/adam/learner/subset.py
+++ b/adam/learner/subset.py
@@ -1,4 +1,4 @@
-from typing import Dict, Generic, Mapping, Tuple, Any
+from typing import Dict, Generic, Mapping, Tuple, Any, Optional
 
 from attr import Factory, attrib, attrs
 from immutablecollections import immutabledict
@@ -16,7 +16,11 @@ from adam.perception import PerceptionT, PerceptualRepresentation, ObjectPercept
 from adam.perception.developmental_primitive_perception import (
     DevelopmentalPrimitivePerceptionFrame,
 )
-from adam.perception.perception_graph import PerceptionGraph, PerceptionGraphPattern
+from adam.perception.perception_graph import (
+    PerceptionGraph,
+    PerceptionGraphPattern,
+    DebugCallableType,
+)
 
 
 @attrs
@@ -31,7 +35,7 @@ class SubsetLanguageLearner(
     _descriptions_to_pattern_hypothesis: Dict[
         Tuple[str, ...], PerceptionGraphPattern
     ] = attrib(init=False, default=Factory(dict))
-    _render_for_debug: bool = attrib(default=False, init=True)
+    _debug_callback: Optional[DebugCallableType] = attrib(default=None)
 
     def observe(
         self, learning_example: LearningExample[PerceptionT, LinguisticDescription]
@@ -64,7 +68,7 @@ class SubsetLanguageLearner(
             hypothesis_pattern_common_subgraph = get_largest_matching_pattern(
                 previous_pattern_hypothesis,
                 observed_perception_graph,
-                render_for_debug=self._render_for_debug,
+                debug_callback=self._debug_callback,
             )
             # Update the leading hypothesis
             self._descriptions_to_pattern_hypothesis[
@@ -105,7 +109,7 @@ class SubsetLanguageLearner(
             common_pattern = get_largest_matching_pattern(
                 pattern_hypothesis,
                 observed_perception_graph,
-                render_for_debug=self._render_for_debug,
+                debug_callback=self._debug_callback,
             )
             common_pattern_size = len(common_pattern.copy_as_digraph().nodes)
             if common_pattern_size > max_matching_subgraph_size:
@@ -123,14 +127,14 @@ def get_largest_matching_pattern(
     pattern: PerceptionGraphPattern,
     graph: PerceptionGraph,
     *,
-    render_for_debug: bool = False,
+    debug_callback: Optional[DebugCallableType] = None,
 ) -> PerceptionGraphPattern:
     """ Helper function to return the largest matching pattern for learner from a perception pattern and graph pair."""
     # Initialize matcher in debug version to keep largest subgraph
     matching = pattern.matcher(graph)
     debug_sink: Dict[Any, Any] = {}
     match_mapping = list(
-        matching.matches(debug_mapping_sink=debug_sink, render_for_debug=render_for_debug)
+        matching.matches(debug_mapping_sink=debug_sink, debug_callback=debug_callback)
     )
 
     if match_mapping:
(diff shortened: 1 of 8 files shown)


adam/learner/subset.py
```diff
@@ -31,6 +31,7 @@ class SubsetLanguageLearner(
     _descriptions_to_pattern_hypothesis: Dict[
         Tuple[str, ...], PerceptionGraphPattern
     ] = attrib(init=False, default=Factory(dict))
+    _render_for_debug = attrib(default=False, init=True)
```
gabbard: needs type annotation

adam/learner/subset.py
```diff
@@ -115,13 +120,17 @@ def describe(
 
 
 def get_largest_matching_pattern(
-    pattern: PerceptionGraphPattern, graph: PerceptionGraph
+    pattern: PerceptionGraphPattern,
+    graph: PerceptionGraph,
+    render_for_debug: bool = False,
```
gabbard: Make this kw-only

adam/perception/_matcher.py
```diff
@@ -248,6 +250,10 @@ def match(self, *, debug: bool = False):
         we yield the mapping.
 
         """
+        # Check rendering debug flag to see if we should render the graph
+        if debug_function is not None:
+            pass
```
gabbard: delete `pass`

adam/perception/_matcher.py
```diff
@@ -248,6 +250,10 @@ def match(self, *, debug: bool = False):
         we yield the mapping.
 
         """
+        # Check rendering debug flag to see if we should render the graph
+        if debug_function is not None:
```
gabbard: This can probably be hidden under the `if debug` conditional below so that we are spared the functional call overhead (significant in Python :-( ) when debug mode is not enabled.

adam/perception/perception_graph.py
```diff
@@ -561,6 +567,27 @@ def matches(
             debug_mapping_sink.clear()
             debug_mapping_sink.update(matching.debug_largest_match)
 
+    def rendering_debug_function(
```
gabbard: Pull the debug function and its counter into its own object (which implements `__call__`)

adam/learner/subset.py
```diff
@@ -31,6 +31,7 @@ class SubsetLanguageLearner(
     _descriptions_to_pattern_hypothesis: Dict[
         Tuple[str, ...], PerceptionGraphPattern
     ] = attrib(init=False, default=Factory(dict))
+    _render_for_debug: bool = attrib(default=False, init=True)
```
gabbard: nit: `init=True` is the default, so you don't need to specify it

adam/perception/_matcher.py
```diff
@@ -239,7 +239,9 @@ def isomorphisms_iter(self):
         for mapping in self.match():
             yield mapping
 
-    def match(self, *, debug: bool = False):
+    def match(
+        self, *, debug: bool = False, debug_function: Callable[[Any, Any], None] = None
```
gabbard: Can you rename this to `debug_callback`?

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
         dot_graph.render(output_file)
 
 
-@attrs(frozen=True, slots=True, eq=False)
+class DebugRenderingObject:
```
gabbard: Perhaps `DumpPartialMatchCallback` would be a more explicit name?

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
...
+    """
+        Helper callable class for debugging purposes. We render graph every hundred calls to match. The outputs can be
```
gabbard: style: Perhaps say something like "An instance of this object can be provided as the whatever-it-is argument of `GraphMatching.match` to render the match search process at every 100 time steps"

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
...
+        found in test/renders folder. This callable is called within GraphMatching.match
+    """
+
+    def __init__(self, graph_to_match_against: PerceptionGraph) -> None:
```
gabbard: Why do we need to provide `graph_to_match_against` to the constructor? Shouldn't we receive what graph information we need in `__call__`?

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
...
+        self.graph_to_match_against: PerceptionGraph = graph_to_match_against
+        self.calls_to_match_counter: int = 0
```
gabbard: nit: the type annotation here should be unnecessary.

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
...
+        self.start_time = process_time()
+        self.time_window = 0
```
gabbard: Something like `millis_to_wait_before_rendering` might be a clearer name.

adam/perception/perception_graph.py
gabbard: Docstring should note that we don't start rendering for a certain time period

adam/perception/perception_graph.py
```diff
@@ -501,7 +505,44 @@ def to_dot_node(pattern_node: "NodePredicate") -> str:
...
+            title = (
+                "graph_"
+                + str(id(self.graph_to_match_against))
+                + "_call_"
+                + str(self.calls_to_match_counter)
+            )
+            mapping = {k: "match" for k, v in graph_node_to_pattern_node.items()}
+            perception_graph.render_to_file(
+                graph_name=title,
+                output_file=Path("../renders/" + title),
```
gabbard: Never write to a path which depends on your programs current working directory. Make the user specify the directory to write to in the constructor.

adam/perception/perception_graph.py
```diff
@@ -519,11 +560,17 @@ class PerceptionGraphPatternMatching:
         validator=instance_of(PerceptionGraph)
     )
 
+    # Callable object for debugging purposes. We use this to track the number of calls to match and render the graphs.
+    rendering_debug_function: "DebugRenderingObject" = attrib(
```
gabbard: * Call this `debug_callback`
* Make the type `Optional[Callable[[DiGraph, Dict[Any, Any]]]]`. You probably want to make a type alias for the `Callable` portion to make it more readable.
* Make it default to `None` (we only want to do the debug rendering on request)

PR #435, opened 2019-11-20
Experiment Logging

gabbard:


adam/experiment/m6_experiments.py
```diff
@@ -0,0 +1,41 @@
...
+    output_dir = params.creatable_directory("output_directory")
+    experiment_name = params.optional_string("experiment", default="m6_experiments")
+
+    logger = HTMLLogger(
+        output_dir=output_dir,
+        experiment_name=experiment_name,
+        curriculum_name="m6_curriculum",
+    )
+
+    m6_curriculum = make_m6_curriculum()
```
gabbard: nit: you might as well in-line this below, since nothing else is done with it and the name doesn't clarify anything

adam/experiment/m6_experiments.py
```diff
@@ -0,0 +1,41 @@
...
+    execute_experiment(
+        Experiment(
+            name="static-prepositions",
+            training_stages=m6_curriculum,
+            learner_factory=SubsetLanguageLearner,
+            pre_example_training_observers=[
+                HTMLLoggerObserver(name="Pre-observer", html_logger=logger)
```
gabbard: Can you make methods on `HTMLogger` to  generate these, so you can just do `logger.pre_observer()`?

adam/experiment/observer.py
```diff
@@ -85,6 +88,160 @@ def report(self) -> None:
...
+
+table td { p
+    padding: 1em; 
+    background-color: #FAE5D3 ;
+}
+"""
+
+
+@attrs(slots=True)
+class HTMLLogger:
```
gabbard: This makes the reader ask: HTML _what_ logger?  Maybe `LearningProgressHtmlLogger`?

adam/experiment/observer.py
```diff
@@ -85,6 +88,160 @@ def report(self) -> None:
...
+
+    output_dir: Path = attrib(validator=instance_of(Path), kw_only=True)
+    experiment_name: str = attrib(validator=instance_of(str), kw_only=True)
+    curriculum_name: str = attrib(validator=instance_of(str), kw_only=True)
+    logging_dir: Path = attrib(init=False)
+    outfile: io = attrib(init=False)
```
gabbard: should this be private?  When is it closed? Should this be a context manager?

adam/experiment/observer.py
```diff
@@ -85,6 +88,160 @@ def report(self) -> None:
...
+    def __attrs_post_init__(self):
+        self.logging_dir = self.output_dir / self.experiment_name
+        self.logging_dir.mkdir(parents=True, exist_ok=True)
+        self.outfile = open(self.logging_dir / (self.curriculum_name + ".html"), "w")
+        self.html_dumper = CurriculumToHtmlDumper()
+        self.outfile.write(f"<head>\n\t<style>{CSS}\n\t</style>\n</head>")
+        self.outfile.write(
+            f"\n<body>\n\t<h1>{self.experiment_name} - {self.curriculum_name}</h1>"
+        )
+        self.outfile.write(
```
gabbard: comment on what this is for

adam/experiment/observer.py
```diff
@@ -85,6 +88,160 @@ def report(self) -> None:
...
+
+    output_dir: Path = attrib(validator=instance_of(Path), kw_only=True)
+    experiment_name: str = attrib(validator=instance_of(str), kw_only=True)
+    curriculum_name: str = attrib(validator=instance_of(str), kw_only=True)
+    logging_dir: Path = attrib(init=False)
+    outfile: io = attrib(init=False)
+    html_dumper: CurriculumToHtmlDumper = attrib(init=False)
+
+    def __attrs_post_init__(self):
+        self.logging_dir = self.output_dir / self.experiment_name
```
gabbard: Use `__attrs_post__init__` just for validation. If you need to do work at initialization-time, prefer object creation via a factory method.

adam/experiment/m6_experiments.py
```diff
@@ -0,0 +1,41 @@
...
+            ],
+            post_example_training_observers=[
+                HTMLLoggerObserver(name="Post-observer", html_logger=logger)
```
gabbard: I don't think we want to do two distinct loggings for pre-and-post, but rather to have a single entry, showing both pre-and-post on one line.  This would require storing some state in the pre-observation and doing the actual output on the post-observation.

adam/experiment/observer.py
```diff
@@ -85,6 +88,160 @@ def report(self) -> None:
...
+    r"""
+    Logs the true description and learner's descriptions throughout the learning process.
+    """
+    name: str = attrib(validator=instance_of(str))
+    html_logger: HTMLLogger = attrib(
+        init=True, validator=instance_of(HTMLLogger), kw_only=True
+    )
+    counter: int = attrib(kw_only=True, default=0)
+
+    def observe(  # pylint:disable=unused-argument
```
gabbard: It's a little cleaner if `observe` here does no work within this class but just forwards to a method on `HTMLLogger`.

adam/experiment/observer.py
denizbeser: What do you mean by a factory method? __init__()?

gabbard: Something like:

adam/experiment/m6_experiments.py
denizbeser: This requires restructuring DescriptionObserver - is that what you have in mind or something else?

denizbeser: I found a work-around, nevermind!

PR #415, opened 2019-11-04
414 positioning solver

chris-jenkins-isi:
In progress:
- Add orientation to BoundingBox 
- Test collision detection in non-axis-aligned situations (but no edge-to-edge collisions)
- Convert to matrix representations, configure for solving via linear programming

PR #446, opened 2019-11-27
Use SGD for to block collisions of axis-aligned bounding boxes

chris-jenkins-isi:
Replaces #415 PR by removing changes from #397 PR


adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
+from itertools import combinations
```
gabbard: module-level 📖

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+import torch.nn as nn
+from torch.nn import Parameter
+import torch.optim as optim
+from torch.optim.lr_scheduler import ReduceLROnPlateau
+
+from immutablecollections import immutabledict, immutableset, ImmutableDict
+from vistautils.preconditions import check_arg
+
+# see https://github.com/pytorch/pytorch/issues/24807 re: pylint issue
+ORIGIN = torch.tensor([0.0, 0.0, 0.0], dtype=torch.float)  # pylint: disable=not-callable
```
gabbard: Use `zeroes`?

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+COLLISION_PENALTY = 10
+
+
+@attrs(frozen=True, auto_attribs=True)
+class AdamObject:
+    name: str
+
+
+def main() -> None:
```
gabbard: 📖  to explain this is testing code and what it does.

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+
+        learning_rate_schedule.step(loss)
+
+        positioning_model.dump_object_positions()
+
+
+@attrs(frozen=True, slots=True)
+class AxisAlignedBoundingBox:
+    """
+    Defines a 3D Box that is oriented to world axes given a center point (3,)
```
gabbard: It is good to have a short single-line description as the first sentence of a docstring because some tools extract and use it.  For example,

```
Defines a 3D box that is oriented to world axes.

This box is defined by a center point of shape (3,)
and a scale (also of shape (3,)),
defines how a unit cube with the given center 
should be scaled in each dimension
to create this box.

For example, (your stuff after e.g.) 
```

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+                [1, -1, 1],
+                [-1, 1, 1],
+                [1, 1, 1],
+            ],
+            dtype=torch.float,
+        ).matmul(self.scale)
+        # see https://github.com/pytorch/pytorch/issues/24807 re: pylint issue
+
+    # helper functions giving names to a few corners used in calculations:
+    def right_corner(self) -> torch.Tensor:
```
gabbard: These three `X_corner` functions are used in only one place each; it might be better to just in-line them there unless you expect a general need for them.

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        Corner in the direction of the positive z axis from center (negative direction from other axes).
+        (Assuming box is oriented to world axes)
+        Returns: Tensor (3,)
+
+        """
+        return self.center + torch.tensor(  # pylint: disable=not-callable
+            [-1, -1, 1], dtype=torch.float
+        ).matmul(self.scale)
+
+    def minus_ones_corner(self) -> torch.Tensor:
```
gabbard: This one is used in several places, but maybe prefix it by `_` since I am not sure we want it as part of the "public API" of the bounding box class.

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        Corner in the direction of the negative x, y, z axes from center
+        Returns: Tensor (3,)
+
+        """
+        return self.center + torch.tensor(  # pylint: disable=not-callable
+            [-1, -1, -1], dtype=torch.float
+        ).matmul(self.scale)
+
+    # functions returning normal vectors from three perpendicular faces of the box
+    def right_face(self) -> torch.Tensor:
```
gabbard: The name `right_face_normal_vector` (and similarly below) will make their usages clearer to readers.

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def up_face(self) -> torch.Tensor:
+        """
+        Normal vector for the up face of the box (toward positive z axis when aligned to world axes)
+        Returns: Tensor (3,)
+        """
+        diff = self.up_corner() - self.minus_ones_corner()
+        return diff / torch.norm(diff)
+
+    def get_face_norms(self) -> torch.Tensor:
```
gabbard: `face_normal_vectors` could be better, since "norm" has its own, different meaning.  I think you can drop the `get_`.

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        """
+        Stacks the face norms from the right, forward, and up faces of the box
+        Returns: Tensor (3,3)
+
+        """
+        # in axis-aligned case these are always the same
+        return torch.stack([self.right_face(), self.forward_face(), self.up_face()])
+
+    def corner_onto_axes_projections(self, axes: torch.Tensor) -> torch.Tensor:
```
gabbard: Shouldn't this be `corners`, plural?

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        Returns:
+            (3, 8) tensor -> each point projected onto each of three dimensions
+
+        """
+        check_arg(axes.shape == (3, 3))
+        corners = self.get_corners()
+        return axes.matmul(corners.transpose(0, 1))
+
+
+class AdamObjectPositioningModel(torch.nn.Module):
```
gabbard: 📖

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+            f"distance penalty: {distance_penalty}\ncollision penalty: {collision_penalty}"
+        )
+        return distance_penalty + collision_penalty
+
+    def dump_object_positions(self) -> None:
+        for (adam_object, bounding_box) in self.adam_object_to_bounding_box.items():
+            print(f"{adam_object.name} = {bounding_box.center.data}")
+
+
+class DistanceFromOriginPenalty(nn.Module):
```
gabbard: 📖

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+    def __init__(self) -> None:  # pylint: disable=useless-super-delegation
+        super().__init__()
+
+    def forward(  # pylint: disable=arguments-differ
+        self, bounding_box: AxisAlignedBoundingBox
+    ) -> torch.Tensor:
+        return bounding_box.center_distance_from_point(ORIGIN)
+
+
+class CollisionPenalty(nn.Module):
```
gabbard: 📖

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        # stack the minim
+        return torch.stack((minima, maxima), 1)
+
+    @staticmethod
+    def get_min_max_overlaps(
+        min_max_proj_0: torch.Tensor, min_max_proj_1: torch.Tensor
+    ) -> torch.Tensor:
+        """
+        Given min/max corner projections onto 3 axes from two different objects,
+        return an interval for each dimension representing the degree of overlap or
```
gabbard: Comment on how the returned values should be interpreted in terms of overlap/separation

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+        # with columns like:
+        # [ min0x   min1x
+        #   min0y   min1y
+        #   min0z   min1z
+        #                ]
+        # then find the maximum element from each row
+
+        # repeat the process for the min of the max projections
+        maxs_0 = min_max_proj_0.gather(
+            1, torch.tensor([[1], [1], [1]])  # pylint: disable=not-callable
```
gabbard: here and below, `ones`?

adam/visualization/positioning.py
```diff
@@ -0,0 +1,426 @@
...
+            min_max_overlaps: (3, 2) tensor -> intervals describing degree of overlap between the two boxes
+
+        Returns: Tensor with a positive scalar of the collision penalty, or tensor with zero scalar
+        for no collision.
+        """
+        check_arg(min_max_overlaps.shape == (3, 2))
+        # subtract each minimum max from each maximum min:
+        overlap_distance = min_max_overlaps[:, 0] - min_max_overlaps[:, 1]
+
+        # if ANY element in overlap_distance is positive, the minimum positive value
```
gabbard: This comment seems garbled to me - am I missing something?

adam/visualization/positioning.py
chris-jenkins-isi: I would need the shorthand to turn a flat vector of [1, 1, 1,] into 
[
    [1],
    [1],
    [1]
]

chris-jenkins-isi: which would be torch.zeros(3, 1) 😅

tests/visualization/positioning_test.py
```diff
@@ -0,0 +1,44 @@
...
+    AxisAlignedBoundingBox,
+    main,
+    CollisionPenalty,
+    DistanceFromOriginPenalty,
+)
+
+
+def test_positioning() -> None:
+    np.random.seed(127)
+    main()
```
gabbard: I would split this into three separate tests so that each test tests one thing.

Seeing `main` in a test looks weird to the reader, so perhaps rename it to something else which makes clear what it is doing.  Also comment that we are just checking the optimization executes without crashing, but not checking its result so that the reader is not confused by the lack of an assertion.

tests/visualization/positioning_test.py
```diff
@@ -0,0 +1,44 @@
...
+    # positioned to be colliding with one another:
+    aabb0 = AxisAlignedBoundingBox.create_at_center_point(
+        center=np.ndarray(shape=(3,), buffer=np.array([0, 0, 0]))
+    )
+    aabb1 = AxisAlignedBoundingBox.create_at_center_point(
+        center=np.ndarray(shape=(3,), buffer=np.array([0.5, 0.5, 0.5]))
+    )
+
+    # set to definitely not be near the origin:
+    aabb_rand = AxisAlignedBoundingBox.create_at_random_position(
```
gabbard: For a test, rather than creating at random positions which are known not to collide, it would be better to create the boxes explicitly at colliding and non-colliding positions.

tests/visualization/positioning_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+    collision_penalty = CollisionPenalty()
+
+    res = collision_penalty.forward(aabb0, aabb1)
+    assert res > 0
+
+    res = collision_penalty.forward(aabb0, aabb_rand)
+    assert res <= 0
+
+    distance_from_origin = DistanceFromOriginPenalty()
```
gabbard: Let's document `DistanceFromOriginPenalty` as penalizing by the Euclidean distance and then check the value exactly (or exactly with some tolerance, which is what you want to do for floating point values in tests - for pytest this is e.g. `assert distance == pytest.approx(4.5)`)

Issue #456, opened 2019-12-05
Make `far` and `near` implement symmetric relationships for situation generation


LichMaster98:
This issue already has a PR open for it which begins the work of this change see: 

When we add regional relationships to out scenes we expect to add symmetric relationships to the situation because the perception just copies over the situation relationships remapping the objects. However, when we implement the change which creates the symmetric relationships the English language generator is no longer able to determine how to create a sentence for the scene. For example: 
```RuntimeError: Error while generating English for situation {
E               salient objects:
E                   ball
E                   cookie
E               other objects:
E                   book
E                   table
E                   mom[is-speaker[binary,perceivable]]
E                   learner[is-addressee[binary,perceivable]]
E               always relations:
E                   in-region(ball, Region(cookie,distance=proximal,direction=+FacingAddresseeAxis(_object=cookie)))
E                   in-region(cookie, Region(ball,distance=proximal,direction=-FacingAddresseeAxis(_object=cookie)))
E               from template: preposition-training-ball-behind-cookie
E           }```

One would be able to assert "a ball behind a cookie" or "a cookie in front of a ball" from this situation. To resolve this issue, we need to give the English language generator (or any language generator) a hint as to what sentence we want to return for the situation.

**Comment by paynesa** [comment]
A potential solution to this would be to add some kind of attribute for relations not to translate into language, either to `Situation` or as a part of the input to the language generator. For the example above, we could do something like: 

```
RuntimeError: Error while generating English for situation {
E               salient objects:
E                   ball
E                   cookie
E               other objects:
E                   book
E                   table
E                   mom[is-speaker[binary,perceivable]]
E                   learner[is-addressee[binary,perceivable]]
E               always relations:
E                   in-region(ball, Region(cookie,distance=proximal,direction=+FacingAddresseeAxis(_object=cookie)))
E                   in-region(cookie, Region(ball,distance=proximal,direction=-FacingAddresseeAxis(_object=cookie)))
E               relations_to_not_translate = [in-region(ball, Region(cookie,distance=proximal,direction=+FacingAddresseeAxis(_object=cookie)))]
E               from template: preposition-training-ball-behind-cookie
E           }
```
so that we get "the cookie is near the ball" rather than trying to translate both. Something similar could work for "over" and "under" where the mirrored relation is added to the relations that shouldn't be translated to language. This is a bit hackish though, and it might be better to fix this in the language generator itself.

[...]

PR #457, opened 2019-12-05
Fix english generation to handle near and far as Symmetric relationship for Situation

LichMaster98:
WIP for #456


LichMaster98:
Closing as any work on solving this symmetry should, at this point, just be started from the main branch as this is so far behind.

PR #458, opened 2019-12-06
Introduce Gaze to Preposition Curriculum 

### LINKED ISSUE #430: add 'gaze' to the first element of a prepositional phrase in the linguistic description in preposition curriculum

LichMaster98:
Closes #430

### DIFF
--- a/adam/situation/templates/phase1_templates.py
+++ b/adam/situation/templates/phase1_templates.py
@@ -200,6 +200,13 @@ class Phase1SituationTemplate(SituationTemplate):
     All `TemplateObjectVariable`\ s in the situation, 
     both salient and auxiliary to actions.
     """
+    gazed_objects: ImmutableSet[
+        Union[TemplateObjectVariable, Region[TemplateObjectVariable]]
+    ] = attrib(converter=_to_immutableset, kw_only=True)
+    """
+    A set of `TemplateObjectVariables` s which are the focus of the speaker. 
+    Defaults to all semantic role fillers of situation actions.
+    """
 
     def __attrs_post_init__(self) -> None:
         check_arg(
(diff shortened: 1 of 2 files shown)


adam/situation/templates/phase1_templates.py
```diff
@@ -444,6 +459,12 @@ def _instantiate_situation(
             ],
             syntax_hints=template.syntax_hints,
             axis_info=self._compute_axis_info(object_var_to_instantiations),
+            gazed_objects=immutableset(
```
gabbard: Why do we allow `Region`s to be gazed at?

LichMaster98: @gabbard By default, we grab the `Defaults to all semantic role fillers of template actions.` and regions can fill a role such as a `GOAL` in an action.

gabbard: @LichMaster98 : I am confused by:
> By default, we grab the Defaults to all semantic role fillers of template actions

Can you clarify?

LichMaster98: @gabbard Yes. The image below is an instance of the "put" curriculum in our phase 1 curriculum. Where the semantic role fillers for the action are `dog`, `hat`, and `Region(reference_object=car, distance=exterior-but-in-contact, direction=+_GravitationalAxis())`

The default which is used in the scene generation code grabs all three of these and adds them to a list of gazed items. I copied this implementation into the template builder to allow for specifying explicit items which are being gazed at in the template scene. Because regions can appear as semantic role fillers (as in this example) I handled them in a way that would mirror how the situation code appears to be handling them.

image

gabbard: I think this is a bug in the situation code.  Can you make an issue to block `Region`s from being gazeable?

LichMaster98: And I'll fix the template code to block regions as well.

PR #490, opened 2019-12-12
Return the last pattern node to fail to match on failed graph pattern matches

gabbard:


adam/learner/preposition_subset.py
```diff
@@ -396,13 +393,5 @@ def replace_template_variables_with_object_names(
                         1.0,
                     )
                 )
-                if debug_mapping:
```
gabbard: re: this deletion - note this would never be reached if the match were in fact partial, because if it were `.matches()` would return nothing.

gabbard:
Superseded by #493

PR #493, opened 2019-12-12
Pattern relaxation

### LINKED ISSUE #461: Implement pattern relaxation in the case of partial matches as a general capability
From a conversation on Slack with @denizbeser : 

> So, the direct way to find the maximum common isomorphic subgraph of two graphs is to construct this structure: https://en.wikipedia.org/wiki/Modular_product_of_graphs and then find the maximal clique in it
> However, for two graphs of size (in terms of nodes) X and Y you get a module product graph with NX nodes which is in general much dense than the original graph
> And maximum clique algorithms are themselves largely NP-complete, and the densification resulting from constructing the modular product graph will probably lose use the special structural properties which are allowing us to solve the (also NP-complete) graph isomorphism problem relatively quickly
> There is a backtracking method for MCS similar to VF++: https://www.semanticscholar.org/paper/Backtrack-Search-Algorithms-and-the-Maximal-Common-McGregor/ed10c6f788922cd5e7cb26c3d55f676979958852
> Another challenge, not yet addressed, is that some of the backoff we want to do is not just in terms of node pruning but also changing the node and edge properties of pattern elements. For example, we may want to decide that particular aspects of a geon shape don’t actually matter
> Or our initial hypothesis may be that a certain color is required for an object, and we then relax that to a certain range of colors
(imagine juice/milk/water)
> So I think we have two problems: (1) how do we compute maximum common isomorphic sub-graph? (2) how do we account for “relaxing pattern” nodes
> Let’s put (1) aside for the moment, because the answer to (2) informs our approach to (1)
> One possibility is that we find a way to re-express our patterns so that everything is atomic
> Let’s see if we can do that.
> What non-atomic pattern nodes do we have?
> Let’s see, there’s AxisPredicate:
> curved: Optional[bool] = attrib(validator=optional(instance_of(bool)))
> directed: Optional[bool] = attrib(validator=optional(instance_of(bool)))
> aligned_to_gravitational: Optional[bool] =attrib(validator=optional(instance_of(bool)))
> But we could recast this as a property-less IsAxis node which is linked to those three bools by hasProperty edges
> Then we could relax that, for example, aligned_to_gravitational doesn’t matter by simply deleting that property edge
> GeoPredicate is complex, so let’s return to it in a moment
> RegionPredicate could treat Distance just like AxisPredicate
> This approach doesn’t fix the color node, or IsOntologyNode if we want to expand it to know about the ontology hierarchy
> (e.g. so you could have a pattern node for “animate thing”)
> Not clear re:AndNode, but that is currently unused
> Let’s see about edges:
> DirectionPredicate is tricky because it involved a reference to an axis
> If we took that approach it would need to be “reified” into a node somehow
> Ah, but because the direction is never used independently of a Region, that could be folded onto another edge coming out of Region
> So DirectionPredicate poses no fundamental problem
> Ok, I think Geon can be dealt with similar to AxisPredicate
> So the obstacles to handling relaxing with a pure graph pruning algorithm are > just IsOntologyNode and the color nodes
> So that lets us know the scope of what we might have to special case, if we can get (1) to work
> hmm - a possible heuristic approach could be this:
> 1. attempt a regular match
> 2. if it fails, identify the pattern node we were trying to match and could not either excise that node (and its children, if any) or relax it
> 3. repeat the process until either you get a match of some sort or end up deleting your whole pattern graph
> Now this is not perfect because it is not guaranteed that the final node you failed to match is at fault - it could have been that you made a bad decision earlier which locked you into a place this node could not be matched
> But I think it might deal with almost all our cases

### LINKED ISSUE #479: Return pattern node which last failed to match on failed pattern match

### LINKED ISSUE #480: Allow graph matching process to begin from a partial match

### LINKED ISSUE #481: Implement pruning-based backoff for partial matching of patterns

gabbard:
Closes #461
Closes #479 
Closes #480 
Closes #481

### DIFF
--- a/adam/learner/preposition_subset.py
+++ b/adam/learner/preposition_subset.py
@@ -375,10 +375,7 @@ class PrepositionSubsetLanguageLearner(
             matcher = preposition_pattern.graph_pattern.matcher(
                 recognized_object_perception.perception_graph
             )
-            debug_mapping: Dict[Any, Any] = dict()
-            for match in matcher.matches(
-                debug_mapping_sink=debug_mapping if self._debug_file else None
-            ):
+            for match in matcher.matches():
                 # if it is, use that preposition to describe the situation.
                 description_to_score.append(
                     (
@@ -396,13 +393,5 @@ class PrepositionSubsetLanguageLearner(
                         1.0,
                     )
                 )
-                if debug_mapping:
-                    self._print(
-                        subgraph(
-                            preposition_pattern.graph_pattern.copy_as_digraph(),
-                            nodes=debug_mapping,
-                        ),
-                        "Largest Partial Match",
-                    )
 
         return immutabledict(description_to_score)
--- a/adam/learner/subset.py
+++ b/adam/learner/subset.py
@@ -1,27 +1,27 @@
-from typing import Dict, Generic, Mapping, Tuple, Optional, Any
+from typing import Dict, Generic, Mapping, Optional, Tuple
 
-from attr import Factory, attrib, attrs
 from immutablecollections import immutabledict
 from more_itertools import first
 from networkx import DiGraph, isolates
 
 from adam.language import (
+    LinguisticDescription,
     LinguisticDescriptionT,
     TokenSequenceLinguisticDescription,
-    LinguisticDescription,
 )
 from adam.learner import LanguageLearner, LearningExample
 from adam.ontology.phase1_ontology import LEARNER
-from adam.perception import PerceptionT, PerceptualRepresentation, ObjectPerception
+from adam.perception import ObjectPerception, PerceptionT, PerceptualRepresentation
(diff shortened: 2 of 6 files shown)


gabbard:
What is left to do here:
* @denizbeser will add unit tests for matching starting from a partial match, including:
  * one where we are given a partial pattern match we can successfully extend
  * one where our partial pattern match is broken and fails semantic feasibility for some node (and should raise an exception)
  * one like the above, but which fails syntactic feasibility

Someone needs to add simple tests for node-relaxation-by-pruning, including 
* the simple cases of pruning off a leaf node
* pruning off an object and its tree of sub-objects
* a pattern getting pruned away to nothing

It is likely there are bugs in the node relaxation since I haven't even run it, but I'm out of ADAM time for this week. :-) . @denizbeser: I can answer questions as tests reveal bugs, though.

PR #505, opened 2019-12-20
restructured parameter files and log structure for m6

denizbeser:


parameters/experiments/m6/pursuit-with-noise.params
```diff
@@ -2,7 +2,7 @@ _includes:
    - "../../root.params"
    - "pursuit-nouns.m6.params"
 
-output_directory: '%adam_experiment_root%/m6/deniz'
+output_directory: '%adam_experiment_root%/'
```
gabbard: We should never output directly to the experiment root because the users might have other experiments besides this. Put this in an `m6` sub-directory which will be the root of our archive.

adam/learner/pursuit.py
```diff
@@ -279,7 +279,7 @@ def lexicon_step(self, word: str) -> None:
             # Remove the word from hypotheses
             self._words_to_hypotheses_and_scores.pop(word)
             logging.info("Lexicalized %s as %s", word, leading_hypothesis_pattern)
-            print(f"LExicalized {word} as {leading_hypothesis_pattern}")
+            # print(f"Lexicalized {word} as {leading_hypothesis_pattern}")
```
gabbard: You can just delete it rather than comment it

adam/experiment/log_experiment.py
```diff
@@ -28,7 +28,7 @@ def main(params: Parameters) -> None:
     logger = LearningProgressHtmlLogger.create_logger(
         output_dir=output_dir,
         experiment_name=experiment_name,
-        curriculum_name="m6_curriculum",
+        curriculum_name="index",
```
gabbard: This seems somewhat of an abuse of the parameter - would it be better to folder the curriculum name into the directory name (or just not take the curriculum name at all, and let the user put it in the experiment name?), and then have the code write to an `index.html`?

PR #532, opened 2020-01-16
Compute differences between perception frames

### LINKED ISSUE #65: Compute the difference between two perception frames
For learning non-static verbs, it is crucial to be able to represent changes in situations.  We need to make an object which can take two perception frames and produce another object which represents the "diff" between these frames. 

You can assume that Python reference identity for objects in situation data structures corresponds to "real" object identity for objects in the virtual situations.

denizbeser:
closes #65


tests/perception/render_perception_graph.py
```diff
@@ -0,0 +1,38 @@
...
+from adam.perception.perception_graph import PerceptionGraph
+from adam.random_utils import RandomChooser
+from adam.situation.templates.phase1_templates import (
+    Phase1SituationTemplate,
+    all_possible,
+    object_variable,
+)
+
+
+def render_perception_graph():
```
gabbard: Do we need to check this in?

denizbeser: @gabbard What do you mean by that - to submit to Darpa? I can move it to a higher level directory if you prefer.

gabbard: @denizbeser : I meant rather "does this need to be kept in the repository"?  Every entry point we have adds to the cognitive burden of someone getting acquainted with the project, so if the program doesn't serve any ongoing purpose, we can remove it from the repo.

denizbeser: @gabbard I see. I will keep it locally for debugging purposes and remove it from the repo then.

Issue #443, opened 2019-11-26
Should we express "contact"  by a relation rather than a region distance?


gabbard:
In defining regions, Landau and Jackendoff distinguish four differences from a reference object: INTERIOR, EXTERIOR-BUT-IN-CONTACT, PROXIMAL, and DISTAL.  

The one tricky case is the distance category EXTERIOR-BUT-IN-CONTACT, which adds the additional requirement of "contactingness" to a purely spatial constraint.  I am wondering if it would be cleaner to diverge from L&J here and just have three distance values, INTERIOR, PROXIMAL, and DISTAL, and to encode contact between objects as a relation.

LichMaster98:
After our discussion on the 27th, we decided to move forward with this idea. 
https://github.com/isi-vista/adam/issues/450

gabbard:
Closed as duplicate of #450

PR #533, opened 2020-01-16
changed represenation of contacts relation

### LINKED ISSUE #450: Move our representation of `contacts` into a relationship rather than a distance in our regions
Originally we intended to do the idea of `contacts` in this relational method but we changed to use the Biederman representation of `exterior-but-in-contacts` being a distance in regional representation. We've found that this model can create a confusing interpretation of how regions of objects overlap. To solve this issue we're going to move the `exterior-but-in-contacts` distance into a relationship of `contacts`.

The DSL for `contacts` should create the relationship between both the objects, as well as assign a `proximal` region relationship. As objects which are in contact are, by definition, `proximal` to each other.

denizbeser:
Closes #450


gabbard:
This change has been backlogged (#443 ); closing this PR for now.

PR #538, opened 2020-01-24
"Added FilePath conversion to work on any OS"

### LINKED ISSUE #537: Support OS Specific File Paths in Visualization (Panda3d)
Panda3d was not able to locate my `.egg` model files on my windows system until I noticed this in `panda3d_interface.py` (Ln 194):
`working_dir = os.path.abspath((sys.path[0]))`
When run on a windows system, this creates a path using `\` characters, which is not recognized by panda3d, raising an IOError.

aakum03:
Closes #537 
Adds:
 `from panda3d.core import Filename` 
and use that to:
 `working_dir = Filename.fromOsSpecific(os.path.abspath((sys.path[0])))`
This creates a Path `(working_dir)` that Panda3d can read regardless of OS.


aakum03:
@chris-jenkins-isi can it be merged?

PR #561, opened 2020-02-06
Initial start at go with prepositions

### LINKED ISSUE #549: GO curriculum with prepositions
region = destination:
_to, in, beside_

region = destination or path:
_behind, in front of, over, under_

LichMaster98:
Closes #549

### DIFF
--- a/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
+++ b/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
@@ -207,7 +207,7 @@ def _go_over_template(
                         Region(
                             goal_object,
                             distance=DISTAL if is_distal else PROXIMAL,
-                            direction=GRAVITATIONAL_DOWN,
+                            direction=GRAVITATIONAL_UP,
                         ),
                     ),
                 ],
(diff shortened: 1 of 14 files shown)


LichMaster98:
This PR is rebased on #562 as it needs the requirements of having a default addressee in the scene.

adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
```diff
@@ -0,0 +1,443 @@
...
+                GO,
+                argument_roles_to_fillers=[
+                    (AGENT, agent),
+                    (
+                        GOAL,
+                        Region(
+                            goal_object,
+                            distance=DISTAL if is_distal else PROXIMAL,
+                            direction=Direction(
+                                positive=False,
```
gabbard: Can you try to think through what changes we need so that we can account for the positive direction of an axis not being correlated with whether it faces a given person?

We might need to modify `AxisFunctions` to return axis-sign tuples.

LichMaster98:
Due to pulling out the changes for the English language generator and the learner as default I've elected to start a new PR from master rather than fiddle with a rebase as the former will be faster. See https://github.com/isi-vista/adam/pull/576 for the new "go" PR

PR #562, opened 2020-02-06
Ensure Learner is addressee by default

LichMaster98:


adam/situation/templates/phase1_templates.py
```diff
@@ -403,30 +422,73 @@ def _instantiate_objects(
...
+                self._instantiate_object(
+                    obj_var,
+                    variable_assignment,
+                    has_addressee=has_addressee,
+                    default_addressee_node=default_addressee_node,
+                ),
+            )
+            for obj_var in template.all_object_variables
+        ]
+        if (
```
gabbard: Will these assign the addressee to the learner if the learner is present but not yet marked as the addressee?

LichMaster98: Yes. There is a unit test for this case to verify as well.

PR #564, opened 2020-02-08
545 propose but verify

### LINKED ISSUE #545: Test Propose-but-verify model of word learning
In addition to our main Pursuit learning algorithm, we want to test other word-learning models, such as _Propose-but-verify_ and _cross-situational learning_. This will require a separate implementation of the LanguageLearner class (e.g a ProposeButVerifyLanguageLearner) and tests that are similar to the ones we have for Pursuit Learner to evaluate the behavior.

**Details on the algorithm:** The Pursuit paper [here](https://www.ling.upenn.edu/~ycharles/papers/pursuit-final.pdf) contains the necessary background information on how these word-learners behave. Pages 3-4 both propose-but-verify and cross-situational learning algorithms more specifically.

aakum03:
Closes #545 
Created the `propose_but_verify` language learner based on the "propose but verify" word learning model.

### DIFF
--- a/adam/learner/propose_but_verify.py
+++ b/adam/learner/propose_but_verify.py
@@ -41,6 +41,15 @@ class ProposeButVerifyLanguageLearner(
     Generic[PerceptionT, LinguisticDescriptionT],
     LanguageLearner[PerceptionT, LinguisticDescription],
 ):
+    """
+    An implementation of `LanguageLearner` for the 'propose but verify' learning based approach for
+    single object detection. This model does not retain any information pertaining to
+    disconfirmed hypothesis and is based on a "win-stay, lose-shift" ideology.
+
+    Reference: https://www.ling.upenn.edu/~ycharles/papers/pursuit-final.pdf
+    Section 1 of this paper on pursuit learning includes an introduction to the 'propose but
+    verify' learning approach.
+    """
 
     _ontology: Ontology = attrib(validator=instance_of(Ontology), kw_only=True)
     _observation_num = attrib(init=False, default=0)
@@ -61,6 +70,7 @@ class ProposeButVerifyLanguageLearner(
     _rendered_word_hypothesis_pair_ids: Set[str] = attrib(
         init=False, default=Factory(set)
     )
+    _minimum_match_ratio: float = attrib(default=0.8, kw_only=True)
 
     @staticmethod
     def from_parameters(
@@ -89,13 +99,12 @@ class ProposeButVerifyLanguageLearner(
     def observe(
         self, learning_example: LearningExample[PerceptionT, LinguisticDescription]
     ) -> None:
-
         logging.info("Observation %s", self._observation_num)
         self._observation_num += 1
 
         perception = learning_example.perception
         if len(perception.frames) != 1:
-            raise RuntimeError("Pursuit learner can only handle single frames for now")
+            raise RuntimeError("ProposeButVerify learner can only handle single frames for now")
         if isinstance(perception.frames[0], DevelopmentalPrimitivePerceptionFrame):
             original_perception_graph = PerceptionGraph.from_frame(
                 perception.frames[0]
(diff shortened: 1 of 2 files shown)


adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+        observed_linguistic_description = (
+            learning_example.linguistic_description.as_token_sequence()
+        )
+
+        self.learn_propose_but_verify(
+            observed_perception_graph, observed_linguistic_description
+        )
+
+    # learn method
+    def learn_propose_but_verify(
```
denizbeser: Can you add tests that checks if this is working as intended? Something similar to PursuitLearner tests with single-object situations would be enough

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+                pattern_hypothesis,
+                logging.INFO,
+                "Initializing (randomly chosen) meaning for %s ",
+                word,
+            )
+
+        self._words_to_hypotheses[word] = pattern_hypothesis
+
+    # Test Hypothesis
+    def test_hypothesis_step(
```
denizbeser: Rename to "verify_hypothesis()" perhaps? It seems to function as a verification returning a boolean instead of an entire learning step

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+                )
+                self._rendered_word_hypothesis_pair_ids.add(hypothesis_id)
+        else:
+            logging.info(
+                f"The word {word} has never been seen before. There are no current hypotheses for it"
+            )
+
+    # Helper method, from pursuit.py
+    @staticmethod
+    def get_objects_from_perception(
```
denizbeser: It would be nice to have this modularized so we can refer to the same function across learners. Let's just keep it this way for now though, as the pursuit_learner.py is also changing right now.

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+    GraphLogger,
+    PerceptionGraph,
+    PerceptionGraphPattern,
+)
+from adam.utils import networkx_utils
+from attr import Factory, attrib, attrs
+
+
+@attrs()
+class ProposeButVerifyLanguageLearner(
```
denizbeser: Can you add a docstring briefly explaining what this learning algorithm does - with a link to the paper that explains the algorithm?

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+    def observe(
+        self, learning_example: LearningExample[PerceptionT, LinguisticDescription]
+    ) -> None:
+
+        logging.info("Observation %s", self._observation_num)
+        self._observation_num += 1
+
+        perception = learning_example.perception
+        if len(perception.frames) != 1:
+            raise RuntimeError("Pursuit learner can only handle single frames for now")
```
denizbeser: The error message should be "Propose but verify... " instead

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+                    # If hypothesis is confirmed, do nothing
+                    continue
+                else:
+                    # If hypothesis is not confirmed, treat word as an unseen word, and go
+                    # through the initialization step again (to randomly select meaning)
+                    self.initialization_step(word, observed_perception_graph)
+            if self._log_word_hypotheses_to:
+                self._log_hypotheses(word)
+
+    def initialization_step(self, word: str, observed_perception_graph: PerceptionGraph):
```
denizbeser: Might be better to name it `initialize_new_meaning()` or so, since we no longer have "learning steps" like we used to in pursuit.

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,372 @@
...
+        hypothesis_is_confirmed = partial_match.matched_exactly()
+        # If the hypothesis is confirmed, we leave it as it is
+        if hypothesis_is_confirmed and partial_match.matching_subgraph:
+            logging.info("Hypothesis is confirmed")
+        # If the hypothesis is disconfirmed
+        else:
+            logging.info(
+                f"Hypothesis is disconfirmed (Match ratio: {partial_match.match_ratio()}"
+            )
+        return hypothesis_is_confirmed
```
denizbeser: Can you add a parameter (e.g minimum_match_ratio) into the learner such that a hypothesis is confirmed if the match ratio >= the minimum match ratio? The learner as is might never successfully verify because of minor variations in the perceptual representations.

adam/learner/propose_but_verify.py
aakum03: Tests have been added to `learner_tests`. It still uses the pursuit curriculum since it works as intended.

adam/learner/propose_but_verify.py
aakum03: Renamed. I wasn't sure if a naming convention (i.e: with the word 'step')

adam/learner/propose_but_verify.py
aakum03: Added

adam/learner/propose_but_verify.py
aakum03: Changed

adam/learner/propose_but_verify.py
aakum03: Renamed. I wasn't sure if a naming convention (i.e: with the word 'step')

adam/learner/propose_but_verify.py
aakum03: Added. The default minimum_match_ratio has been set to 0.8; with the test parameters using 0.85

aakum03:
@denizbeser 
When running tests, I ran into some issues with `describe()` as it did not return any matches `(observe worked as intended, and the hypotheses were formed and stored as intended)`. So I modified `describe() `to use `PartialMatch` that was originally only used to verify hypotheses instead of `matcher()` that the original `describe()` in pursuit used.

This resulted in the intended description and the tests passed (single object only).

I am unsure of any implications of this change. Please advise.

gabbard:
@aakum03 : Sorry for the long delay. Can you rebase this to the current master prior to review?

LichMaster98:
Replaced by: https://github.com/isi-vista/adam/pull/966

PR #565, opened 2020-02-10
Implementing `Cross_SituationalLanguageLearner`

jamart28:
Implemented `observe()` method so far
Copied from `PursuitLanguageLearner` as a baseline
Used the following paper for model used: [Cross-Situational Learner](https://onlinelibrary.wiley.com/doi/full/10.1111/j.1551-6709.2010.01104.x)
`describe()` still needs to be implemented

Addresses #546


adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+    GraphLogger,
+    PerceptionGraph,
+    PerceptionGraphPattern,
+)
+from adam.utils import networkx_utils
+from attr import Factory, attrib, attrs
+
+
+@attrs
+class CrossSituationalLanguageLearner(
```
denizbeser: Can you write tests for this, similar to those for Pursuit?

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+    Generic[PerceptionT, LinguisticDescriptionT],
+    LanguageLearner[PerceptionT, LinguisticDescription],
+):
+    """
+    An implementation of `LanguageLearner` for cross-situational learning based approach for single
+    object detection.
```
denizbeser: Can you add a brief description of the algorithm here, with a link to the paper?

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+
+    _words_to_hypotheses_and_scores: Dict[
+        str, Dict[PerceptionGraphPattern, float]
+    ] = attrib(init=False, default=Factory(dict))
+    _words_to_hypotheses_and_probability: Dict[
```
denizbeser: It would be useful to have a comment here describing the difference between score and probability

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+                self._words_to_hypotheses_and_scores.pop(word)
+                if self._graph_logger:
+                    self._graph_logger.log_graph(
+                        leading_hypothesis_pattern, logging.INFO, "Lexicalized %s", word
+                    )
+            else:
+                logging.info("Would lexicalize, but haven't see the word often enough")
+
+    @staticmethod
+    def get_objects_from_perception(
```
denizbeser: There are a couple of methods that are shared between learners. It's generally better to modularize methods like this that is shared across the code. Let's keep it this way for now though, since I'm working on a different branch changing the Pursuit class right now.

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+            else:
+                self._words_to_number_of_observations[word] = 1
+
+            # If word is not known, go through learning steps
+            if word not in self._lexicon:
+                logging.info(f"Considering '{word}'")
+                # Initialize necessary components if needed
+                if word not in self._words_to_hypotheses_and_scores:
+                    self._words_to_hypotheses_and_scores[word] = {}
+                    self._words_to_hypotheses_and_probability[word] = {}
```
denizbeser: You can use `defaultdict` class to avoid having to initailize empty dicts: https://docs.python.org/3/library/collections.html#collections.defaultdict

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+                    for hypothesis in self._words_to_hypotheses_and_scores[word]:
+                        if hypothesis.check_isomorphism(meaning):
+                            # Updating association score as described by the paper below
+                            # assoc(w, m) = pass_assoc(w, m) + a(w|m, U)
+                            # https://onlinelibrary.wiley.com/doi/full/10.1111/j.1551-6709.2010.01104.x (2)
+                            self._words_to_hypotheses_and_scores[word][
+                                hypothesis
+                            ] += self.alignment_probability(word, words, hypothesis)
+                            break
+                    else:
```
denizbeser: There seems to be an indentation issue with this clause

adam/learner/cross_situational.py
```diff
@@ -0,0 +1,567 @@
...
+        ) / (sum_of_all_scores + number_of_meanings * self._smoothing_parameter)
+        times_word_has_been_seen = self._words_to_number_of_observations[word]
+        logging.info(
+            "Prob of meaning given word: %s, Times seen: %s",
+            probability_of_meaning_given_word,
+            times_word_has_been_seen,
+        )
+        # file (w, h^) into the lexicon
+
+        # TODO: We sometimes prematurely lexicalize words, so we use this arbitrary counter
```
denizbeser: How did you test this?

adam/learner/cross_situational.py
jamart28: The `else:` is for `for hypothesis in self._words_to_hypotheses_and_scores[word]:` so the indentation is intended unless there's something I'm not seeing. @denizbeser

denizbeser: I see, didn't notice it's for/else.

jamart28:
The tests do pass but take longer than 10 min. The shortest I've gotten it to is 16 min testing with 30 instances.

gabbard:
@jamart28 : Sorry for the long delay. Can you rebase this to the current `master` prior to review?

LichMaster98:
Closing as #998 has implemented this as a sublearner

PR #567, opened 2020-02-11
Add regions as goals for english_language_generator.py

### LINKED ISSUE #566: Increase coverage of Regions as Goals for English Language Generator
Implement "beside", "behind", "in front of", "over", and "under"

LichMaster98:
Closes #566

### DIFF
--- a/adam/language_specific/english/english_language_generator.py
+++ b/adam/language_specific/english/english_language_generator.py
@@ -615,7 +615,7 @@ class SimpleRuleBasedEnglishLanguageGenerator(
                 return "over"
             elif region.direction == GRAVITATIONAL_DOWN:
                 return "under"
-            # region.distance == DISTAL is not check as this does not define a specific preposition in scope for M9
+            # region.distance == DISTAL is not check as this does not define a specific preposition in scope for Phase 1
             elif region.direction and self.situation.axis_info:
                 if not self.situation.axis_info.addressee:
                     raise RuntimeError(
@@ -625,6 +625,12 @@ class SimpleRuleBasedEnglishLanguageGenerator(
                 # HACK, from M3
                 # see: https://github.com/isi-vista/adam/issues/573
                 if isinstance(region.direction.relative_to_axis, FacingAddresseeAxis):
+                    # "in front of" and "behind" is defined without a distance as you can accurate use the phrase
+                    # regardless of distance example:
+                    # "the teacher is in front of your laptop"
+                    # (Assuming the laptop is near the back of class and the addressee is facing the front of the room)
+                    # "your friend is in front of your laptop"
+                    # (Assuming the friend is one row up in the classroom)
                     if region.direction.positive:
                         return "in front of"
                     else:
--- a/adam/language_specific/english/english_language_generator.py
+++ b/adam/language_specific/english/english_language_generator.py
@@ -609,8 +609,6 @@ class SimpleRuleBasedEnglishLanguageGenerator(
                 return "on"
             elif region.distance == PROXIMAL and not region.direction:
                 return "to"
-            elif region.distance == INTERIOR:
-                return "in"
             elif region.direction == GRAVITATIONAL_UP:
                 return "over"
             elif region.direction == GRAVITATIONAL_DOWN:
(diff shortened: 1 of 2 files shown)


adam/language_specific/english/english_language_generator.py
```diff
@@ -606,8 +607,36 @@ def _preposition_for_region_as_goal(self, region: SituationRegion) -> str:
...
+                return "to"
+            elif region.distance == INTERIOR:
+                return "in"
+            elif region.direction == GRAVITATIONAL_UP:
+                return "over"
             elif region.direction == GRAVITATIONAL_DOWN:
                 return "under"
+            elif region.direction and self.situation.axis_info:
+                if not self.situation.axis_info.addressee:
+                    raise RuntimeError(
```
chris-jenkins-isi: You could include a test that covers this error.

adam/language_specific/english/english_language_generator.py
```diff
@@ -606,8 +607,36 @@ def _preposition_for_region_as_goal(self, region: SituationRegion) -> str:
...
                 return "on"
+            elif region.distance == PROXIMAL and not region.direction:
+                return "to"
+            elif region.distance == INTERIOR:
+                return "in"
+            elif region.direction == GRAVITATIONAL_UP:
+                return "over"
             elif region.direction == GRAVITATIONAL_DOWN:
                 return "under"
+            elif region.direction and self.situation.axis_info:
```
chris-jenkins-isi: Does this `elif` actually get hit for cases w/ addressees? In the event that `region.distance` is specified (in all cases other than DISTAL), this statement won't be reached.

chris-jenkins-isi: Follow up question: is it necessarily always the case that things in-front-of or behind other things are necessarily PROXIMAL? If not - the additional check on line 610 isn't sufficient to make sure addressee situations are handled.

LichMaster98: 1) Currently, this `elif` stack should be in proper order so that only the desired preposition is returned. Additionally, the addressee is only relevant information if we have failed to meet another condition for a preposition with the reference object. There is an addressee marked in all scenes regardless of if there is an object in the scene which is `FacingAddresseeAxis` 

2) The independent preposition curriculum defines "in front of" and "behind" as either `PROXIMAL` or `DISTAL` as I could say "the teacher is in front of my laptop" while I'm in the back of the classroom and the teacher is at the front and the preposition usage is still valid, even if not the most accurate to describe the location of the teacher. Likewise, I could say "my classmate is in front of my laptop" for a student directly in front of me one row up. (These could be defined as `PROXIMAL` and `DISTAL` in my understanding)

chris-jenkins-isi: Sounds good. I would recommend inserting a comment that notes why DISTAL doesn't need an entry in the elif stack for the benefit of future readers.

adam/language_specific/english/english_language_generator.py
```diff
@@ -606,8 +607,37 @@ def _preposition_for_region_as_goal(self, region: SituationRegion) -> str:
...
                 return "on"
+            elif region.distance == PROXIMAL and not region.direction:
+                return "to"
+            elif region.distance == INTERIOR:
+                return "in"
+            elif region.direction == GRAVITATIONAL_UP:
+                return "over"
             elif region.direction == GRAVITATIONAL_DOWN:
                 return "under"
+            # region.distance == DISTAL is not check as this does not define a specific preposition in scope for M9
```
gabbard: I would say "for phase 1"

adam/language_specific/english/english_language_generator.py
gabbard: I second @chris-jenkins-isi on the comment

adam/language_specific/english/english_language_generator.py
```diff
@@ -606,8 +607,37 @@ def _preposition_for_region_as_goal(self, region: SituationRegion) -> str:
                 # TODO: put constraints on the axis
             ):
                 return "on"
+            elif region.distance == PROXIMAL and not region.direction:
+                return "to"
+            elif region.distance == INTERIOR:
```
gabbard: The `in` and `over` branches seem to lack test coverage?

LichMaster98: Updated to include an `over` branch. `in` already has a unit test case which is succeeding, so I'm confused why that branch isn't covered?

gabbard: Aha - because it is redundant - we already check for `INTERIOR` on line 600.

Issue #568, opened 2020-02-11
When generating experiment logs, log the algorithm's performance


gabbard:
Currently, when we generate experiment logs, we show the users the learner's predictions compared to the "gold" descriptions. We would like to score accuracy for this in an automated way - how often was the "gold" description present among the learner's candidate descriptions?

Do this separately for the train and test sets.

It would also be nice to be able to do this without actually generating the dump files (e.g. for unit tests).

chris-jenkins-isi:
Closed by https://github.com/isi-vista/adam/pull/571

PR #571, opened 2020-02-11
Log prediction accuracy during learning / testing

chris-jenkins-isi:
Added prediction accuracy counter in html output (also available to be used in tests) References https://github.com/isi-vista/adam/issues/568

### DIFF
--- a/adam/experiment/log_experiment.py
+++ b/adam/experiment/log_experiment.py
@@ -15,7 +15,7 @@ from adam.curriculum.m6_curriculum import (
 from adam.curriculum.phase1_curriculum import _make_each_object_by_itself_curriculum
 from adam.curriculum.pursuit_curriculum import make_simple_pursuit_curriculum
 from adam.experiment import Experiment, execute_experiment
-from adam.experiment.observer import LearningProgressHtmlLogger
+from adam.experiment.observer import LearningProgressHtmlLogger, CandidateAccuracyObserver
 from adam.learner import LanguageLearner
 from adam.learner.object_recognizer import ObjectRecognizer
 from adam.learner.preposition_subset import PrepositionSubsetLanguageLearner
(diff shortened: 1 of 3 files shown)


adam/experiment/log_experiment.py
```diff
@@ -51,10 +51,10 @@ def log_experiment_entry_point(params: Parameters) -> None:
             name=experiment_name,
             training_stages=training_instance_groups,
             learner_factory=learner_factory_from_params(params, graph_logger),
-            pre_example_training_observers=[logger.pre_observer()],
-            post_example_training_observers=[logger.post_observer()],
+            pre_example_training_observers=[logger.pre_observers()],
```
gabbard: Why change the name to plural but still only return one thing?

adam/experiment/observer.py
```diff
@@ -90,6 +90,52 @@ def report(self) -> None:
...
+
+        descriptions_as_token_sequences = [
+            desc.as_token_sequence() for desc in predicted_descriptions
+        ]
+        if true_description.as_token_sequence() in descriptions_as_token_sequences:
+            self._num_predictions_with_gold_in_candidates += 1
+
+    def report(self) -> None:
+        logging.info(
+            f"{self.name}: accuracy of learner's predictions ('gold' description was in learner's candidates) "
```
gabbard: `pylint` will yell at you for using f-strings in logging (because it's inefficient if the log level is below `info`).  Prefer to put `%s`s in the string and then the things to intepolate as additionall arguments to `.info`.

adam/experiment/observer.py
```diff
@@ -90,6 +90,52 @@ def report(self) -> None:
...
+        )
+
+    def accuracy(self) -> float:
+        """
+        Return accuracy value of the number of predictions made where the 'gold' description was present in
+        the learner's candidate descriptions.
+        Returns: accuracy score (float)
+
+        """
+        if not self._num_predictions:
```
gabbard: Maybe return `Optional[float]` rather than throwing an exception. Or `NaN`.

adam/experiment/observer.py
```diff
@@ -284,14 +341,11 @@ def report(self) -> None:
         pass
 
 
-@attrs(slots=True)
-class HTMLLoggerPostObserver(
-    DescriptionObserver[SituationT, LinguisticDescriptionT, PerceptionT]
-):
+@attrs(slots=True)  # pragma: no cover
+class HTMLLoggerPostObserver(CandidateAccuracyObserver):  # type: ignore
```
gabbard: Prefer composition over inheritance. :-) . `HTMLLoggerPostObserver` should have a `has-a` relation to `CandidateAccuracyObserver`, not `is-a`.

adam/experiment/log_experiment.py
chris-jenkins-isi: Good catch. That was an edit that I went back and forth on.

PR #576, opened 2020-02-19
Add "go" with prepositions

### LINKED ISSUE #549: GO curriculum with prepositions

LichMaster98:
Closes #549

### DIFF
--- a/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
+++ b/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
@@ -4,7 +4,11 @@ from typing import Iterable
 from immutablecollections import immutableset
 from more_itertools import flatten
 
-from adam.axes import HorizontalAxisOfObject, FacingAddresseeAxis
+from adam.axes import (
+    HorizontalAxisOfObject,
+    FacingAddresseeAxis,
+    GRAVITATIONAL_AXIS_FUNCTION,
+)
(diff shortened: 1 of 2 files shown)


LichMaster98:
@elizlee By indicating a `SpatialPath` during the `DuringAction` should the generated english be something like "a dog goes over a box to a car" ?

elizlee:
@LichMaster98 I don't think it's necessary to describe the destination in these cases, but if we _do_ need a way to indicate in the language that e.g. `over` here refers to a path and not a destination, I believe adding _to X_ would be an appropriate choice.

adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
```diff
@@ -545,6 +579,11 @@ def _make_go_with_prepositions(num_samples: int = 5, *, noise_objects: int = 0):
     goal_object_hollow = standard_object(
         "goal_object_hollow", required_properties=[HOLLOW]
     )
+    path_object = standard_object(
+        "path_object",
+        required_properties=[CAN_HAVE_THINGS_RESTING_ON_THEM, HAS_SPACE_UNDER],
```
elizlee: Nitpick: are the required properties here necessary for _in front of/behind_? My thought was to make a `path_object` for each of the two sets of PPs.

PR #585, opened 2020-02-20
Support rendering of hierarchical objects (current only chair)

chris-jenkins-isi:
- Separate model pieces for chair (basis for adding separate model pieces for other hierarchical objects)
- Currently crashing on any scene with a hierarchical object unless it's a chair (the only one with its pieces currently provided)
- Provided that the appropriate geometry is available, IRREGULAR objects no longer pose an issue
- Minor adjustments to how scale factors are applied to object graph
- Updates to tests are inconsequential


adam/visualization/panda3d_interface.py
```diff
@@ -174,33 +200,36 @@ def add_model(
         # nested
         else:
             new_model.reparentTo(parent)
-        new_model.setPos(position[0], position[1], position[2])
+        # new_model.setPos(position[0], position[1], position[2])
```
LichMaster98: This comment looks like it could be removed unless there is another reason for it to stay here?

adam/visualization/panda3d_interface.py
```diff
@@ -210,7 +239,7 @@ def add_dummy_node(
             self.geo_nodes[new_node.name] = new_node
         else:
             new_node.reparentTo(parent)
-        new_node.setPos(*position)
+        # new_node.setPos(*position)
```
LichMaster98: Likewise for this comment?

adam/visualization/make_scenes.py
```diff
@@ -111,21 +111,33 @@ def main(params: Parameters) -> None:
         logging.info("SCALE: %s -> %s", model_name, scale.__str__())
 
     for i, scene_elements in enumerate(SceneCreator.create_scenes([make_curriculum()])):
-
+        if i < 1:
```
gabbard: What is this for?

chris-jenkins-isi: Updating to make the _specific_ selection of a scene for testing purposes explicit in the params file

gabbard: I think I need more context to understand.

chris-jenkins-isi: I broke a bunch of things by changing back to rendering separate objects for sub-objects

The new params option allows one to select a specific numbered scene if they don't want to bother scrolling through a bunch of them. I just pushed an update to fix the back-up option of rendering a basic Geon if there isn't a specific model available. So at the moment, rendering a table is still broken (haven't separated the model), but at least it doesn't crash the program. 

The current working example of a separated model is the chair.

PR #594, opened 2020-02-24
Add curriculum for `take x to y`

### LINKED ISSUE #552: Curriculum for "take X to Y"

elizlee:
Closes #552


adam/language_specific/english/english_language_generator.py
```diff
@@ -792,6 +792,8 @@ def relation_to_prepositional_modifier(
                                 preposition = "behind"
                         elif region.distance == PROXIMAL:
                             preposition = "beside"
+            elif region.distance == PROXIMAL:
```
chris-jenkins-isi: Would it make more sense to put this conditional up with the check if region.distance == INTERIOR at line 768 ?

That way all of the things where distance is the primary criteria are together before dealing with cases where direction is a factor as well.

Issue #591, opened 2020-02-21
Extract individual components of hierarchical objects


chris-jenkins-isi:
We would like to be able to manipulate individual components of hierarchical objects (or at least easily point to each piece) within the visualization system. Currently, all custom models are imported as one contiguous piece of geometry. The following is a process for exporting each piece separately. 

- **Download blender 2.73a** (highly recommend using a mouse with it)
- **Install the Yabee plugin**: https://github.com/09th/YABEE (allows for exporting models into Panda3d)

a/the **general process for exporting separate pieces of a model** is as follows:

	1.) Go into edit mode
		- If anything is already selected, you can toggle 'A' to select all / deselect all
	2.) Shift+R-click to select edges/vertices
		- Selecting all edges around a face selects that face
		- It may be helpful to enter fly mode (shift + F) to reposition the camera
		- You may need to delete edges / faces around your target area to more easily select it
		- Circle select (shortcut: C) is also very helpful, but you select with the left mouse button with it instead of the right.
	3.) When you have selected the entirety of the target geometry, press 'p' to split it into a new model (select split by selection)
	4.) Delete the rest of the original model
	5.) Save as a new .blend file (don't overwrite the original!)
	6.) Select all (A)
	7.) Export to .egg. (Warning: .egg files are ignored by our .gitignore!)
	8.) Check to see that exported .egg file actually contains vertices and etc
		- If not, just try again

- The original model _should_ be centered at the origin before any splitting occurs, so its separate pieces will be exported out at the correct positional offset from their fellows. 

**File naming conventions:** 
	
	- Prefer to list sides in x/y/z order (e.g. left_back_top)
	- If object has a 'canonical' facing (perspectival for animate objects, the direction that cars drive when not in reverse gear, etc), use that

	- Otherwise, orient descriptions based on the 'canonical' view of the object as seen when importing it into Panda3d using the utility script in panda3d_interface.py

	Example: table-leg_left_front (for the leg of a table)

chris-jenkins-isi:
`.blend` files are located at `/nas/home/cjenkins/adam-graphics/blender-models-sliced`

We still need to find / make a model for the cookie. Everything I'm seeing online is either not free, requires attribution, or is insanely detailed.

chris-jenkins-isi:
Another tip: in order for the models to receive the correct color settings at run time, you may need to remove any material currently in place on a model when exporting from Blender. 
[...]

PR #595, opened 2020-02-25
591 extract graphics components

chris-jenkins-isi:
Ongoing process of including models sliced into individual components. 
see #591 for process.

Completed:
- chair
- table
- dog

remaining:
- house (might need to be simplified)
- bird
- truck
- car
- liquids


jamart28:
Updated completed/remaining lists:

- Completed:
  - chair
  - table
  - dog
  - car

- remaining:
  - house (might need to be simplified)
  - bird
  - truck
  - liquids

jamart28:
It looks like all models are added is this ready for review? @chris-jenkins-isi

jamart28:
I just realized I misnamed car so I'm going to fix that as well

jamart28:
> We should also come up with some kind of trailer component for the truck, to house its wheels 4-7

This might be kind of a hacky solution but we could just add wheels to the current truck so that it has all 8 wheels. Not sure if this would work really

PR #600, opened 2020-02-27
Corrected implicit ground relation to consider before and after action relations

### LINKED ISSUE #596: Perception: dynamic prepositions: objects on GOAL also on ground
Screen Shot 2020-02-25 at 5 37 02 PM
In cases where objects end up _on_ or _in_ a goal object, we don't want the object to be touching the ground after the action.

jamart28:
- Changed `_perceived_ground_relations()` to take a relations parameter
so that before and after relations can be included
- Moved to `_perceived_ground_relations()` call from
`_perceive_action()` to `_real_do()` as to include all relations

Tested fix:
image

Closes #596

### DIFF
--- a/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
+++ b/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
@@ -1,6 +1,16 @@
 from enum import auto, Enum
 from itertools import chain
-from typing import AbstractSet, Dict, List, Mapping, Optional, Union, cast, MutableMapping
+from typing import (
+    AbstractSet,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Union,
+    cast,
+    MutableMapping,
+    Iterable,
+)
 
 from attr.validators import instance_of
 from immutablecollections import (
@@ -250,7 +260,7 @@ class _PerceptionGeneration:
 
         if not self._situation.actions:
             if self._include_ground:
-                self._perceive_ground_relations()
+                self._perceive_ground_relations(self._relation_perceptions)
             return PerceptualRepresentation.single_frame(
                 DevelopmentalPrimitivePerceptionFrame(
                     perceived_objects=self._object_perceptions,
@@ -271,6 +281,17 @@ class _PerceptionGeneration:
             self._perceive_relation(relation)
             for relation in self._situation.after_action_relations
         ]
+
+        if self._include_ground:
+            self._perceive_ground_relations(
+                relations=chain(
+                    self._relation_perceptions,
+                    explicit_before_relations,
+                    _action_perception.before_relations,
+                    explicit_after_relations,
+                    _action_perception.after_relations,
+                )
+            )
 
         first_frame = DevelopmentalPrimitivePerceptionFrame(
             perceived_objects=self._object_perceptions,
@@ -705,9 +726,9 @@ class _PerceptionGeneration:
             self._objects_to_perceptions
         )
 
-    def _perceive_ground_relations(self):
-        objects_to_relations = self._objects_to_relations()
-        perceived_ground: ObjectPerception = None
+    def _perceive_ground_relations(self, relations: Iterable[Relation[ObjectPerception]]):
+        objects_to_relations = self._objects_to_relations(relations)
+        perceived_ground: Optional[ObjectPerception] = None
         for object_ in self._object_perceptions:
             if self._object_perceptions_to_ontology_nodes[object_] == GROUND:
                 perceived_ground = object_
@@ -745,10 +766,10 @@ class _PerceptionGeneration:
                     )
 
     def _objects_to_relations(
-        self
+        self, relations: Iterable[Relation[ObjectPerception]]
     ) -> ImmutableSetMultiDict[ObjectPerception, Relation[ObjectPerception]]:
         return immutablesetmultidict(
-            (relation.first_slot, relation) for relation in self._relation_perceptions
+            (relation.first_slot, relation) for relation in relations
         )
 
     @attrs(frozen=True, slots=True)
@@ -788,9 +809,6 @@ class _PerceptionGeneration:
             conditions=action_description.postconditions,
             action_object_variables_to_object_perceptions=action_objects_variables_to_perceived_objects,
         )
-
-        if self._include_ground:
-            self._perceive_ground_relations()
 
         return _PerceptionGeneration._ActionPerception(
             before_relations=immutableset(chain(enduring_relations, before_relations)),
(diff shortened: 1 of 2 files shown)


jamart28:
New test:
image

Ground is still present but no longer persists once baby is on truck

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -730,25 +773,27 @@ def _perceive_ground_relations(self):
                             ):
                                 object_is_on_something = True
                     if not object_is_on_something:
-                        self._relation_perceptions.append(
+                        ground_relations.append(
                             on(
                                 self._objects_to_perceptions[situation_object],
                                 perceived_ground,
-                            )
+                            )[0]
```
LichMaster98: Why are you only appending the first relation returned by the `on` function? The relationship for the `contacts` is symmetric on purpose.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -730,25 +773,27 @@ def _perceive_ground_relations(self):
...
                         )
                 else:
-                    self._relation_perceptions.append(
+                    ground_relations.append(
                         on(
                             self._objects_to_perceptions[situation_object],
                             perceived_ground,
-                        )
+                        )[0]
```
LichMaster98: Same selection of `[0]` here on the return of `on`

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -272,12 +284,41 @@ def _real_do(self) -> PerceptualRepresentation[DevelopmentalPrimitivePerceptionF
...
+                    relations=chain(
+                        self._relation_perceptions,
+                        explicit_before_relations,
+                        _action_perception.before_relations,
+                        explicit_after_relations,
+                        _action_perception.after_relations,
+                    )
+                )
+            )
+            before_ground = self._perceive_ground_relations(
```
LichMaster98: Add a comment indicating why we need to handle `before_ground` and `after_ground` separately for a future reader.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -730,25 +773,27 @@ def _perceive_ground_relations(self):
...
-                    self._relation_perceptions.append(
+                    ground_relations.append(
                         on(
                             self._objects_to_perceptions[situation_object],
                             perceived_ground,
-                        )
+                        )[0]
                     )
 
+        return ground_relations
```
LichMaster98: Can you return the `ground_relations` in an `immutableset` I don't think passing out a list here is a big deal because it's immediately going to get chained into an `immutableset` before it would be iterated over but we prefer to return immutable structures for determinism purposes.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
jamart28: When I was running the tests all I got from `on()` was a tuple with the relation I wanted on the inside and nothing else. @LichMaster98

jamart28: Looking at `on()` since we aren't passing `iterable`s (as far as I can tell) we should always get a `tuple` of size 1 with the relation we want.

LichMaster98: `on` returns a tuple but you should want to append both i.e:

`on(ground, book)` -> `(Relation(IN-REGION, ground, Region(book, distance=EXTERIOR_BUT_IN_CONTACT)), Relation(IN-REGION, book, Region(ground, distance=EXTERIOR_BUT_IN_CONTACT))` 

The program may return in the opposite order I don't recall but both of these relationships are true and should exist in the final output.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -730,25 +773,27 @@ def _perceive_ground_relations(self):
                             ):
                                 object_is_on_something = True
                     if not object_is_on_something:
```
elizlee: I wanted to chime in here with the reminder that we want to handle objects going _in_ something as well.
Screen Shot 2020-02-27 at 4 01 29 PM

LichMaster98: Good catch, thanks @elizlee 

I think it's safe to assume if something is `in` an object the default assumption should be the ground isn't supporting it. (Counterexample: A ball in an upsidedown cup) but that feels like a situation where we should template the `on(ball, ground)` relationship rather than assuming the default is handled that way.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
jamart28: When I was reviewing the error I found that the Tuple only had one relation each time. Following:
https://github.com/isi-vista/adam/blob/82e8089a4d2d9f1cd31178217b18854e22576deb/adam/ontology/phase1_ontology.py#L574
I found:
https://github.com/isi-vista/adam/blob/82e8089a4d2d9f1cd31178217b18854e22576deb/adam/relation.py#L245-L261
which by my understanding only returns a tuple larger than 1 if an actual iterable is used for the arguments which we don't use here. Let me know if there is anything I'm not seeing or if this changed somewhere that I'm again not seeing.

LichMaster98: @jamart28 Apologies, on isn't symmetric I'm just crazy. Regardless you shouldn't need to explicitly select 0. I think you may want to `extend` rather than `append`

jamart28: I can do that.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
jamart28: Fixed:
image

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -716,7 +764,7 @@ def _perceive_ground_relations(self):
             if situation_object.ontology_node != GROUND:
                 if self._objects_to_perceptions[situation_object] in objects_to_relations:
                     # If this object is not on anything else, it should be on the ground
-                    object_is_on_something = False
+                    object_is_on_ground = True
```
LichMaster98: Can you refactor this variable to `object_is_on_ground_or_in_object`

gabbard:
@jamart28 : this needs a test before review

jamart28:
@gabbard Added two new tests. Let me know if these test what is necessary or not.

Issue #603, opened 2020-02-27
Need a way to scale people (as in, baby should be smaller than Mom or Dad)


chris-jenkins-isi:
Right now all 'person' objects are exactly the same size. (It is also the case that we are using the same geometry for adults as well as the baby).

This could be addressed in a couple of ways:

1.) We could change the debug_handle for the object_perception of the baby and of adults to something more specific than just 'person_#'

2.) We could carve out a special case handling for scaling 'person' models to check if they have the binary properties 'is-mom' or 'is-dad', the absence of either of which could be interpreted as the nonexistent property 'is-baby'.

LichMaster98:
@chris-jenkins-isi See: https://github.com/isi-vista/adam/issues/83 We wanted to add this information in via a Relation that would assert that `bigger_than([MOM, DAD], BABY)` in the situation. However the current runtime implementation of this issue is abysmally slow and we haven't returned to this issue. 

Granted this doesn't help for a quick patch but I think the Relationship information would be the ideal solution.

chris-jenkins-isi:
A hacky approach would be changing the method `_instantiate_object_schema` in `high_level_semantics_situation_to_developmental_primitive_perception.py` to take the debug_handle for an `ObjectPerception` from its related SituationObject's handle, instead of from the related `ObjectStructuralSchema`

This doesn't seem to cause any test failures outside of a couple of obvious ones (looking for an object with handle "person" in a frame and instead finding "dad")

gabbard:
Since we are willing to incur some technical debt to have the best system we can for the M9 demo, I would do option (2) for now and make an issue to handle it more nicely later.

gabbard:
Visualization is on hold.

PR #606, opened 2020-02-28
603 scaling people

chris-jenkins-isi:
This seemed like the simplest point to intervene in the process of creating an `ObjectPerception` to ensure that its `debug_handle` corresponds with the particular type of human involved. (Mom, Dad, Baby). 

For context, the visualization system relies on the debug_handle strings of `ObjectPerception`s (strictly speaking, not information given to any learner) in order do things that depend on the object's type such as loading the correct geometry for the object type, or scaling it according to its type, etc. In the case of people (see #603) `ObjectPerceptions` for particular sorts of people have their debug_handles set to their supertype: Person. 

I can add some explanatory comments before merging, but I wanted to check if this overall is in _bad idea_ territory.

### DIFF
--- a/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
+++ b/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
@@ -1049,6 +1049,9 @@ class _PerceptionGeneration:
 
         # find axes for the object
         if situation_object:
+            # use the debug handle from the situation object if it is available, as it is more
+            # specific in the case of people (e.g. mom, dad, baby) than the debug handle
+            # generated from the ObjectStructuralSchema
             debug_handle = (
                 situation_object.ontology_node.handle + "_" + debug_handle.split("_")[1]
             )
--- a/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
+++ b/adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
@@ -1030,9 +1030,17 @@ class _PerceptionGeneration:
         situation_object: Optional[SituationObject] = None,
     ) -> ObjectPerception:
 
-        debug_handle = self._object_handle_generator.subscripted_handle(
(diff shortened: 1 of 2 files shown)


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -1049,6 +1049,9 @@ def _instantiate_object_schema(
 
         # find axes for the object
         if situation_object:
+            debug_handle = (
```
LichMaster98: If we are going to make this change can you add a comment explaining why.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -1049,6 +1049,12 @@ def _instantiate_object_schema(
 
         # find axes for the object
         if situation_object:
+            # use the debug handle from the situation object if it is available, as it is more
+            # specific in the case of people (e.g. mom, dad, baby) than the debug handle
+            # generated from the ObjectStructuralSchema
+            debug_handle = (
+                situation_object.ontology_node.handle + "_" + debug_handle.split("_")[1]
```
gabbard: Some questions:
* What if the debug handle lacks a `_`?  
* If the ontology node handle and the debug handle are the same, do we want to duplicate? If this something which frequently happens?
* Does this make the debug logs any less readable?

chris-jenkins-isi: By the look of it (Running the m6 experiment script), the debug logs mostly refer to the spoken utterance (seemingly unaffected by this change), or else refer to specific subcomponents of an object (e.g. head_0 - for the head of a person).

prior to this change, the ontology node handle (of the schema) and debug handle appear to always be the same. After this change, the debug handle and ontology node handle (of the schema) only differ in the case of people.

PR #608, opened 2020-02-28
Translating subject (speaker) as "I" when using HAS

### LINKED ISSUE #605: English Generator fails to translate "I" in relation as verb
We fail to correctly translate "I" as the subject as a sentence: 
("me", "have", "my", "ball")
is a resulting generation of a has(speaker, ball)

aakum03:
Closes #605 
My attempt at this task:
I think this has to do with `HAS` being a `RELATION` and not an `ACTION` even though it is a verb. As such, the speaker is never recognized as the `NOMINAL_SUBJECT`, resulting in the default `ME` being used.
I added a special case in noun_for_object() in the english_language_generator:
This seems to fix it, and I don't see any obvious cases where it affects the other parts:
```
# Check if the situation object is the speaker
            if IS_SPEAKER in _object.properties:
                if syntactic_role_if_known == NOMINAL_SUBJECT:
                    noun_lexicon_entry = I
                elif any(
                    relation
                    for relation in self.situation.always_relations
                    if (
                        relation.relation_type == HAS
                        and any(
                            property_ in relation.first_slot.properties
                            for property_ in [IS_SPEAKER]
                        )
                        and len(self.situation.actions) == 0
                    )
                ):
                else:
                    noun_lexicon_entry = ME
```
NOTE: I only added the elif case

With this, all the test cases containing a `HAS` relation in test_english_language_generator pass including a test I added:
```
def test_i_have_my_ball():
    baby = situation_object(BABY, properties=[IS_SPEAKER])
    ball = situation_object(BALL)
    situation = HighLevelSemanticsSituation(
        ontology=GAILA_PHASE_1_ONTOLOGY,
        salient_objects=[baby, ball],
        always_relations=[Relation(HAS, baby, ball)],
    )

    assert only(
        _SIMPLE_GENERATOR.generate_language(situation, FixedIndexChooser(0))
    ).as_token_sequence() == ("I", "have", "my", "ball")
```

What do you think?

### DIFF
--- a/adam/language_specific/english/english_language_generator.py
+++ b/adam/language_specific/english/english_language_generator.py
@@ -233,12 +233,18 @@ class SimpleRuleBasedEnglishLanguageGenerator(
             if IS_SPEAKER in _object.properties:
                 if syntactic_role_if_known == NOMINAL_SUBJECT:
                     noun_lexicon_entry = I
-                elif any(relation for relation in self.situation.always_relations if
-                         (relation.relation_type == HAS and
-                          any(property_ in relation.first_slot.properties
-                              for property_ in [IS_SPEAKER]) and
-                          len(self.situation.actions) == 0)
-                         ):
+                elif any(
+                    relation
+                    for relation in self.situation.always_relations
+                    if (
+                        relation.relation_type == HAS
+                        and any(
+                            property_ in relation.first_slot.properties
+                            for property_ in [IS_SPEAKER]
+                        )
+                        and len(self.situation.actions) == 0
+                    )
+                ):
                     noun_lexicon_entry = I
                 else:
                     noun_lexicon_entry = ME
--- a/adam/language_specific/english/english_language_generator.py
+++ b/adam/language_specific/english/english_language_generator.py
@@ -242,7 +242,7 @@ class SimpleRuleBasedEnglishLanguageGenerator(
                             property_ in relation.first_slot.properties
                             for property_ in [IS_SPEAKER]
                         )
-                        and len(self.situation.actions) == 0
+                        and not self.situation.actions
                     )
                 ):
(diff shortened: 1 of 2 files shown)


adam/language_specific/english/english_language_generator.py
```diff
@@ -228,12 +228,24 @@ def _noun_for_object(
                     f"an ontology node currently: {_object}"
                 )
             # TODO: we don't currently translate modifiers of nouns.
```
LichMaster98: As this issue has been resolved I think we can also remove this TODO.

adam/language_specific/english/english_language_generator.py
```diff
@@ -228,12 +228,24 @@ def _noun_for_object(
...
-            # Issue: https://github.com/isi-vista/adam/issues/58
 
             # Check if the situation object is the speaker
             if IS_SPEAKER in _object.properties:
                 if syntactic_role_if_known == NOMINAL_SUBJECT:
                     noun_lexicon_entry = I
+                elif any(
```
LichMaster98: Add a comment here which indicates why we have to special case the subject "I" for the relationship has being the verb in our sentence.

adam/language_specific/english/english_language_generator.py
aakum03: Removed

adam/language_specific/english/english_language_generator.py
aakum03: Added. Should I make this more concise?

LichMaster98: @aakum03 I think this is fine.

gabbard:
This works fine for now, but can you make an issue for handling verbally-expressed relations for elegantly later, and point to this as a hack which should be removed at that time?

PR #612, opened 2020-03-02
557 jump with new prepositions

### LINKED ISSUE #557: JUMP curriculum with new prepositions
"X jumps up"

X jumps and lands somewhere in relation to Y:
_in, on, beside, behind, in front of_

elizlee:
Closes #557


adam/language_specific/english/english_language_generator.py
```diff
@@ -696,6 +697,10 @@ def _collect_action_modifiers(
                     modifiers.append(
                         (ADVERBIAL_MODIFIER, DependencyTreeToken("down", ADVERB))
                     )
+                elif action.action_type == JUMP:
+                    modifiers.append(
+                        (ADVERBIAL_MODIFIER, DependencyTreeToken("up", ADVERB))
```
chris-jenkins-isi: Should jumping over / in / in front of / behind / beside things be covered in the language generator?

tests/language_specific/english/test_english_language_generator.py
```diff
@@ -768,6 +768,24 @@ def test_bird_flies_up():
     assert generated_tokens(situation) == ("a", "bird", "flies", "up")
 
 
+def test_jump_up():
```
chris-jenkins-isi: I take it that this test and the one below it cover the two basic formats: jumping w/o an explicit goal location and jumping with a goal? If that's the case, these tests should be sufficient!

adam/language_specific/english/english_language_generator.py
elizlee: Those prepositions are covered in `preposition_for_region_as_goal`; they're not specific to `jump`

PR #622, opened 2020-03-06
554 throw curriculum with prepositions

### LINKED ISSUE #318: Template for two people throwing things to each other.
I expect problems with binding two manipulators in a scene, which is why this is split off from #208.

### LINKED ISSUE #554: THROW curriculum with prepositions
Related: #318 

X throws Y where Y lands somewhere in relation to Z (possibly intending for Z to catch Y):
_to, in, on, beside, behind, in front of, under_

X throws Y with a path across Z:
_over, under_

X throws Y in some vertical direction:
_up, down_

elizlee:
Closes #554


elizlee:
@chris-jenkins-isi: up/down are generated in the phase1 curriculum.

Since the SpatialPath for "up" is _away from the ground_, that is something to think about. My initial reasoning was 1. this will generate "up" in the language and 2. there is still an implicit `goal`, presumably resting on the ground, that it must be near at the end of the action.

Potentially I could change this such that "up" isn't quite the opposite of "down," but an optional modifier when the goal region is on top of something bigger than the thrower.

PR #630, opened 2020-03-12
Use of Action description to fix objects in place during positioning (proof of concept)

chris-jenkins-isi:
Specifically: 
- Scenes involving GO actions

Limitations:
- Implementation needed in positioning system for distance=in (e.g. "the dog goes in the car")
- Only this specific combination of action + arguments is handled so far.

### DIFF
--- a/adam/visualization/make_scenes.py
+++ b/adam/visualization/make_scenes.py
@@ -15,6 +15,7 @@ from typing import (
     Any,
     Generator,
     Mapping,
+    Dict,
 )
 from functools import partial
 
@@ -25,7 +26,9 @@ import numpy as np
 import logging
 
 # currently useful for positioning multiple objects:
-from adam.curriculum.phase1_curriculum import _make_take_curriculum as make_curriculum
+from adam.curriculum.verbs_with_dynamic_prepositions_curriculum import (
+    _make_go_with_prepositions as make_curriculum,
+)
 
 
 import attr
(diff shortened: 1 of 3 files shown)


adam/visualization/make_scenes.py
```diff
@@ -25,7 +26,9 @@
 import logging
 
 # currently useful for positioning multiple objects:
-from adam.curriculum.phase1_curriculum import _make_take_curriculum as make_curriculum
```
elizlee: Would we not use `phase1_curriculum` here anymore, or the switch just for the proof-of-concept?

adam/visualization/make_scenes.py
```diff
@@ -407,7 +425,26 @@ def create_scenes(
                     ObjectPerception,
                     List[Optional[Union[RgbColorPerception, OntologyNode]]],
                 ] = defaultdict(list)
-                # we only care about the perception at the moment
+
+                # TODO: change HighLevelSemanticsSituation to keep track of objects with a full (disambiguating) handle
```
elizlee: If there is an issue for this, can you link it here?

adam/visualization/make_scenes.py
chris-jenkins-isi: That's the idea, just so that if you run `adam.visualization.make_scenes parameters/make_scenes.params` a reasonable scene comes up to demonstrate whatever the latest changes are.

elizlee: Got it

PR #647, opened 2020-03-17
Alteration to the process of freezing action goal to include non-movi…

### LINKED ISSUE #646: Freeze positions of Goal objects for other verb types.
Follow-up on #630 , in pursuit of #621

chris-jenkins-isi:
…ng action agent / theme. Changed PositioningModel to facilitate this better.

Will need to check if this generalizes well to additional action types. 

Closes #646 

Example queued up with curriculum selection is:

Throwing + PP. In the first frame, all objects are parameters, the object to be thrown is in the region of the person's hand. In the second frame, the person and the goal object for the throw are stationary, only the thrown object is repositioned.

### DIFF
--- a/adam/visualization/make_scenes.py
+++ b/adam/visualization/make_scenes.py
@@ -56,6 +56,7 @@ from adam.perception.developmental_primitive_perception import (
     Relation,
 )
 from adam.ontology import OntologyNode
+from adam.ontology.phase1_ontology import AGENT, THEME, GOAL
 from adam.relation import IN_REGION
 from adam.situation import SituationObject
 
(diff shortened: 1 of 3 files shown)


adam/visualization/make_scenes.py
```diff
@@ -644,33 +643,45 @@ def objects_to_freeze(
...
-                    ):
-                        frozen_objects.append(
-                            situation_obj_to_handle[goal.reference_object]
-                        )
+        agent: Optional[SituationObject] = None
+        theme: Optional[SituationObject] = None
+        goal: Optional[Union[SituationObject, Region[SituationObject]]] = None
+        for (ontology_node, object_or_region) in action.argument_roles_to_fillers.items():
+
+            if ontology_node.handle == "goal":
```
LichMaster98: I think you should just import GOAL, AGENT, and THEME to compare with rather than using the debug handles

chris-jenkins-isi: addressed in latest commit.

PR #671, opened 2020-03-21
Various tweaks from triaging which verbs work and which don't

gabbard:


gabbard:
Closing in favor of https://github.com/isi-vista/adam/pull/679 , which is an extension of this.

PR #711, opened 2020-03-24
specifying not on ground

### LINKED ISSUE #597: Rules for automatically putting things in contact with the ground are not aggressive enough.
Currently if we don't specify where an object is "on" we automatically assume the ground, however in the case we explicitly don't want the object to be supported by anything we can't assert this relationship.

The capability implemented in `Situation` should be replicated in `SituationTemplate` for ease of use

### LINKED ISSUE #706: If an object is in contact with some other object, it shouldn't also be automatically put in contact with the ground
For example, the given object here (SLOT2) is marked in contact with the ground,

denizbeser:
Closes #597 
Closes #706


adam/curriculum/phase1_curriculum.py
```diff
@@ -354,6 +355,9 @@ def make_fall_templates() -> Iterable[Phase1SituationTemplate]:
             syntax_hints=[USE_ADVERBIAL_PATH_MODIFIER]
             if use_adverbial_path_modifier
             else [],
+            before_action_relations=[
+                rel.negated_copy() for rel in on(arbitary_object, ground)
```
gabbard: You can do just `negate(on(arbitrary_object, ground))`.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -763,8 +771,10 @@ def _perceive_ground_relations(
         for situation_object in self._situation.all_objects:
             if situation_object.ontology_node != GROUND:
                 if self._objects_to_perceptions[situation_object] in objects_to_relations:
-                    # If this object is not on anything else, it should be on the ground
+                    # If this object is not on anything else, it should be on the ground, unless it's explicitly
+                    # specified to be unsupported
                     object_is_on_or_in_something = False
+                    supported = True
```
gabbard: As long as we are fixing this, we might as well fix #706 as well.  The strategy for doing this:

- first check for the negated "on the ground" relation and bail if it is present
- then initialized `supported=False` and loop over the relations. If it is asserted to be in contact with any other object, set `supported=True`
- make the object on the ground if `supported` is `False`

adam/curriculum/phase1_curriculum.py
denizbeser: Fixed.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
denizbeser: don’t we already do 706 in lines 780-790?

PR #723, opened 2020-03-25
Recognize ground for preposition and verb purposes

### LINKED ISSUE #717: Preposition learner is learning many "particular object on ground" patterns instead of "X on Y"
e.g. `a_box_on_the_ground.pdf` in the pattern graph dumps

LichMaster98:
Closes #717

### DIFF
--- a/adam/learner/object_recognizer.py
+++ b/adam/learner/object_recognizer.py
@@ -200,23 +200,15 @@ class ObjectRecognizer:
         else:
             object_names_to_patterns = self._object_names_to_static_patterns
 
-        candidate_object_subgraphs = self.extract_candidate_objects(perception_graph)
-
-        for candidate_object_graph in candidate_object_subgraphs:
-            root_node = self._get_root_object_perception(
-                perception_graph.copy_as_digraph(),
-                candidate_object_graph._graph.nodes,  # pylint:disable=protected-access
-            )
-            # We special case handling the ground perception
-            # Because we don't want to remove it from the graph, we just want to use it's
-            # Object node as a recognized object
-            if root_node == GROUND_PERCEPTION:
+        # We special case handling the ground perception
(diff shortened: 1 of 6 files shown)


adam/learner/object_recognizer.py
```diff
@@ -202,6 +203,34 @@ def match_objects(
         candidate_object_subgraphs = self.extract_candidate_objects(perception_graph)
 
         for candidate_object_graph in candidate_object_subgraphs:
+            root_node = self._get_root_object_perception(
```
gabbard: Rather than do this on every `candidate_object_graph`, can we move it outside the loop and just search the graph for the `GROUND_PERCEPTION` node?

adam/learner/object_recognizer.py
```diff
@@ -199,9 +200,33 @@ def match_objects(
         else:
             object_names_to_patterns = self._object_names_to_static_patterns
 
+        # We special case handling the ground perception
+        # Because we don't want to remove it from the graph, we just want to use it's
+        # Object node as a recognized object
+        for node in graph_to_return._graph.nodes:  # pylint:disable=protected-access
```
gabbard: This is a big enough chunk of logic that we should pull it into its own method.

adam/learner/object_recognizer.py
```diff
@@ -199,9 +200,33 @@ def match_objects(
         else:
             object_names_to_patterns = self._object_names_to_static_patterns
 
+        # We special case handling the ground perception
+        # Because we don't want to remove it from the graph, we just want to use it's
```
gabbard:
```suggestion
        # Because we don't want to remove it from the graph, we just want to use its
```

gabbard: Also it might be good to describe in the comment the situation which motivated us to do this.

PR #755, opened 2020-04-20
619 add gaze arrow

jamart28:
Currently testing. This will be marked ready for review once quickly tested.


adam/visualization/panda3d_interface.py
```diff
@@ -382,6 +395,11 @@ def clear_debug_nodes(self) -> None:
             node.remove_node()
         self.debug_bounding_boxes = []
 
+    def clear_gaze_arrows(self) -> None:
+        for node in self.gaze_arrows:
+            node.remove_node()
+        self.gaze_arrows = []
+
```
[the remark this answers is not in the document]
chris-jenkins-isi: That way it'll get cleaned up w/ the rest of the scene

adam/visualization/make_scenes.py
```diff
@@ -330,6 +336,14 @@ def main(
                         repositioned_map.name_to_scale[name],
                     )
 
+            if gaze_arrows:
+                for _, handle in scene_elements.situation_object_to_handle.items():
```
chris-jenkins-isi, line 340: The issue here is that not every object has the `gazed_at` property. (Annoyingly, in the basic curriculum everything is gazed at. I tried 

```from adam.curriculum.verbs_with_dynamic_prepositions_curriculum import (
     _make_take_with_prepositions as make_curriculum,
)
```
and found a scene where only some of the objects were being looked at. 

So you'll want to use the `SceneElements.property_map` to check which things have the binary property gazed-at.

chris-jenkins-isi: I would suggest changing the key of that `property_map` (line 464) from `ObjectPerception` just to a `str` (using `ObjectPerception.debug_handle` as the key.

PR #762, opened 2020-05-04
Implement MultiDiGraph as perception and pattern graphs

### LINKED ISSUE #751: Consider switching to a multi-digraph representation for perception graphs
We currently cannot represent multiple relationships between the same objects, which sometimes leads to errors when learning actions. We should asses what the consequences would be of switching to `MultiDiGraphs`. Will our matching algorithms still work? Will it slow down a lot?

LichMaster98:
Closes #751


adam/perception/perception_graph.py
```diff
@@ -357,30 +370,31 @@ def from_dynamic_perceptual_representation(
...
 
-        for (source, target) in digraph.edges():
-            unwrapped_label = digraph.edges[source, target]["label"]
+        for (source, target, edge) in multidigraph.edges():
+            unwrapped_label = edge["label"]
             temporally_scoped_label = TemporallyScopedEdgeLabel.for_dynamic_perception(
                 unwrapped_label, when=temporal_scopes
             )
-            digraph.edges[source, target]["label"] = temporally_scoped_label
+            multidigraph.edges[source, target, edge]["label"] = temporally_scoped_label
```
LichMaster98, line 395: This may be the wrong way to do this update on a multidigraph...

Issue #570, opened 2020-02-11
Add more detailed relationships between sub-objects


gabbard:
In object schemata, we currently only specify sub-object relations which are either "contacts" or specify directions relative to global axes (e.g. gravitational up).

We would like to also specify relations of sub-objects with respect to the parent object's local axes - for example, the relative positions of table legs.

This would not only allow us to better distingish objects, but would also substantially speed up object matching, where the exponential search for alignment of identical sub-objects (e.g. legs, wheels) takes up most of the time.

There are a few difficulties with implementing this:
* in the perception generator, we don't currently (I think) remap abstract to concrete axes for regions arising from sub-object relationships. However, each situation object in the perception generator does track a correspondence between "Platonic" schema axes and concrete object axes, so this shouldn't be too hard to do.
* Currently we have many sub-object relations defined in terms of the global gravitational axis.  We would ideally want to make all sub-object relations in terms of parent object local axes so there is no inconsistency to confuse the learning algorithm.

For the moment, this does not contribute directly to our Phase 1 goals and is time-consuming, so it's going on the backlog.

This is related to https://github.com/isi-vista/adam/issues/316

tag @chris-jenkins-isi

PR #764, opened 2020-05-04
Support spatial relations between sub-objects which are sensitive to sub-object axes

gabbard:
Enables https://github.com/isi-vista/adam/issues/570


adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -1058,16 +1059,41 @@ def _perceive_object_or_region_relation_filler(
         else:
             return action_object_variables_to_object_perceptions[slot_filler]
 
+    @attrs(frozen=True, slots=True)
+    class _InstantiateObjectSchemaReturn:
```
gabbard, line 1063: @denizbeser @LichMaster98 : I would read this module's changes first.

adam/relation_dsl.py
```diff
@@ -0,0 +1,272 @@
+"""
```
gabbard, line 1: This was all just moved to break an import cycle and does not need review.

adam/relation_dsl.py
```diff
@@ -0,0 +1,272 @@
...
+                    for arg1 in arg1s
+                    for arg2 in arg2s
+                ),
+            ]
+        )
+
+    return dsl_relation_function
+
+
+def located(
```
gabbard, line 228: Except this one.

adam/ontology/phase1_ontology.py
```diff
@@ -1396,34 +1400,62 @@ def _make_human_arm_segment():
```
LichMaster98, line 1400: We should raise an issue to check all our defined schema to add in this `located` information with regards to left/right and front/back as many schema have distinctions like this which are not currently expressed.

gabbard: @LichMaster98 : I think we can continue to use #570 for that (which is why I was careful to say "enables #570" rather than "closes" in the PR description. :-) ).

Issue #742, opened 2020-04-08
Curriculum for imprecise descriptions


gabbard:

LichMaster98:
> We therefore plan to learn the remaining frequent words: big, little, tall, short, fast, slow, and (if there is enough time to enhance our temporal reasoning component) new.

# Imprecise Size Descriptions
We plan to implement a curriculum for 4 imprecise size descriptions:
- [ ] Big
- [ ] Little
- [ ] Tall
- [ ] Short

In preparing this curriculum we need to decide how rich of a size environment we want available to our learner as currently, we have three outstanding issues regarding implicit size between objects:

1. [Relative Size Perceptions](#251)
2. [Perceiving All Implicit Size - Not just relative to learner](#215)
3. [Original Issue for Generating Implicit Size and Spatial Relations](#83)

Option 3 also contains a suggested change for the implicit size function. 

The above will become more relevant I suspect when learning rather than at curriculum generation time. 

Our target learner for these descriptions is probably the attribute learner for "big truck", "small ball", etc. However, I could also see examples where we need some type of average or reference for the size information, in the current case I assume that would be the learner. In that case, we may need the hypothesis space of the relationship learner however that would change our text to something closer to "a ball smaller than you" -- We could learn still "small ball" from this relationship but our current implementation wouldn't allow for it.

# Imprecise Temporal Descriptions
We plan to implement at minimum 2 imprecise temporal descriptions:
- [ ] Fast
- [ ] Slow

In order to learn these we are going to modify our `SpatialPath` of an action to contain temporal qualities indicating various different ways of making subtle changes to the primary action.

- [ ] New
Learning new will require adapting a situation, template, and perception to mark an object as *only visible* in one of the two frames of a situation.

LichMaster98:
From our meeting, these thoughts below are also relevant for #743:

These attributes indicate either a situation where there is more than one of the object in the scene, say dogs, that the "big dog" is the biggest from the group of dogs we are looking at while "big dog" with only one dog in the scene implies that the dog is bigger than our prototypical assumption of the size of a dog.

Attributes that can be scalars can get the English "very" added to them. For example "very fast".

These attributes are closer to relationships for our learner capabilities.

Does all of this sound accurate to what we discussed @gabbard, @charles-yang-upenn, and @mitchmarcus ?

LichMaster98:
@denizbeser @gabbard I believe this issue is closed now?

gabbard:
@LichMaster98 : I believe so

PR #774, opened 2020-05-21
Add Imprecise Temporal Descriptions Curriculum

LichMaster98:
Addresses part of #742

### DIFF
--- a/adam/curriculum/curriculum_utils.py
+++ b/adam/curriculum/curriculum_utils.py
@@ -8,7 +8,13 @@ from adam.language_specific.english.english_language_generator import (
     GAILA_PHASE_1_LANGUAGE_GENERATOR,
 )
 from adam.ontology import OntologyNode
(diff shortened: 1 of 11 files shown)


LichMaster98:
While trying to rebase this branch onto master to account for the changes in #767 we now have an issue where I was trying to avoid having to specify the path operator in this PR as to not replace the information contained in the ontology entry for the action. @gabbard and I discussed simply collapsing SpatialPath's which share a reference object into one path. While this would work for the `perceptual representation` portion of the code, this solution will not work for the language generator as the only `Path` information available at that point is the SpatialPath defined by the curriculum designer as the situation only contains the salient (or possibly salient) path information.

That leaves me with the options to log when we 'skip' the `None` operator case in the language generator (a non-robust solution as I think we'd rather crash here), require the curriculum designer to know which operator to use in the event they want to assert this type of property on the `SpatialPath` (Could be made easier via a macro function), or change the location of this action property information entirely.

LichMaster98:
Taking this back to draft so I remember to implement the unification of SpatialPath's in the perceptual representation.

Issue #778, opened 2020-06-01
Creating Chinese Lexicon


denizbeser:
Here is the lexicon file for English:
https://github.com/isi-vista/adam/blob/master/adam/language_specific/english/english_phase_1_lexicon.py
We need a similar file for Chinese. Objects could be a good start.

PR #780, opened 2020-06-02
778 chinese lexicon

paynesa:
#778: creates language_specific/chinese directory with Chinese lexicon and dummy syntax file

### DIFF
--- a/adam/language_specific/chinese/chinese_phase_1_lexicon.py
+++ b/adam/language_specific/chinese/chinese_phase_1_lexicon.py
@@ -73,18 +73,8 @@ from adam.ontology.phase1_ontology import (
     DARK_BROWN,
 )
 
-ME = LexiconEntry(
-    "wo3",
-    NOUN,
-    plural_form="wo3 men",
-    intrinsic_morphosyntactic_properties=[FIRST_PERSON],
-)
-YOU = LexiconEntry(
-    "ni3",
-    NOUN,
-    plural_form="ni3 men",
-    intrinsic_morphosyntactic_properties=[SECOND_PERSON],
-)
+ME = LexiconEntry("wo3", NOUN, intrinsic_morphosyntactic_properties=[FIRST_PERSON])
+YOU = LexiconEntry("ni3", NOUN, intrinsic_morphosyntactic_properties=[SECOND_PERSON])
 GAILA_PHASE_1_CHINESE_LEXICON = OntologyLexicon(
     ontology=GAILA_PHASE_1_ONTOLOGY,
     ontology_node_to_word=(
--- a/adam/language_specific/chinese/chinese_phase_1_lexicon.py
+++ b/adam/language_specific/chinese/chinese_phase_1_lexicon.py
@@ -1,7 +1,7 @@
-'''Outstanding questions:
+"""Outstanding questions:
 handling of mass nouns in Chinese?
 can we have spaces in our Chinese transcriptions?
-should we specify plurals even though they aren't morphologically salient?'''
+should we specify plurals even though they aren't morphologically salient?"""
 
 # import universal dependencies
 from adam.language.dependency.universal_dependencies import (
@@ -87,7 +87,7 @@ ALLOWS_DITRANSITIVE = LexiconProperty("allows-ditransitive")
 (v.s. Pinyin, Simplified Chinese, etc.)"""
 
 # first person nominative & accusative pronoun lexicon entry
-I = LexiconEntry(
+ME = LexiconEntry(
     "wo3",
     NOUN,
     plural_form="wo3 men",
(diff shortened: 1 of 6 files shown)


paynesa:
I moved universal properties to the module level.

adam/language_specific/chinese/chinese_phase_1_lexicon.py
```diff
@@ -0,0 +1,140 @@
...
+YOU = LexiconEntry(
+    "ni3",
+    NOUN,
+    plural_form="ni3 men",
+    intrinsic_morphosyntactic_properties=[SECOND_PERSON],
+)
+GAILA_PHASE_1_CHINESE_LEXICON = OntologyLexicon(
+    ontology=GAILA_PHASE_1_ONTOLOGY,
+    ontology_node_to_word=(
+        (GROUND, LexiconEntry("di4 myan4", NOUN)),
```
gabbard, line 91: @paynesa : Can you make an issue for having these checked over by a native speaker?

adam/language_specific/chinese/chinese_phase_1_lexicon.py
```diff
@@ -0,0 +1,140 @@
...
+ME = LexiconEntry(
+    "wo3",
+    NOUN,
+    plural_form="wo3 men",
+    intrinsic_morphosyntactic_properties=[FIRST_PERSON],
+)
+YOU = LexiconEntry(
+    "ni3",
+    NOUN,
+    plural_form="ni3 men",
```
gabbard, line 85: We should probably omit the plural forms of these, since I think they are realized syntactically and not morphologically.

adam/language_specific/chinese/chinese_phase_1_lexicon.py
```diff
@@ -0,0 +1,140 @@
...
+        (COOKIE, LexiconEntry("chyu1 chi2 bing3", NOUN)),
+        (DAD, LexiconEntry("ba4", PROPER_NOUN)),
+        (BABY, LexiconEntry("bau3 bau3", NOUN)),
+        (DOG, LexiconEntry("gou3", NOUN)),
+        (BIRD, LexiconEntry("nyau3", NOUN)),
+        (GO, LexiconEntry("dzou3", VERB)),
+        (COME, LexiconEntry("lái", VERB)),
+        (TAKE, LexiconEntry("na2", VERB)),
+        (EAT, LexiconEntry("chr1", VERB)),
+        (GIVE, LexiconEntry("gei3", VERB, properties=[ALLOWS_DITRANSITIVE])),
```
gabbard, line 120: Is the di-transitive allowed in Chinese?

adam/language_specific/chinese/chinese_phase_1_lexicon.py
```diff
@@ -0,0 +1,140 @@
+"""We use Yale romanization in this code to maintain UTF-8 encoding;
+spaces indicate borders between Chinese characters.
+
+Currently, mass nouns are specified despite this not being realised in
```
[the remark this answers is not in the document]
paynesa: http://phonetics.linguistics.ucla.edu/wpl/issues/wpl17/papers/26_liu.pdf

This is the most informative overview I was able to find. The literature isn't conclusive but it looks like mass/count is either realized by use of certain classifiers or it's not realized at all. A lot of authors argue that all nouns are mass by default and certain classifiers are needed to make them function as count nouns based on the number-classifier-noun relationship in Chinese. I kept this property for now for semantics, but it might turn out we don't need it if we don't deal too much with classifiers.

adam/language_specific/chinese/chinese_phase_1_lexicon.py
paynesa: From this: https://resources.allsetlearning.com/chinese/grammar/Verbs_that_take_double_objects
my understanding is yes.

adam/language_specific/chinese/chinese_phase_1_lexicon.py
paynesa: Issue #781

adam/language_specific/chinese/chinese_phase_1_lexicon.py
paynesa: Just pushed a new version to fix that

adam/language_specific/chinese/chinese_phase_1_lexicon.py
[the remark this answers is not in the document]
gabbard: Thanks - can you make an issue to deal with classifiers in Chinese, in general?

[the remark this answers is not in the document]
paynesa: Issue #782

PR #810, opened 2020-06-16
Hierarchical object recognition

### LINKED ISSUE #502: How to deal with hierarchical object recognitions without allowing overlapping object recognitions?
When getting the preposition subset learned to work with a less controlled curriculum, I ran into the problem where some of the object recognitions were overlapping, resulting in weird edges in the post-recognition graph like one "recognized object" node being marked as a component of another one (when neither object was actually a component of another).

In https://github.com/isi-vista/adam/pull/501, I modified the object recognizer to delete the nodes which make up a recognized object from the graph, which eliminates the problem above. However, it does create another problem: we can no longer recognize objects which truly are a part of another object, like hands being a part of people.

I think the correct solution here is do do all the matches without deletion, then do all the matched node insertions and deletions jointly, so that we can make explicit decisions about overlaps and create relationships appropriately for hierarchical objects.

spigo900:
Closes #502.

### DIFF
--- a/adam/learner/object_recognizer.py
+++ b/adam/learner/object_recognizer.py
@@ -560,7 +560,7 @@ def _get_edge_label(
 
 def _get_root_object_perception(
     graph: DiGraph, matched_subgraph_nodes: AbstractSet[PerceptionGraphNode]
-) -> PerceptionGraphNode:
+) -> ObjectPerception:
     matched_object_perceptions = immutableset(
         node for node in matched_subgraph_nodes if isinstance(node, ObjectPerception)
     )
--- a/adam/learner/object_recognizer.py
+++ b/adam/learner/object_recognizer.py
@@ -879,8 +879,12 @@ def replace_match_root_with_object_graph_node(
                     label=edge_label,
                 )
 
-    if root not in SHARED_WORLD_ITEMS:
-        perception_digraph.remove_node(root)
+    if root in SHARED_WORLD_ITEMS:
+        raise RuntimeError(
+            f"Pattern match root cannot be a shared world item, "
+            f"but got match {pattern_match}"
+        )
+
     perception_digraph.remove_node(root)
     perception_digraph.remove_nodes_from(duplicate_nodes_to_remove)
 
--- a/adam/learner/object_recognizer.py
+++ b/adam/learner/object_recognizer.py
@@ -637,7 +637,7 @@ def replace_match_with_object_graph_node(
 ) -> PerceptionGraph:
     """
     Internal function to replace the nodes of the perception matched by the object pattern
-    with an `ObjectSemantcNode`.
+    with an `ObjectSemanticNode`.
 
     Any external relationships those nodes had is inherited by the `ObjectSemanticNode`.
     """
@@ -772,7 +772,7 @@ def replace_match_root_with_object_semantic_node(
 ) -> PerceptionGraph:
     """
     Internal function to replace the root node of the perception matched by the object pattern
-    with an `ObjectSemantcNode`.
+    with an `ObjectSemanticNode`.
 
     The `ObjectSemanticNode` inherits both the root node's internal relationships and all external
     relationships involving either the root node or its children.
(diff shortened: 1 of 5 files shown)


adam/learner/object_recognizer.py
```diff
@@ -725,7 +754,142 @@ def replace_match_with_object_graph_node(
...
+        original_graph=current_perception.copy_as_digraph(),
+        output_graph=perception_digraph,
+        matched_nodes=matched_subgraph_nodes,
+        matched_object_node=matched_object_node,
+    )
+
+    return PerceptionGraph(perception_digraph, dynamic=current_perception.dynamic)
+
+
+def replace_match_root_with_object_graph_node(
```
spigo900, line 767: I think Ryan had a (possibly better) name for this function but I can't recall what it was.

adam/learner/object_recognizer.py
```diff
@@ -725,7 +754,142 @@ def replace_match_with_object_graph_node(
...
+                        external_properties.add(matched_subgraph_node_predecessor[0])
+
+                perception_digraph.add_edge(
+                    matched_subgraph_node_predecessor,
+                    matched_object_node,
+                    label=edge_label,
+                )
+
+    if root not in SHARED_WORLD_ITEMS:
+        perception_digraph.remove_node(root)
```
spigo900, line 883: Is this the behavior we want? I assumed so based on the original function's behavior.

spigo900:
Note this doesn't change the object recognizer, which I think is as intended.

adam/learner/object_recognizer.py
```diff
@@ -725,7 +754,142 @@ def replace_match_with_object_graph_node(
...
+    perception_digraph.remove_node(root)
+    perception_digraph.remove_nodes_from(duplicate_nodes_to_remove)
+
+    # We want to re-add any relationships linked directly to the root node of an object.
```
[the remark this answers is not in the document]
spigo900: @gabbard Right, we preserve any internal relationships between the sub-objects by not deleting the entire sub-graph. But here we need to add those that involve the root itself because we replaced the original root with a new root.

The external relationships should be handled by the `matched_subgraph_node` loop?

adam/learner/object_recognizer.py
```diff
@@ -725,7 +754,142 @@ def replace_match_with_object_graph_node(
...
+                perception_digraph.add_edge(
+                    matched_subgraph_node_predecessor,
+                    matched_object_node,
+                    label=edge_label,
+                )
+
+    if root not in SHARED_WORLD_ITEMS:
+        perception_digraph.remove_node(root)
+    perception_digraph.remove_node(root)
+    perception_digraph.remove_nodes_from(duplicate_nodes_to_remove)
```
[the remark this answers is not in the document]
spigo900: @gabbard Ah, I see.

Follow-up -- are there any constraints on external relationships the root should inherit? When `Mom` gets replaced in my test situation, I'm getting a relationship edge that looks like `(ObjectSemanticNode(Mom), the ground, reference-object)` (in order, `(from_node, to_node, label)`). This relationship doesn't appear in the original graph. It seems to be inherited from one of the regions. Is this intended behavior?

adam/learner/object_recognizer.py
spigo900: Yes, that makes sense. I made the change.

tests/learner/object_learner_test.py
```diff
@@ -97,6 +126,107 @@ def test_subset_learner_dog():
...
+        PerceptionSemanticAlignment.create_unaligned(perception_graph)
+    )
+
+    semantic_node_types_and_debug_strings = {
+        (type(semantic_node), semantic_node.concept.debug_string)
+        for semantic_node in enriched.semantic_nodes
+    }
+    assert (ObjectSemanticNode, "Mom") in semantic_node_types_and_debug_strings
+    assert (ObjectSemanticNode, "head") in semantic_node_types_and_debug_strings
+    assert (ObjectSemanticNode, "hand") in semantic_node_types_and_debug_strings
```
spigo900, line 227: Is there a better way to write these assertions?

adam/learner/object_recognizer.py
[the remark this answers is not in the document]
gabbard: Got it.

Issue #55, opened 2019-08-28
Better repr for HighLevelSemanticsSituation


rgabbard:
The way we currently `repr` `HighLevelSemanticsSituation`s doesn't distinguish multiple nodes of the same ontology type.  We'd like to use subscripts (`_0`, `_1`) to distinguish them, which requires pulling all the `repr` logic up to this level and not delegating to the `repr`s of the sub-objects.

chris-jenkins-isi:
This could be useful for conditional positioning of objects -- being able to say *this* particular house is the goal object for this action, so it should stay in place. #621

gabbard:
@chris-jenkins-isi - how is that related to the `repr`? We really don't want logic depending on `repr`s...

chris-jenkins-isi:
Related in the sense that it's at the level of the `HighLevelSemanticsSituation` to assign unique handles (`house_0`, `house_1`) to the objects in the situation.

gabbard:
@chris-jenkins-isi : Can't we distinguish them by object-identity?

PR #812, opened 2020-06-16
Chinese classifiers implementation

### LINKED ISSUE #782: Handle Chinese Classifiers
This has implications for count/mass noun distinction.

### LINKED ISSUE #790: Handle Chinese Obliques
According to https://wals.info/chapter/84, Chinese generally has an XVO construction for obliques. However, there are several caveats. Namely:

1. Non-ditransitive verbs in Chinese don’t allow 2 objects to occur after the verb, so one is moved to the preverbal position as an oblique, preceded by “ba.” For example, “I give you a present” but “you (ba)book put at-table-on” There is a obl:patient universal dependency tag we could use for this construction, but I’m not sure if we necessarily need to given that this is still in XVO construction. 
 
2. Adverbs typically occur between X and V, except in the case of negation in “ba” clauses in which the adverb must come before “ba"-- this is where the "ba" clauses become more of a problem (https://resources.allsetlearning.com/chinese/grammar/Using_%22ba%22_sentence). It’s valid to place adverbs before all “ba” clauses with slightly different meaning than after, so I’m not sure how to handle this without either having two adverb slots or using the obl:patient tag (or ignoring negation?). 
 
3. Duration or frequency obliques can occur after a verb as long as there’s not also an object (e.g. “I ran 3 days” = “I ran for 3 days”). I’m really struggling on how to handle this. We could, for instance, use the obl:temp modifier, but I’m not sure how much this would help vs. overcomplicate things. 

Right now, I've put the obl in the XVO position, but I think we still need to handle the other cases above (and probably have a native speaker check them). This solution necessarily takes out a lot of the noise in the data re oblique locations, which could be a good thing for our learner but is also arguable a hack.

### LINKED ISSUE #795: Handle localizers vs. prepositions in Chinese
Chinese indicates location (e.g. in the case of "put the book on the table") with localizers which attach at the end of NP's rather than the beginning like prepositions (i.e. "the book put the table on"). If we want to indicate prepositions that can't be indicated with localizers, we'll need two separate ways to indicate this.

### LINKED ISSUE #796: Guaranteed goal and Dao vs. Zai distinction
In Chinese, I believe localiser phrases (similar to English PP's) depend on whether the verb guarantees an end location or not. If there is no end location guaranteed, we use "dao," but if there is a guaranteed one, we use "zai." For example: 

push, take (to), move (to), jump (onto), fly, throw all take "dao"

put, sit, fall (onto) all take "zai"

We need a way to distinguish between these two for the Chinese language generator. There's already an open issue (#582) about roll, take, fall, fly, jump not taking GOALS, and all of these are also "dao" verbs, which means that we could potentially use this as a distinction. However, from a user interface POV, this might not be ideal for reasons discussed in #582.

Other options include the in-region distinction.

### LINKED ISSUE #805: Chinese Localiser Phrase native speaker check
Once our localiser phrases are up and running they will need to be checked by a native speaker since there are subtle localiser differences in Chinese that do not exist in English.

paynesa:
Updates to `LexiconEntry` class and Chinese lexicon and generator to be able to count nouns using classifiers.

Closes #805 
Closes #782 
Closes #790 
Closes #795 
Closes #796 (Followup for this issue includes how I handled it for various cases)


adam/language_specific/chinese/chinese_language_generator.py
```diff
@@ -266,24 +646,57 @@ def add_classifier(
...
+                    possessor = DependencyTreeToken("wo3", NOUN)
+                elif IS_ADDRESSEE in possession_relations[0].first_slot.properties:
+                    possessor = DependencyTreeToken("ni3", NOUN)
+                # if the possessor is a 3rd person, check that "has" isn't the main verb
+                elif (
+                    IGNORE_HAS_AS_VERB not in self.situation.syntax_hints
+                    and not self.situation.is_dynamic
+                ):
+                    return
+                    # TODO: we currently return here since we can't handle one possessive node and one not -- once we fix this we don't need this case
```
LichMaster98, line 694: Should this perhaps error out rather than silently return? If in the future we go to implement needing this feature we get an error rather than a silent failure. Silent failures are significantly hard to debug. We should also track this in an issue.

It's also possible I'm not understanding the case in which we simply return here, in that case the comment needs to just be clearer.

adam/language_specific/chinese/chinese_syntax.py
```diff
@@ -34,7 +34,7 @@
 ] = [
     # Currently we treat locations associated with verbs as IO's since obliques occur before the verb.
     # https://github.com/isi-vista/adam/issues/797
-    # TODO: find a better way to handle the above
+    # TODO: find a better way to handle the aboveA
```
LichMaster98, line 37:
```suggestion
    # TODO: find a better way to handle the above
```

tests/language_specific/chinese/test_chinese_language_generator.py
```diff
@@ -1702,8 +2221,72 @@ def test_roll():
...
+        always_relations=[Relation(HAS, dad, ball)],
+        syntax_hints=[IGNORE_HAS_AS_VERB, ATTRIBUTES_AS_X_IS_Y],
+    )
+    assert generated_tokens(situation) == ("ni3", "de", "chyou2", "shr4", "hung2 se4")
+
+
+# x_is_y for classifier sentence
+def test_two_balls_are_green():
+    ball1 = situation_object(BALL, properties=[GREEN], debug_handle="ball1")
+    ball2 = situation_object(BALL, properties=[GREEN], debug_handle="ball2")
```
LichMaster98, line 2273: Is there a reason you need to specify the debug handle here? My guess is because the default is simply "ball"

tests/language_specific/chinese/test_chinese_language_generator.py
```diff
@@ -1702,8 +2221,72 @@ def test_roll():
...
+        syntax_hints=[ATTRIBUTES_AS_X_IS_Y],
+    )
+    assert generated_tokens(situation) == ("lyang3", "ge4", "chyou2", "shr4", "lyu4 se4")
+
+
+# x_is_y for a sentence with many (> 2) objects
+def test_many_balls_are_red():
+    ball1 = situation_object(BALL, properties=[RED], debug_handle="ball1")
+    ball2 = situation_object(BALL, properties=[RED], debug_handle="ball2")
+    ball3 = situation_object(BALL, properties=[RED], debug_handle="ball3")
```
LichMaster98, line 2286: Also here with the debug handles.

tests/language_specific/chinese/test_chinese_language_generator.py
paynesa: Yes, it's so if something goes wrong they can be distinguished to determine where the error was. I can take them out now they're passing if you think that's a good idea.

LichMaster98: No, if that's all it is it's fine to leave in. Helpful if someone accidentally breaks the tests in the future. I was just under the impression we had tried to automatically handle distinguishing between multiple of the same ontology node, but I may be mistaken for this implementation.

adam/language_specific/chinese/chinese_language_generator.py
paynesa: @LichMaster98 I updated the comment so it should be clearer now; this is in reference to issue #55 which also affects the English generator and this is how it's handled in the English one. I tend to think "Dad has a cookie" is better than just throwing an error in these cases but I can change it if you prefer.

LichMaster98: Ok! If it's handled this way in the English language generator that's fine then.

PR #818, opened 2020-06-19
Translate subobject properties

### LINKED ISSUE #814: Perception Generation fails to translate properties of sub-objects
For example a `person` contains the sub-objects of `head, arm, leg, torso` each of these has independent sub-properties as well. `head` for example includes `HOLLOW` 

The lack of translation causes the matcher to be unable to recognize such sub-components in the larger structure.

spigo900:
Closes #814.


spigo900:
Note: A side effect of this is that things like `the ground` also have their properties and properties of their subobjects added, not just situation objects. I think at the moment this only really affects `the ground`.

spigo900:
Ready to merge.

PR #819, opened 2020-06-19
Update learner tests and learner for Chinese

paynesa:
closes https://github.com/isi-vista/adam/issues/834


adam/learner/relations.py
```diff
@@ -79,17 +79,48 @@ def is_legal_template_span(candidate_relation_token_span: Span) -> bool:
...
+                        if span_for_right_object.end < len(sentence_tokens) and Span(
+                            span_for_right_object.end, span_for_right_object.end + 1
+                        ) not in list(
+                            language_concept_alignment.node_to_language_span.values()
+                        ):
+                            rightone = sentence_tokens[
+                                span_for_right_object.end : span_for_right_object.end + 1
+                            ]
+                            if rightone[0] == "a" or rightone[0] == "the":
+                                rightone = None
```
LichMaster98, line 110: @gabbard Should this perhaps function closer to how the verb learner generates multiple different options for the verb phrase? So a version with `left_start_between_end`, `start_between_end`, and `start_between_end_right` ? Giving us up to 3 possible hypotheses for the semantics of the relationship?

LichMaster98: @paynesa Can you add an issue to track this hack? We'll probably want to adapt the solution I mention above in the future so it's more general.

paynesa: https://github.com/isi-vista/adam/issues/849

PR #840, opened 2020-06-24
Fix bad hypothesis intersections for verb subset learner

### LINKED ISSUE #822: Bad verb subset learner hypothesis intersections
As discussed on Slack, we are having an issue where the subset verb learner is learning some strange patterns for a few verbs (`put` and `push`). The subset learner code is intersecting apparently correct hypothesis graphs and getting incorrect results.

From discussion with @LichMaster98 , we think that the problem with the hypothesis pattern intersections being weird is that the first iteration is matching slot 1 in the first pattern with an object that's not slot 1 in the second pattern and then iterating on that.

The solution we discussed was as follows:
  - Add a new keyword argument to the pattern intersection method that gives a mapping from nodes in pattern 1 to sets of nodes in pattern 2 they are allowed to match with.
      - We can then use this when intersecting hypotheses to specify that (for all k) slot k in the first pattern must match with slot k in the second pattern.
  - Push this down the call stack into the graph matching algorithm so that the graph matching algorithm will actually implement this functionality.

spigo900:
Closes #822.

### DIFF
--- a/adam/learner/subset.py
+++ b/adam/learner/subset.py
@@ -20,7 +20,12 @@ from adam.perception.perception_graph import DebugCallableType
 from adam.semantics import Concept
 from attr import Factory, attrib, attrs
 from attr.validators import instance_of
-from immutablecollections import ImmutableSet, immutabledict, immutableset
+from immutablecollections import (
+    ImmutableSet,
+    immutabledict,
+    immutableset,
+    immutablesetmultidict,
+)
 
 
 @attrs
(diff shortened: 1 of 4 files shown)


adam/learner/subset.py
```diff
@@ -44,10 +49,23 @@ def _learning_step(
...
+            raw_new_pattern_hypothesis = self._hypothesis_from_perception(
+                preprocessed_input
+            )
 
             updated_hypothesis = previous_pattern_hypothesis.intersection(
-                self._hypothesis_from_perception(preprocessed_input),
+                raw_new_pattern_hypothesis,
                 ontology=self._ontology,
+                # Q: Is there a better way to compute this? Is it guaranteed that the slots are
+                # in order so that we can do zip(prev, new)?
```
LichMaster98, line 60: I don't think it's strictly guaranteed (as in by a `attrs_post_init` or something similar) that the slots are in order however they should be because we only pull the slots in order.

spigo900: Thanks. In that case, I think it makes sense to keep it like this. I'll remove the comment.

spigo900:
Ready to merge.

PR #854, opened 2020-06-30
Fix subset verb learner tests

### LINKED ISSUE #853: Subset verb learner tests use training example for testing
The subset verb learner tests consistently use the first training example for testing. Ideally we want to test on an example not in the training set.

The reason it is using the first training example is that `run_verb_test()` uses the same chooser factory for constructing/sampling the training and test instances. This chooser factory uses a fixed seed (the seed 0).

To fix this we'll want to use different factory with a different seed for constructing the test curriculum.

spigo900:
Closes #853.

### DIFF
--- a/adam/curriculum/curriculum_utils.py
+++ b/adam/curriculum/curriculum_utils.py
@@ -34,6 +34,7 @@ from adam.situation.templates.phase1_templates import (
 
 GROUND_OBJECT_TEMPLATE = object_variable("ground", GROUND)
 PHASE1_CHOOSER_FACTORY = lambda: RandomChooser.for_seed(0)  # noqa: E731
+PHASE1_TEST_CHOOSER_FACTORY = lambda: RandomChooser.for_seed(1)  # noqa: E731
 Phase1InstanceGroup = InstanceGroup[  # pylint:disable=invalid-name
     HighLevelSemanticsSituation,
     LinearizedDependencyTree,
--- a/adam/learner/object_recognizer.py
+++ b/adam/learner/object_recognizer.py
@@ -30,6 +30,8 @@ from adam.ontology.phase1_ontology import (
     GAILA_PHASE_1_ONTOLOGY,
     PART_OF,
     PHASE_1_CURRICULUM_OBJECTS,
+    BIGGER_THAN,
+    SMALLER_THAN,
 )
 from adam.perception import GROUND_PERCEPTION, LEARNER_PERCEPTION, ObjectPerception
 from adam.perception.deprecated import LanguageAlignedPerception
@@ -731,7 +733,13 @@ def replace_match_with_object_graph_node(
                 if edge_equals_ignoring_temporal_scope(edge_label, HAS_PROPERTY_LABEL):
                     # Prevent multiple `has-property` assertions to the same color node
                     # On a recognized object
-                    if matched_subgraph_node_successor[0] in external_properties:
+                    # Also prevent size relations from being inherited on root object.
+                    if matched_subgraph_node_successor[
+                        0
+                    ] in external_properties or matched_subgraph_node_successor[0] in {
+                        SMALLER_THAN,
+                        BIGGER_THAN,
+                    }:
                         if (
                             perception_digraph.degree(matched_subgraph_node_successor)
                             != 1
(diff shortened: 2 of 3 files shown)


spigo900:
Testing this locally, the Chinese subset verb learner tests seem to be failing. The failing Chinese tests don't seem to be failing for the integrated learner. I suspect this may have something to do with how the integrated learner is created (a new one for each test) vs. how the old-style subset learner is created (one for all tests at the top of the file). However I'm not sure as this pattern doesn't happen for one of the tests. (The test is `test_put`, where both the integrated and old-style learner fail only for the Chinese version of the test).

Another problem is that both the integrated and subset learner are failing for a few verbs in both English and Chinese variants. The verbs are

- take
- spin
- fall
- throw
- put

spigo900:
Re: the first issue, one of the failing cases is eat_simple. Here are the linguistic descriptions I
had it print:
```
TRAINING: linguistic description "gou3 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "ma1 ma1 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "nyau3 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "gou3 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "gou3 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "nyau3 chr1 syi1 gwa1"
TRAINING: linguistic description "nyau3 chr1 syi1 gwa1"
TRAINING: linguistic description "bau3 bau3 chr1 syi1 gwa1"
TRAINING: linguistic description "ma1 ma1 chr1 chyu1 chi2 bing3"
TRAINING: linguistic description "nyau3 chr1 syi1 gwa1"
TEST: linguistic description "ba4 ba4 chr1 syi1 gwa1"
```

The underlying situations should be the same between the two since they're using
the same chooser factories.

The specific error in this case is that the learner doesn't produce any
descriptions of the situation at all.

From running the old-style learner's test in the debugger, that learner seems to be learning a different hypothesis for each of the 5 linguistically distinct situations it sees in training. The integrated learner is at least learning a two-slot template for `chr1`. I don't know why the old-style learner is doing this but not the new-style.

spigo900:
`test_fall` fails for both old and new learners, both English and Chinese. In these cases the integrated learner generates descriptions of objects with no action descriptions, while the subset learner generates no descriptions at all. The English gold descriptions:

```
TRAINING: linguistic description "a cookie falls"
TRAINING: linguistic description "a ball falls"
TRAINING: linguistic description "a dog falls"
TRAINING: linguistic description "a cookie falls"
TRAINING: linguistic description "a baby falls"
TRAINING: linguistic description "a cup falls"
TRAINING: linguistic description "a truck falls"
TRAINING: linguistic description "a book falls"
TRAINING: linguistic description "a bird falls"
TRAINING: linguistic description "a truck falls"
TEST: linguistic description "a house falls"
```

The integrated subset learner at least is learning an erroneous description. It somehow is learning that `slot` must be smaller than the learner. This seems to be the cause of the problem for this test case. (It's also learning that the ground is about the same size as the learner, which seems wrong, but unrelated.)

(The subset learner [should fail anyway](https://usc-isi.slack.com/archives/GL17LRSTG/p1593549895398900) since the linguistic description "a house falls" is not one of the training examples.)

spigo900:
After more testing, the problem in the [above case](https://github.com/isi-vista/adam/pull/854#issuecomment-652047318) seems to be that the object recognizer is inheriting learner-relative size properties from subobjects up to their parent objects. Both `truck` and `dog` have subobjects that are smaller than `baby`, so they end up having the `smallerThan` property and this is never removed from `slot1` in the hypothesis. `house` meanwhile does not have any small subobjects so it is unable to match as `slot1`.

spigo900:
That was indeed the problem. Resolving it seems to have fixed the English versions of the tests that were failing on both learners. (That is, take, spin, fall, throw, and put.)

spigo900:
The problem is *not* just that the Chinese tests are reusing the subset learner. If I change the tests to create a new instance every time the learner simply doesn't generate any descriptions of the situations.

Also, the integrated learner is having trouble with the Chinese test for `put`.

PR #857, opened 2020-07-01
Chinese HTML curriculum dump

### LINKED ISSUE #855: Chinese Curriculum Dump
We'll want to parameterize the curriculum_dump code so that we have a language parameter that we can use to specify which language to generate a dump for

### LINKED ISSUE #856: Chinese big/small distinction currently only applies to NP
Since we explicitly translate always relations separately for NPs and VPs in the Chinese language generator, we need a separate place to translate these.

paynesa:
closes #856 
closes #855 
Issues discovered while going through curricula: #860, #859, #858

### DIFF
--- a/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
+++ b/adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
@@ -433,6 +433,7 @@ def _go_beside_template(
                 ],
             )
         ],
+        after_action_relations=[near(agent, goal_object)],
     )
 
 
@@ -467,6 +468,7 @@ def _go_behind_in_front_template(
                 ],
             )
         ],
+        after_action_relations=[near(agent, goal_object)],
     )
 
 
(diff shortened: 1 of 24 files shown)


adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
```diff
@@ -945,12 +956,14 @@ def _x_throws_y_to_z_template(
...
-                    (GOAL_MANIPULATOR, recipient),
+                    # hack since English language generator doesn't generate correct language for GOAL_MANIPULATOR currently
+                    (GOAL, Region(recipient, distance=PROXIMAL)),
                 ],
             )
         ],
+        after_action_relations=[near(theme, recipient)],
         constraining_relations=flatten_relations(bigger_than([agent, recipient], theme)),
-        gazed_objects=[theme],
+        gazed_objects=[agent, theme, recipient],
```
LichMaster98, line 966: I realize this param is called `gazed_objects` but in the normal case we only want to be gazing at one object. We're set up to handle more than one to see if input noise on the `gaze` attribute causes learning issues.

Issue #498, opened 2019-12-18
Remove hard-coded ignoring of determiners from object learner


gabbard:

denizbeser:
Still relevant post-M6. Ties to integration of modifier learning #474 and preposition learning #475

gabbard:
Places which will need fixing: 
- [ ] `ObjectRecognizer`

LichMaster98:
An additional place which needs fixing:
[ ] `covers_entire_utterance` in `learner_utils` module

gabbard:
We also have special-cased English determiner information in `_handle_quantifiers` in `IntegratedLearner`.

PR #868, opened 2020-07-07
Verify that all templates need to learn from the entire utterance.

### LINKED ISSUE #863: Integrated learner should only try to learn an utterance if it can fully understand the example
Per a discussion on July 1st with @charles-yang-upenn and @mitchmarcus we want to have our learner only try to learn new examples when it fully understands what is has been presented.

For example from a situation "Mom eats a cookie" we won't learn "eats cookie" or "mom eats" because that wouldn't be a complete match of the situation text. 

By limiting our learning to only these 'complete examples', which we believe kids do as well, we constrain our utterance space considerably and don't have to deal with the question of how to eventually filter out the "eats cookie" or "mom eats" examples which would always match the longer situation text of "mom eats a cookie".

LichMaster98:
Closes #863

### DIFF
--- a/adam/learner/attributes.py
+++ b/adam/learner/attributes.py
@@ -107,6 +107,8 @@ class AbstractAttributeTemplateLearnerNew(AbstractTemplateLearnerNew, ABC):
             if covers_entire_utterance(
                 bound_surface_template,
                 language_concept_alignment,
+                # We need to explicitly ignore determiners here for some reason
+                # See: https://github.com/isi-vista/adam/issues/871
                 ignore_determiners=True,
             )
         )
--- a/adam/learner/attributes.py
+++ b/adam/learner/attributes.py
@@ -9,6 +9,7 @@ from adam.learner.alignments import (
 from adam.learner.learner_utils import (
     assert_static_situation,
     pattern_remove_incomplete_region_or_spatial_path,
+    covers_entire_utterance,
 )
 from adam.learner.object_recognizer import (
     ObjectRecognizer,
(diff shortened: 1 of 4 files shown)


adam/learner/attributes.py
```diff
@@ -98,7 +99,25 @@ def _candidate_templates(
...
+            # We strictly ignore english determiners here for determining
+            # the length needed of our surface template
+            if covers_entire_utterance(
+                bound_surface_template,
+                language_concept_alignment,
+                token_sequence_count=len(
+                    [
+                        token
+                        for token in language_concept_alignment.language.as_token_sequence()
+                        if token not in ["a", "the"]
```
LichMaster98, line 116: @Reviewer - Any ideas why the attribute learner needs this specific condition to be able to learn color attributes while the relationship and verb learner can handle the determiners?

LichMaster98: This was found as a result of the attribute SurfaceTemplate not appropriately adding the SLOT variables as having determiners.

adam/learner/learner_utils.py
```diff
@@ -191,3 +195,37 @@ def sort_by_num_nodes(g: DiGraph) -> int:
...
+    for element in bound_surface_template.surface_template.elements:
+        if isinstance(element, str):
+            num_covered_tokens += 1
+        else:
+            num_covered_tokens += len(
+                language_concept_alignment.node_to_language_span[
+                    bound_surface_template.slot_to_semantic_node[element]
+                ]
+            )
+    # We may need to
```
gabbard, line 216: I think this comment is missing something

adam/learner/learner_utils.py
```diff
@@ -191,3 +195,37 @@ def sort_by_num_nodes(g: DiGraph) -> int:
...
+    sized_tokens = (
+        len(language_concept_alignment.language.as_token_sequence())
+        if not ignore_determiners
+        else len(
+            [
+                token
+                for token in language_concept_alignment.language.as_token_sequence()
+                if token not in ["a", "the"]
```
gabbard, line 224: Do we have a "constant" for the determiner set somewhere?  I think we also somewhere track English-specific places in the code (e.g. on an issue). If so, add this there.

LichMaster98: Added: https://github.com/isi-vista/adam/issues/498#issuecomment-654942842

PR #873, opened 2020-07-07
FixedStrings no longer include recognized concepts in them

### LINKED ISSUE #867: Integrated Learner for Verbs needs to fully respect slot variables
After reviewing a complete run of the Integrated Learner with the ObjectRecognizerAsObjectLearner the verb learner produces verb patterns with the name of "slot1_passes_a_watermelon_to_slot2" even though it has a pattern for "slot1_passes_slot2_to_slot3" 

Pattern validity should rule the middle case invalid. Perhaps if we should require a verb template to match the number of matched objects in the scene?

### LINKED ISSUE #875: Chinese Integrated Learner is unable to log pattern hypotheses
Strangely the chinese learner is unable to log pattern hypotheses

LichMaster98:
Closes #867 Closes #875

### DIFF
--- a/adam/learner/learner_utils.py
+++ b/adam/learner/learner_utils.py
@@ -26,6 +26,7 @@ from enum import Enum, auto
 from adam.language import LinguisticDescription, TokenSequenceLinguisticDescription
 from adam.learner import LearningExample
 from adam.learner.alignments import LanguageConceptAlignment
+from adam.learner.language_mode import LanguageMode
 from adam.learner.perception_graph_template import PerceptionGraphTemplate
 from adam.perception import PerceptualRepresentation
 from adam.perception.developmental_primitive_perception import (
@@ -56,6 +57,7 @@ from adam.semantics import (
 from immutablecollections import immutabledict, immutableset, ImmutableSet
 
 from adam.utils.networkx_utils import subgraph
+from vistautils.span import Span
 
 
 def pattern_match_to_description(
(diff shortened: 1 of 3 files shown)


adam/learner/learner_utils.py
```diff
@@ -231,3 +231,14 @@ def covers_entire_utterance(
...
     # which is true for how we construct them.
     return num_covered_tokens == sized_tokens
+
+
+def no_object_alignment_in_fixed_strings(
+    bound_surface_template: SurfaceTemplateBoundToSemanticNodes,
+    language_concept_alignment: LanguageConceptAlignment,
+) -> bool:
+    for (node, _) in language_concept_alignment.node_to_language_span.items():
+        if node.concept.debug_string in bound_surface_template.surface_template.elements:
```
gabbard, line 241: Will this approach fail for Chinese? Would it be better to examine the aligned token positions directly?

LichMaster98: In our current unit tests, this doesn't fail for Chinese. I don't believe it should fail because the `debug_string` of a concept is derived from the language which was originally aligned with the concept. -- However, I should be able to run a more thorough test by running the experiment pipeline over the learner in Chinese. I'll do so and report back.

gabbard: Got it.  In any case, though, is the information available here to do the task by comparing token indices?  That is less brittle, since we really want to minimize the amount we rely on `debug_string` (since it is, after all, for debugging and not for logic)

LichMaster98: Hm, I think I may be able to do so given we assume that the entire utterance is covered before checking this condition. Without that assumption, there could be potential token offsets to account for.

adam/learner/learner_utils.py
```diff
@@ -251,6 +251,47 @@ def covers_entire_utterance(
...
+    # over the given span are at all strings, if so we've included an already aligned
+    # SemanticNode in our fixed string structures so we fail out
+    for formated_span in formated_spans:
+        for token in bound_surface_template.surface_template.elements[
+            formated_span.start : formated_span.end
+        ]:
+            if isinstance(token, str):
+                return False
+
+    return True
```
LichMaster98, line 292: @gabbard This version doesn't rely on debug strings, but I'm now special case handling English determiners. Is there a better way to handle this casing?

adam/learner/learner_utils.py
```diff
@@ -303,23 +303,27 @@ def in_left_to_right_order(semantic_nodes: Tuple[SemanticNodeWithSpan, ...]) ->
             previous_node = semantic_nodes[i]
         return True
 
-    def is_legal_template_span(candidate_verb_token_span: Span) -> bool:
+    def is_legal_template_span(
+        candidate_token_span: Span, *, invalid_token_spans: ImmutableSet[Span]
```
gabbard, line 307: Can you add a comment giving a sample user case for `invalid_token_spans` (that it, the one motivated by the current issue)

adam/learner/learner_utils.py
```diff
@@ -503,17 +520,25 @@ def process_aligned_objects_with_template(
...
             # Which the spans go from left most to right most
+            # We also provide a set of invalid token spans
+            # for fixed string positions
             for (
                 surface_template_bound_to_semantic_nodes
-            ) in process_aligned_objects_with_template(verb_template, aligned_nodes):
+            ) in process_aligned_objects_with_template(
+                candidate_template,
+                aligned_nodes,
+                invalid_token_spans=immutableset(
```
gabbard, line 538: Might as well lift the creation of this set out of the loops.

adam/learner/learner_utils.py
```diff
@@ -503,17 +520,25 @@ def process_aligned_objects_with_template(
...
         for aligned_nodes in aligned_object_nodes(
-            sum(1 for token in verb_template if token == AlignmentSlots.Argument),
+            sum(1 for token in candidate_template if token == AlignmentSlots.Argument),
             num_arguments_to_alignments_sets,
             language_concept_alignment,
         ):
             # aligned_object_nodes is guaranteed to only give us alignments
             # Which the spans go from left most to right most
+            # We also provide a set of invalid token spans
+            # for fixed string positions
```
gabbard, line 532: Link to the relevant issue for the motivation.

PR #877, opened 2020-07-07
try + catch with warning for complex objects with colours

paynesa:


adam/learner/attributes.py
```diff
@@ -69,36 +69,46 @@ def _candidate_templates(
...
-                                language_mode=self._language_mode,
-                            ),
-                            {SLOT1: object_node},
+                                {SLOT1: object_node},
+                            )
+                        )
+                    except RuntimeError as e:
+                        raise RuntimeError(
+                            f"{str(e)} with preceding token index = {preceding_token_index}"
                         )
```
LichMaster98, line 88: Rather than continue to pass this error up the chain we should just do the logging here.

adam/learner/attributes.py
```diff
@@ -69,36 +69,46 @@ def _candidate_templates(
...
-                                language_mode=self._language_mode,
-                            ),
-                            {SLOT1: object_node},
+                                {SLOT1: object_node},
+                            )
+                        )
+                    except RuntimeError as e:
+                        raise RuntimeError(
+                            f"{str(e)} with following token index = {following_token_index}"
                         )
```
LichMaster98, line 111: Rather than continue to pass this error up the chain we should just do the logging here.

PR #878, opened 2020-07-08
Use monomorphism mode for perception graph pattern matching

### LINKED ISSUE #400: Graph matching only handles node-induced sub-graph isomorphisms
We identify objects by matching "perceptual patterns" represented as graphs against a graph representation of the learner's perception using the VF2 sub-graph isomorphism algorithm.  The implementation we borrowed from NetworkX and altered only handles *node-induced* isomorphism.  In such isomorphisms, we align a set of nodes between the pattern and the target graph and require that all edges between the nodes in the pattern must have aligned edges in the target and vice-versa.

This can be a problem if additional edges are present in the target graph.  For example, imagine that a concrete perception has more relationships between a target's axes than are specified in the pattern; we would still want this to match, but the extra edges would block a node-induced isomorphism.

We could switch to trying an *edge-induced* isomorphism algorithm, which aligns edges rather than nodes.  Or we could make some sort of modification to our current algorithm to allow extra edges on the graph side.

I don't *think* this is an immediate problem, but it may become a problem as we enrich the perceptual representation with "noise" features.

### LINKED ISSUE #673: Using node-induced graph isomorphism causes object recognition to fail when an action has preconditions which relate and object to a part of itself
We have known for some time that using a node-induced rather than edge-induced graph isomorphism algorithm was a problem (https://github.com/isi-vista/adam/issues/400), but it hadn't really bit us too much.

However, it starts to be a bigger problem for verbs. Consider a situation with _give_: _Dad gives a ball to a baby_.  Our object recognizer fails to recognize _Dad_ (or for that matter, _baby_).  This puzzled me at first, but I believe the cause is that the precondition `partOf(_GIVE_AGENT_MANIPULATOR, _GIVE_AGENT)` (and a similar post-condition for `baby`) introduce a `part-of` edge between the object root node and one of the `hand` nodes where no edge at all is ordinarily present.  If we were using edge-induced isomorphism, adding extra edges between aligned nodes is fine, but because we are using node-induced, this blocks alignment.  

Long-term, we need to fix our isomorphism algorithm. As a temporary stopgap, I have disabled the constraint that there be a direct `partOf` relationship between an agent and the manipulator.  This should be re-enabled after #400 is fixed.

spigo900:
Closes #400.
Closes #673.

### DIFF
--- a/adam/perception/_matcher.py
+++ b/adam/perception/_matcher.py
@@ -531,34 +531,6 @@ class GraphMatching:
                     return False
         return True
 
-    def subgraph_is_isomorphic(self):
-        """Returns True if a subgraph of G1 is isomorphic to G2."""
-        try:
-            next(self.subgraph_isomorphisms_iter())
-            return True
-        except StopIteration:
-            return False
-
-    def subgraph_isomorphisms_iter(
-        self,
-        *,
-        collect_debug_statistics: bool = False,
-        debug_callback: Optional[Callable[[Any, Any], None]] = None,
-        initial_partial_match: Mapping[Any, Any] = immutabledict(),
-    ):
-        """Generator over isomorphisms between a subgraph of G1 and G2."""
-        # Declare that we are looking for graph-subgraph isomorphism.
-        self.test = "subgraph"
-        self.initialize(initial_partial_match=initial_partial_match)
-        self.debug_largest_match = {}
-        self.failing_pattern_node_for_deepest_match = None
-        self._reset_debugging_maps()
-        for mapping in self.match(
-            collect_debug_statistics=collect_debug_statistics,
-            debug_callback=debug_callback,
-        ):
-            yield mapping
-
     def subgraph_monomorphisms_iter(
         self,
         *,
--- a/adam/perception/_matcher.py
+++ b/adam/perception/_matcher.py
@@ -560,11 +560,11 @@ class GraphMatching:
             yield mapping
 
     def subgraph_monomorphisms_iter(
-            self,
-            *,
-            collect_debug_statistics: bool = False,
-            debug_callback: Optional[Callable[[Any, Any], None]] = None,
-            initial_partial_match: Mapping[Any, Any] = immutabledict(),
+        self,
+        *,
+        collect_debug_statistics: bool = False,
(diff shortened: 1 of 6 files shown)


adam/perception/_matcher.py
```diff
@@ -559,6 +559,26 @@ def subgraph_isomorphisms_iter(
...
+        self,
+        *,
+        collect_debug_statistics: bool = False,
+        debug_callback: Optional[Callable[[Any, Any], None]] = None,
+        initial_partial_match: Mapping[Any, Any] = immutabledict(),
+    ):
+        """Generator over monomorphisms from G2 to a subgraph of G1."""
+        # Declare that we are looking for graph-subgraph isomorphism.
+        self.test = "mono"
+        self.initialize(initial_partial_match=initial_partial_match)
```
gabbard, line 572: Can you refactor the shared code between this and `sub_graph_isomorphisms_iter` into a "private" method they both call?

gabbard: Actually, if `subgraph_isomorphisms_iter` is no longer used, let's just remove it altogether.

spigo900: @gabbard Done.

spigo900:
This currently runs into some watermelon-related test failures. I'll look into the problem.

gabbard:
@spigo900 : you might check with @denizbeser about those test failures.

spigo900:
@gabbard I found the problem. Watermelons have some extra axis size relations that prevent the object recognizer from recognizing them as balls under node-induced subgraph isomorphism but not under edge-induced subgraph isomorphism.

The issue is [how we sort patterns](https://github.com/isi-vista/adam/blob/96ed8a22c14a9ea86bfd14335d6cbc44b464ce99/adam/learner/object_recognizer.py#L123). Balls end up coming first in that order so it always recognizes watermelons as balls. [Changing it](https://github.com/isi-vista/adam/pull/878/commits/4668462bfe7585c56acfc4a0530839effa573186) to sort on `(len((...)nodes), len((...)edges))` (rather than just `len((...)nodes)` fixes the object recognition problem and all the downstream issues that that caused.

spigo900:
Ready for re-review.

PR #879, opened 2020-07-08
ignore determiners in objects and attribute learner

paynesa:


adam/learner/attributes.py
```diff
@@ -112,6 +113,10 @@ def _candidate_templates(
                 # See: https://github.com/isi-vista/adam/issues/871
                 ignore_determiners=True,
             )
+            and not all(
+                (e in ["a", "the"] or isinstance(e, SyntaxSemanticsVariable))
+                for e in bound_surface_template.surface_template.elements
+            )
```
LichMaster98, line 119: Can you add a comment explaining this restriction?

PR #897, opened 2020-07-10
close #888

### LINKED ISSUE #888: Subset learner should return None rather than throw an error when a slot is pruned
For example, if a relation learner returns none, this means that this cannot be learned as a relation, while if it throws a Runtime error, it stops learning altogether.

paynesa:
closes #888

### DIFF
--- a/adam/learner/perception_graph_template.py
+++ b/adam/learner/perception_graph_template.py
@@ -165,6 +165,7 @@ class PerceptionGraphTemplate:
             if graph_logger:
                 graph_logger.log_graph(intersected_pattern, INFO, "Intersected pattern")
             for (_, object_wildcard) in template_variable_to_pattern_node.items():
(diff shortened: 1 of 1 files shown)


adam/learner/perception_graph_template.py
```diff
@@ -164,15 +164,9 @@ def intersection(
...
-                surface_template_variable,
-                object_wildcard,
-            ) in template_variable_to_pattern_node.items():
+            for (_, object_wildcard) in template_variable_to_pattern_node.items():
                 if object_wildcard not in intersected_pattern:
-                    raise RuntimeError(
-                        f"Result of intersection lacks a wildcard node "
-                        f"for template variable {surface_template_variable}"
-                    )
+                    return None
```
LichMaster98, line 169: Can you add a comment explaining why we return None here?

Issue #852, opened 2020-06-30
Subset verb learner tests are slow


spigo900:
Our subset verb learner tests take a long time, around 7 minutes. We'd like them to take less time.

The test that takes the longest time to run is `test_sit`. It tests a relatively large number of examples (8 templates, for each of those 10 train), but not combinatorially many (only 88 samples).

To profile `test_sit`, I ran the attached script using both cProfile and py-spy. From both profilers' results, it looks like most of the time is getting spent in the graph matching code via the object recognizer's `match_objects()`. The second biggest time sink is `PerceptionGraph.from_dynamic_perceptual_representation()`.

`cProfile` estimates 53.49% of the running time is in `match_objects()`, 39.08% is in `PatternMatching.matches()`, 43.54% in `_internal_matches()` and 37.91% in `subgraph_isomorphisms_iter()`. The second longest-running part of the code, `PerceptionGraph.from_dynamic_perceptual_representation()`, took 25.53% of the running time. Most of that time is spent in `translate_frames()`. The time sinks within that call seem to be `copy_with_temporal_scopes()` and `translate_frame()`.

(Note that these are all based on the PDF visualization. In that visualization, the test time is 97.35% of the total running time. The visualization includes the module loading and import time in its assessment, so the time spent executing the code is not 100% of the time.)

The script used for profiling and the profiling results (including the raw `pstats` file) are attached: [06-29_test_profiles.zip](https://github.com/isi-vista/adam/files/4853450/06-29_test_profiles.zip). The graph for `cProfile` was generated using [gprof2dot](https://github.com/jrfonseca/gprof2dot).

spigo900:
@gabbard [Per your comment](https://usc-isi.slack.com/archives/GL17LRSTG/p1593706576465200) I tried running the tests in Pypy on `adam-dev`, however it crashes while collecting the verb learner tests because `immutablecollections` doesn't know that Pypy's dictionaries have deterministic iteration order. Here's the error:

```
E   ValueError: ImmutableDicts can only be initialized from built-in dicts when the iteration of built-in dicts is guaranteed to be deterministic (Python 3.7+; CPython 3.6+)
```

It took some digging to find out but Pypy [does guarantee deterministic dictionary iteration order of dictionaries](https://doc.pypy.org/en/latest/cpython_differences.html?highlight=dictionary#miscellaneous):

[...]

PR #902, opened 2020-07-13
Tests: Use one object recognizer copy for each language mode

spigo900:
Optimization related to #852.


spigo900:
Ready to merge.

PR #907, opened 2020-07-14
Remove unnecessary usage of immutableset

### LINKED ISSUE #906: Unnecessary use of immutableset in GraphMatchingState.__init__ takes 2-3% of time
[In __init__ of `GraphMatchingState`](https://github.com/isi-vista/adam/blob/f7f4c5bb5d3ebff790a21cc8b812de4c5883da8a/adam/perception/_matcher.py#L897) (which is created repeatedly, once for each search step), we have the following pattern repeated three times:

```python
new_nodes_0: ImmutableSet[Any] = immutableset(
                flatten(
                    [
                        predecessor
                        for predecessor in GM.graph.predecessors(node)
                        if predecessor not in GM.graph_node_to_pattern_node
                    ]
                    for node in GM.graph_node_to_pattern_node
                )
            )
            for node in new_nodes_0:
                if node not in GM.graph_nodes_in_or_preceding_match:
                    GM.graph_nodes_in_or_preceding_match[node] = self.depth
```

We expect the collisions that `immutableset` is trying to deduplciated to be (a) relatively rare and (b) not to matter anyway since the assignment is idempotent.  Dropping the set and just using nested for loops should be faster, since creating an `immutableset` is relatively expensive.

spigo900:
Closes #906.

### DIFF
--- a/adam/perception/_matcher.py
+++ b/adam/perception/_matcher.py
@@ -897,36 +897,36 @@ class GraphMatchingState(object):
                 for predecessor in GM.graph.predecessors(node):
                     if (
                         predecessor not in GM.graph_node_to_pattern_node
-                        and node not in GM.graph_nodes_in_or_preceding_match
+                        and predecessor not in GM.graph_nodes_in_or_preceding_match
                     ):
-                        GM.graph_nodes_in_or_preceding_match[node] = self.depth
+                        GM.graph_nodes_in_or_preceding_match[predecessor] = self.depth
 
             # Updates for T_2^{in}
             for node in GM.pattern_node_to_graph_node:
                 for predecessor in GM.pattern.predecessors(node):
                     if (
                         predecessor not in GM.pattern_node_to_graph_node
-                        and node not in GM.pattern_nodes_in_or_preceding_match
+                        and predecessor not in GM.pattern_nodes_in_or_preceding_match
                     ):
-                        GM.pattern_nodes_in_or_preceding_match[node] = self.depth
+                        GM.pattern_nodes_in_or_preceding_match[predecessor] = self.depth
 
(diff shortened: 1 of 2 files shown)


adam/perception/_matcher.py
```diff
@@ -893,67 +893,40 @@ def __init__(self, GM: GraphMatching, graph_node=None, pattern_node=None) -> Non
...
-                )
-            )
-            for node in new_nodes_0:
-                if node not in GM.graph_nodes_in_or_preceding_match:
-                    GM.graph_nodes_in_or_preceding_match[node] = self.depth
+            for node in GM.graph_node_to_pattern_node:
+                for predecessor in GM.graph.predecessors(node):
+                    if (
+                        predecessor not in GM.graph_node_to_pattern_node
+                        and node not in GM.graph_nodes_in_or_preceding_match
```
gabbard, line 900: Should this be:
```suggestion
                        and predecessor not in GM.graph_nodes_in_or_preceding_match
```

spigo900: Hm, possibly, but isn't that what it was doing before?
[Line 908](https://github.com/isi-vista/adam/pull/907/files#diff-055afb1c07a0ca7f2989c5ab842afcefL908).

gabbard: @spigo900 : The naming here is confusing. Note that `node` on line 908 is coming from `new_nodes_0` and therefore corresponds to `precedessor` (and not `node`) from the old list comprehension.

spigo900: Ah, I see. Thanks!

spigo900: @gabbard That fixed it.

PR #908, opened 2020-07-14
Fix `covers_entire_utterance`

### LINKED ISSUE #904: Attribute learner learning prepositions as attributes in English
This results in phrases like "the ball on a on a table" when we have `ignore_determiners=True` when calling `covers_entire_utterance,` but when we set `ignore_determiners=False`, we can't learn true attributes such as colour because `covers_entire_utterance` always returns false. 

We know that this is an issue with determiners since Chinese is able to learn both and doesn't have determiners.

paynesa:
closes #904


parameters/experiments/m9/chinese_integrated_learner.params
```diff
@@ -0,0 +1,15 @@
...
+hypothesis_log_dir: "%experiment_group_dir%hypotheses_recognizer/"
+log_hypothesis_every_n_steps: 25
+
+sort_learner_descriptions_by_length: True
+num_pretty_descriptions: 10
+
+experiment: 'm9-integrated-learner-given-object-recognizer'
+curriculum: 'm9-complete'
+learner: 'integrated-learner-recognizer'
+language_mode: CHINESE
```
LichMaster98, line 15: Can you remove this param from the PR? It needs to be added to the `readme.md` but we're going to request individuals add `language_mode: CHINESE` to their `root.params` to modify the language mode so we don't have to duplicate all our experiment param files.

PR #915, opened 2020-07-15
Use PyPy to execute tests on Travis

spigo900:
Related to #852.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-    - "3.6"
+    - "pypy3.6-7.1.1"
 
 cache:
     pip: true
@@ -13,7 +13,7 @@ jobs:
     include:
         - stage: test
           script:
-              - echo "python version $(python --version) running"
+              - echo "pypy3 version $(pypy3 --version) running"
               - echo "pip version $(pip --version) running"
               - make check coverage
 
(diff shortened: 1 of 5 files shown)


spigo900:
This is currently failing because there's (apparently) no version of typed-ast available for PyPy, so it's unable to install Mypy. I think the next step is to see if we can get Travis to install and run Mypy using CPython and run the tests in PyPy.

I suspect (though I'm not familiar with Travis enough to be sure) that when you list multiple `python`s in the Travis file, it tests on all of the versions listed, using the same install/etc. scripts/commands for each version. If so, I think we could probably do this by having it execute something like "if (python exists on the path) then (run Mypy/etc.) else (run tests)." (This is a bit of a hack, though.)

spigo900:
Currently this is doing the two jobs we want it to do -- running `make check` with CPython and `make coverage` with PyPy.

It is also launching two extra jobs that try to install the requirements and do nothing else. I am having trouble figuring out why. I would expect the original build config to generate a single extra job like this if the problem is just matrix expansion at work, but that doesn't seem to be the case.

spigo900:
This is working. However, it is still generating two extra jobs. The jobs now pass because I explicitly specified that they should do nothing. I'm not sure why they're being generated in the first place.

PR #925, opened 2020-07-20
Add quantification semantics

gabbard:


gabbard:
This shouldn't get merged until after the milestone.

adam/axis.py
```diff
@@ -1,7 +1,6 @@
-from attr import attrs, attrib, evolve
-from attr.validators import instance_of
-
 from adam.utilities import sign
+from attr import attrib, attrs, evolve
```
gabbard, line 2: Sorry for all the import diff spam. I asked PyCharm to optimize imports without remembering that I don't think we do `isort` on this repository.

adam/experiment/log_experiment.py
```diff
@@ -210,6 +215,10 @@ def learner_factory_from_params(
                 beam_size=beam_size,
                 language_mode=language_mode,
             ),
+            number_learner=QuantifierTemplateLearner.pretrained_for_language_mode(
```
gabbard, line 218: I haven't tested at all that this doesn't cause problems for other curricula, which is the main reason this hasn't been merged to master yet. It could be totally buggy.

LichMaster98:
@paynesa is going to add unit tests for this learner

paynesa:
Ok this is currently failing about half of our existing unit tests, I'm going to look into debugging it

PR #926, opened 2020-07-21
Remove extra graph copying

spigo900:
Related to #852.

### DIFF
--- a/adam/perception/perception_graph.py
+++ b/adam/perception/perception_graph.py
@@ -357,8 +357,7 @@ class PerceptionGraph(PerceptionGraphProtocol):
 
     @staticmethod
     def copy_digraph_with_temporal_scopes(
-            digraph: DiGraph,
-            temporal_scopes: Union[TemporalScope, Iterable[TemporalScope]],
+        digraph: DiGraph, temporal_scopes: Union[TemporalScope, Iterable[TemporalScope]]
     ) -> DiGraph:
         wrapped_graph = digraph.copy()
 
@@ -386,7 +385,9 @@ class PerceptionGraph(PerceptionGraphProtocol):
                 "already dynamic"
             )
 
-        wrapped_graph = self.copy_digraph_with_temporal_scopes(self._graph, temporal_scopes)
+        wrapped_graph = self.copy_digraph_with_temporal_scopes(
+            self._graph, temporal_scopes
+        )
 
         return PerceptionGraph(dynamic=True, graph=wrapped_graph)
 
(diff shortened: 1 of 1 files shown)


adam/perception/perception_graph.py
```diff
@@ -2784,22 +2803,19 @@ def translate_frames(
         # First, we translate each of the two frames into PerceptionGraphs independently.
         # The edges of each graph are marked with the appropriate "temporal specifier"
         # which tells whether they belong to the "before" frame or the "after" frame.
-        before_frame_graph = (
-            self.translate_frame(perceptual_representation.frames[0])
-            .copy_with_temporal_scopes([TemporalScope.BEFORE])
-            .copy_as_digraph()
+        before_frame_graph = PerceptionGraph.copy_digraph_with_temporal_scopes(
```
gabbard, line 2806: Here (and for `AFTER` below), do we need to copy the result of `_translate_frame`, since there is no other reference to the returned `DiGraph`?

gabbard: That is, could we make an in-place `add_temporal_scopes_to_edges` function instead? The copy still take ~1/4 of this function's time in the profile.

spigo900: Yes, I think so.

spigo900: I changed it to add the temporal scopes in-place.

spigo900:
Ready for re-review.

adam/perception/perception_graph.py
```diff
@@ -355,6 +355,25 @@ def from_dynamic_perceptual_representation(
...
 
+    @staticmethod
+    def add_temporal_scopes_to_edges(
+        digraph: DiGraph, temporal_scopes: Union[TemporalScope, Iterable[TemporalScope]]
+    ) -> DiGraph:
+        """
+        Modifies the given digraph in place, applying the given `TemporalScope`\ s to all edges.
+        This new graph will be dynamic.
+
+        Note that this should only be applied to static perception digraphs.
```
LichMaster98, line 366: Can you add a check here for if the incoming dygraph is already marked and dynamic and if so throw a runtime error?

spigo900: @LichMaster98 Good thought. I added the check. It samples and checks an arbitrary edge (the first one) for dynamicness. Let me know if there's a better way to do this.

LichMaster98: hm, only other way I could think of would take more time which given this process is designed to save time I think this quick method is probably sufficient. 

This code will probably get slightly modified when/if we move to MultiDiGraphs so we can re-evaluate there if needed.

spigo900:
Ready to merge.

PR #927, opened 2020-07-21
Fix miscellaneous semantics 

paynesa:
Fixes the following: 

- keep rolling object on rolling surface
- make sure that the thrown object isn't necessarily always on the ground before and after
- make sure that a flying this isn't on the ground 
- add distinguishing between "jump" and "jump up" 
- add distinguishing between "run" and "run down" and "walk" and "walk down"
- the cup when drinking should be the same size throughout the action 
- learn both "over" and "under" (TODO: currently both of these have both objects on the ground #929 )

### DIFF
--- a/adam/curriculum/phase1_curriculum.py
+++ b/adam/curriculum/phase1_curriculum.py
@@ -1414,7 +1414,6 @@ def make_jump_template(
     agent: TemplateObjectVariable,
     *,
     use_adverbial_path_modifier: bool,
-    operator: PathOperator = None,
     spatial_properties: Iterable[OntologyNode] = immutableset(),
     background: Iterable[TemplateObjectVariable] = immutableset(),
 ) -> Phase1SituationTemplate:
@@ -1434,7 +1433,9 @@ def make_jump_template(
                         (
                             agent,
                             SpatialPath(
-                                operator=operator,
+                                operator=AWAY_FROM
+                                if use_adverbial_path_modifier
+                                else None,
                                 reference_object=GROUND_OBJECT_TEMPLATE,
                                 properties=spatial_properties,
                             ),
@@ -1541,7 +1542,6 @@ def _make_jump_curriculum(
                         make_jump_template(
                             jumper,
                             use_adverbial_path_modifier=use_adverbial_path_modifier,
-                            operator=operator,
                             background=background,
                         ),
                         ontology=GAILA_PHASE_1_ONTOLOGY,
@@ -1549,7 +1549,6 @@ def _make_jump_curriculum(
                         max_to_sample=num_samples if num_samples else 25,
                     )
                     for use_adverbial_path_modifier in (True, False)
-                    for operator in [TOWARD, AWAY_FROM]
                 ]
             ),
             flatten(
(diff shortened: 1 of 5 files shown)


adam/curriculum/phase1_curriculum.py
```diff
@@ -950,21 +951,19 @@ def _make_object_under_or_over_object_curriculum(
...
-    bird = object_variable("bird_0", BIRD)
-    object_under_bird = standard_object("object_under_bird_0")
 
     templates = [
         Phase1SituationTemplate(
             f"object-under-object",
             salient_object_variables=[object_above, object_under],
             constraining_relations=[bigger_than(object_above, object_under)],
-            asserted_always_relations=[strictly_over(object_above, object_under)],
+            asserted_always_relations=[strictly_under(object_above, object_under)],
```
LichMaster98, line 960: Is this written the right way? By default, I want to read this as the first argument is under the second argument.

adam/ontology/phase1_ontology.py
```diff
@@ -2436,7 +2426,19 @@ def _make_jump_description() -> Iterable[Tuple[OntologyNode, ActionDescription]]
...
-                ]
+                ],
+                # must be above the ground at some point during the action
+                at_some_point=[
+                    Relation(
+                        IN_REGION,
+                        jump_agent,
+                        Region(
+                            reference_object=JUMP_INITIAL_SUPPORTER_AUX,
+                            distance=DISTAL,
```
LichMaster98, line 2437: Does someone jump 'DISTAL' from the ground? I would have thought you jump `PROXIMAL`

adam/ontology/phase1_ontology.py
```diff
@@ -2452,10 +2454,21 @@ def _make_jump_description() -> Iterable[Tuple[OntologyNode, ActionDescription]]
...
```
LichMaster98, line 2465: Same distance comment here

adam/ontology/phase1_ontology.py
paynesa: Both of these are `at_some_point` relations because I wanted to specify that when you jump you leave the ground at some point -- i.e. you're on it then far away from it then on it again. Or is `PROXIMAL` a better choice because it indicates you don't go too far?

LichMaster98: I think `PROXIMAL` is a better choice here, but if we're consistent in choice it's not a huge deal. I just wasn't sure if there was previous conversation which led to the choice of `DISTAL` @gabbard do you have any opinion here?

adam/curriculum/phase1_curriculum.py
paynesa: Fixed, thanks for catching that!

adam/ontology/phase1_ontology.py
paynesa: I think `THROW` uses `DISTAL`, which is part of why I chose it, but I get the argument for `PROXIMAL`. I can change it if you both think that's a good idea

LichMaster98: We can skip with Distal

PR #933, opened 2020-07-22
add entry-point for noisy pursuit 

### LINKED ISSUE #734: Optionally use gaze in pursuit object learning and experiment with ablating its presence
- [ ] Modify the Pursuit object learner to (optionally, controlled by a flag on the learner object) give gazed objects a higher chance of being selected as the leading hypothesis.
- [ ] Modify the perceptual generation process to (optionally) introduce noise into the perception of gaze (i.e. omit it with a certain probability; add it spuriously with a certain probability).
- [ ] Create experimental scripts and curricula to explore how the variables of Pursuit learning rate, lexicalization threshold, and gaze ablation probabilities covary. 

You'll probably want to split all three of these off into separate issues and link them to this parent issue.

paynesa:
closes #734

### DIFF
--- a/adam/curriculum/m6_curriculum.py
+++ b/adam/curriculum/m6_curriculum.py
@@ -89,7 +89,8 @@ M6_CURRICULUM_ALL_OBJECTS = [
     JUICE,
     CUP,
     BOX,
-    CHAIR,
+    # TODO: https://github.com/isi-vista/adam/issues/935
+    # CHAIR,
     HEAD,
     MILK,
     HAND,
--- a/adam/curriculum/m6_curriculum.py
+++ b/adam/curriculum/m6_curriculum.py
@@ -90,7 +90,7 @@ M6_CURRICULUM_ALL_OBJECTS = [
     CUP,
     BOX,
     # TODO: https://github.com/isi-vista/adam/issues/935
-    # CHAIR,
+    CHAIR,
     HEAD,
     MILK,
     HAND,
--- a/adam/curriculum/m6_curriculum.py
+++ b/adam/curriculum/m6_curriculum.py
@@ -89,7 +89,6 @@ M6_CURRICULUM_ALL_OBJECTS = [
     JUICE,
     CUP,
     BOX,
-    # TODO: https://github.com/isi-vista/adam/issues/935
     CHAIR,
     HEAD,
     MILK,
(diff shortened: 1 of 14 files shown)


adam/curriculum/m6_curriculum.py
```diff
@@ -89,6 +89,7 @@
     JUICE,
     CUP,
     BOX,
+    # TODO: https://github.com/isi-vista/adam/issues/935
```
LichMaster98, line 92: I'm confused by this TODO here?

adam/experiment/experiment_utils.py
```diff
@@ -103,15 +106,37 @@ def build_pursuit_curriculum(
     num_objects_in_instance = pursuit_curriculum_params.integer(
         "num_objects_in_instance", default=3
     )
-
+    add_gaze = pursuit_curriculum_params.boolean("add_gaze", default=False)
+    ontology = pursuit_curriculum_params.string("ontology", default="phase1")
```
LichMaster98, line 110: I think we can safely just use the phase2 ontology here

adam/experiment/log_experiment.py
```diff
@@ -170,6 +176,19 @@ def learner_factory_from_params(
...
+            object_learner=PursuitObjectLearnerNew(
+                learning_factor=0.05,
+                graph_match_confirmation_threshold=0.7,
+                lexicon_entry_threshold=0.7,
+                rng=rng,
+                smoothing_parameter=0.002,
+                ontology=GAILA_PHASE_1_ONTOLOGY,
+                language_mode=language_mode,
+                rank_gaze_higher=True,
+            )
```
LichMaster98, line 190: Go ahead and add the rest of the learners here so if we want to run across the entire curriculum we can.

adam/experiment/log_experiment.py
```diff
@@ -170,6 +176,19 @@ def learner_factory_from_params(
...
         )
+    elif learner_type == "pursuit-gaze":
+        return lambda: IntegratedTemplateLearner(
+            object_learner=PursuitObjectLearnerNew(
+                learning_factor=0.05,
+                graph_match_confirmation_threshold=0.7,
+                lexicon_entry_threshold=0.7,
+                rng=rng,
+                smoothing_parameter=0.002,
+                ontology=GAILA_PHASE_1_ONTOLOGY,
```
LichMaster98, line 187: This should target `GAILA_PHASE_2_ONTOLOGY` to match the curriculum default. Because this is hardcoded perhaps we should remove the option to even select the phase 1 versions?

paynesa: Left the ontology selection for now for debugging, but I will make an issue to remove it at a later date

PR #939, opened 2020-07-24
Learn part-whole relations

gabbard:


spigo900:
[Fixing the intersection order for learner hypotheses](https://github.com/isi-vista/adam/pull/961#issuecomment-679173954) resolved the immediate problem (the increasing number of relaxations) for `eat`, however it unfortunately hasn't resolved the slowness generally. `test_sit` is still quite slow for example.

It looks like `test_sit` is hanging while intersecting hypotheses on the sixth observation (observation 5), "a bird sits down," for the third of the eight templates. (So it's in the template for sit_intransitive with `adverbial_mod=True, surface=False`, if that matters.) I ran `test_sit` with logging enabled (`--log-level INFO -rP --log-file some_file.txt --log-file-format "%(asctime)s -> %(message)s`) to get more information and there's a 20 minute wait between when it first gets the observation and when it finally starts doing relaxations. Once it starts doing relaxations it's fine and it finishes in a normal amount of time.

The nodes it ends up relaxing away are a `smallerThan` , a geon, a cross-section, a table leg object perception (the fourth leg, `**(furniture) leg 3`), and an axis.

There's an identical hang when the learner sees "a bird sits down on a table" later on, another one when it sees "a bird sits" (on a table), another when it sees "a bird sits on a table."

No other observations seem to cause hangs; it's able to learn "a bird sits on a box" and "a bird sits on a chair" before learning "a bird sits on a table" and there's no hang in either case.

spigo900:
This is slow.

It is slow because our matching algorithm is struggling to intersect certain complex verb hypotheses. Specifically, it spends a lot of time trying to match up subobjects in ways that make sense. The problem seems to be that it is hard to tell quickly whether (1) we are trying to match the wrong objects, or (2) there's no reasonable way to match things (so the pattern should be relaxed).

One example where the algorithm struggles is when we intersect "a bird sits on a table" with the current hypothesis after observing various things sitting on chairs. This comes up in `test_sit` under template #3.

I suspect the reason it struggles is our sort order. In order for us to figure out that the chair-shaped hypothesis can't be matched with the chair in any nice way, we need to do a lot of work looking for matches. That work is tangled up in the work we do to match up the agent hypothesis with the bird, which means it takes much longer for the chair matching to fail. My guess is that putting objects closer to their parts will result in faster failures.

Assuming that this is the case, I see a few possible solutions:

-   Change our matching algorithm
    -   Slightly &#x2013; incorporate VF2+ and/or VF2++ improvements
        -   This shouldn't be too difficult as these algorithms are incremental changes to VF2
        -   VF2++ in particular seems promising as the authors introduce a topologically-oriented "matching order" (sort order, I think) on the graph to speed up matching
            -   More on why I think that's a good thing later in this list
        -   I haven't yet read much about VF2+
    -   A lot &#x2013; switch to a VF3 variant
        -   VF3, or (more likely)
        -   VF3 Light
        -   Note that this would involve significant work
            -   I didn't find a Python implementation of VF3 when I searched
            -   The ["official" VF3 implementation](https://github.com/MiviaLab/vf3lib)
                -   does not currently support finding monomorphisms (i.e. non-induced subgraph isomorphisms), which is how we currently match graphs
                -   does not provide a Python wrapper
                -   is written in C++
                    -   so we would have to build wheels if we wanted to make a wrapper
-   Sort nodes differently
    -   I suspect the best way to do this would be to use a topological sort
        -   We can't do a "proper" one because our digraph is not a DAG &#x2013; it has cycles
            -   However, we can do a "topological-ish" sort &#x2013; [this public domain Python code](https://github.com/jeweg/topological-sorting) claims to implement such an algorithm
                -   My understanding is that in the presence of cycles it does the thing we'd want it to do:
                    -   Topologically order the nodes that can be ordered,
                    -   and put those nodes in the correct order relative to the nodes that can't be ordered because they are involved in cycles
        -   Our current way of sorting seems to approximate this
            -   Generally, referring types come before the types they reference
        -   And our current way of sorting showed a significant performance increase over matching unsorted graphs/patterns
        -   So heuristically, it seems like a good idea
    -   There's also a hackish way we could pre-sort things using the object recognizer
        -   Have the object recognizer recursively sort the nodes of each object subgraphs at some point before it returns
             -   Sort using our existing sort order
        -   Arrange things in a sequence of items (sorted object, \*sorted subobjects), putting root-level non-objects (if any) at the end
        -   Then use that ordering instead of the "order by type" ordering we use by default

Overall, I think VF2++ looks like the most promising possible solution, followed by topological sorting. I'm going to look at these in more detail going forward.

spigo900:
This **was** nearly ready for review. Notes:

- It is failing CI because Mypy doesn't like it when you have a keyword-only attribute that has a default value.
    - Unfortunately, it doesn't seem like you can instruct Mypy to ignore this error. I tried putting `# type: ignore` on several of the lines for the attribute and none of them got rid of the error.
    - This [has been fixed](https://github.com/python/mypy/pull/8803) in a (much) newer version of Mypy (0.780).
        - Upgrading to 0.782 seems to create another spurious error and I'm not sure whether that one is any easier to fix.
- Because there are merge conflicts, I tested a merged version on [another branch](https://github.com/isi-vista/adam/tree/882-part-whole-merge-maybe-broken/adam).
    - [Travis CI build](https://travis-ci.com/github/isi-vista/adam/jobs/396070399) of the tests.
    - It passes a lot of the unit tests.
    - However, as it is, it is definitely going to break the propose-but-verify learner.
        - For now I have it completely ignoring the `_match_template()` method because I need to get failure information for the functional learner hack to work.
    - It is failing
        - one of the object learner tests (both languages),
            - unsurprisingly, it's the propose-but-verify version of `test_subset_learner`
        - one of the subset attribute learner tests (both languages),
            - `test_subset_my_attribute_learner_integrated`
        - one of the subset preposition learner tests (both languages),
            - `test_subset_preposition_has`
        - one of the subset verb learner tests (both languages)
            - `test_sit` again
    - Also, test slowness has returned.
        - So far it is not so slow on any individual test that it will actually fail CI.
            - It's just back to being unpleasantly slow (>40 minutes to run).
        - Part of it is that the imprecise description tests is fairly slow.
            - `test_big` using the (old-style) subset attribute learner for English, if I'm reading it right.
                - I think the others for `test_big` are not as slow, though I'm not sure.
        - But also the subset verb learner tests seem to have slowed down again.

spigo900:
Actually, it **is** so slow that it fails CI. It just did.

spigo900:
The subset attribute learner is failing the test that it can learn "my." The problem is that every object X the attribute learner sees described as "my X" (my book, my cup, my hat, my cookie) has a primary axis that is directed. Thus it never subsets away the primary axis as an important feature of "my [slot1]".

This wasn't a problem previously. That's because the object recognizer removed (non-world-item) axes when it recognized an object. (It left in cross-sections.) On this branch, however, we preserve the axes of root objects, so the problem comes through.

I have thought of three reasonable solutions (only two good ones, IMO):

1. Ban watermelons in the train set, force them in the test set. Increase the number of training examples *if needed*. (It shouldn't be needed. Based on my testing, banning watermelons seems to cause it to include balls in the 5-sample train set, for whatever reason.)
2. Explicitly show it a situation corresponding to "my ball" during training so it loses the primary axis.
3. Just increase the train set size and don't worry about if it can generalize.

I prefer (1) or (2). @LichMaster98, @denizbeser, what are your opinions on this?

spigo900:
(Also, as an aside: `test_subset_preposition_has` had the same problem. It's testing on "person has ball." It was already showing specific objects/scenarios to the learner so I just changed one of them to show the learner a watermelon instead of a book.)

LichMaster98:
@spigo900 I think I prefer option 1.

spigo900:
@LichMaster98 I agree, I prefer option 1 as well.

denizbeser:
@spigo900 same here, option 1 sounds good.

lichtefeld:
Closing as P3 is not using ADAM's symbolic representation so this "part-whole" relationship will be reconsidered given the new perceptual layout.

PR #949, opened 2020-07-30
Debugging to run entire M13 without crashing

paynesa:
can run `run_m13` without encountering an error in English and Chinese now with `m13.params`

For `log_experiment` with `m13_complete`, our issue in English is "throw", since we have two possible "throw-to" templates; one indicates goal and one indicates recipient. For English, we express both with "to" but in Chinese we can distinguish since there is a different coverb indicating person goal ("gei")


paynesa:
Can run `log_experiment` with `m13_complete.params` in English, but waiting on #950 because we want to be sure that everything still runs after that's merged.

paynesa:
Now running `m13_complete` with both Chinese and English params correctly

PR #950, opened 2020-07-31
Add Source and Destination Reference Objects to Spatial Paths to resolve ambiguity

### LINKED ISSUE #932: SpatialPaths have ambiguity between source and destination which should be resolved

LichMaster98:
Closes #932


adam/curriculum/phase1_curriculum.py
```diff
@@ -705,7 +709,7 @@ def falling_template(
                             theme,
                             SpatialPath(
                                 operator=TOWARD,
-                                reference_object=ground,
+                                reference_source_object=ground,
```
paynesa, line 712: If the source is the beginning of the path, for falling wouldn't we expect the path to begin away from the ground and end on the ground, so the `reference_source_object` would be whatever it fell off of?

adam/curriculum/phase1_curriculum.py
```diff
@@ -740,17 +746,17 @@ def fall_on_ground_template(
                         (
                             theme,
                             SpatialPath(
-                                None,
-                                reference_object=GROUND_OBJECT_TEMPLATE,
+                                TOWARD,
+                                reference_source_object=ground,
```
paynesa, line 750: same comment re: falling

adam/curriculum/phase1_curriculum.py
```diff
@@ -2033,7 +2050,7 @@ def make_walk_run_template(
                             agent,
                             SpatialPath(
                                 operator=TOWARD,
-                                reference_object=goal,
+                                reference_source_object=goal,
```
paynesa, line 2053: Similar comment to falling; I would think if the goal is where the walker ends up, it shouldn't be the source of the spatial path? Marking subsequent related comments with :book:

adam/curriculum/phase1_curriculum.py
```diff
@@ -2875,8 +2899,8 @@ def _make_come_down_template(
                 argument_roles_to_fillers=[(AGENT, agent), (GOAL, goal_reference)],
                 during=DuringAction(
                     objects_to_paths=[
-                        (agent, SpatialPath(TOWARD, ground)),
-                        (agent, SpatialPath(TO, goal_reference)),
+                        (agent, SpatialPath(TOWARD, reference_source_object=ground)),
```
paynesa, line 2902: :book:

adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
```diff
@@ -655,7 +655,7 @@ def _go_towards_away_template(
                             agent,
                             SpatialPath(
                                 operator=TOWARD if is_toward else AWAY_FROM,
-                                reference_object=spatial_reference,
+                                reference_source_object=spatial_reference,
```
paynesa, line 658: :book: for most of the templates in this function

adam/curriculum/phase1_curriculum.py
```diff
@@ -687,6 +690,7 @@ def falling_template(
     syntax_hints: Iterable[str],
     spatial_properties: Iterable[OntologyNode] = immutableset(),
     background: Iterable[TemplateObjectVariable],
+    start_distal: bool,
```
paynesa, line 693: nit: perhaps add a comment indicating what start-distal means/accomplishes

adam/curriculum/verbs_with_dynamic_prepositions_curriculum.py
```diff
@@ -338,7 +338,7 @@ def _push_towards_away_template(
                             theme,
                             SpatialPath(
                                 operator=TOWARD if is_towards else AWAY_FROM,
-                                reference_object=spatial_reference,
+                                reference_source_object=spatial_reference,
```
paynesa, line 341: :book:

adam/language_specific/chinese/chinese_language_generator.py
```diff
@@ -360,12 +360,21 @@ def _collect_action_modifiers(
                         path
                         for (_, path) in action.during.objects_to_paths.items()
                         if (
-                            isinstance(path.reference_object, SituationObject)
-                            and path.reference_object.ontology_node == GROUND
+                            isinstance(path.reference_source_object, SituationObject)
```
paynesa, line 363: nit: maybe add a one-line comment here explaining what you're checking for on a high level

adam/language_specific/english/english_language_generator.py
```diff
@@ -828,7 +828,23 @@ def _collect_action_modifiers(
                     paths_involving_ground = immutableset(
                         path
                         for (_, path) in action.during.objects_to_paths.items()
-                        if path.reference_object.ontology_node == GROUND
+                        if (
```
paynesa, line 831: same comment as Chinese language generator

adam/ontology/phase1_ontology.py
```diff
@@ -2131,7 +2134,10 @@ def _make_drink_description() -> Iterable[Tuple[OntologyNode, ActionDescription]
     frame=ActionDescriptionFrame({THEME: _FALL_THEME}),
     during=DuringAction(
         objects_to_paths=[
-            (_FALL_THEME, SpatialPath(operator=TOWARD, reference_object=_FALL_GROUND))
+            (
+                _FALL_THEME,
+                SpatialPath(operator=TOWARD, reference_source_object=_FALL_GROUND),
```
paynesa, line 2139: :book:

adam/curriculum/phase1_curriculum.py
LichMaster98: @paynesa Yes, I agree. That's the confusion that led to wanting to be able to distinguish between the two. So the question is when we have a spatial path do we: Require at least one of a source or destination or Require both?

For "Falling" I think the source should be `Region(ground, distance=PROXIMAL or DISTAL, direction=Gravitational_up)` and the destination is `Region(ground, distance=EXTERIOR_BUT_IN_CONTACT_WITH, direction=Gravitational_up)` right?

Adding this level of distinction definitely helps resolve ambiguity but means some thought needs to be put into how we apply it to the curriculum (in both our dictionaries and the templates) so we're consistent across all the actions.

paynesa: @LichMaster98 I agree with your semantics for falling if there's not a specified origin (otherwise I think we'd want to start in contact with that origin?). 

I think requiring both can make sense -- for ones where there's not a clear direction (e.g. when we're just introducing a path for the purpose of imprecise temporal descriptions), we can probably just make the source and the origin the same since there's no motion with regards to that path, but that does seem a bit hackish. Happy to discuss this more!

Also small sidenote: I think you want to have this PR not close #932 since source/destination isn't fully implemented yet :)

LichMaster98: Ah! My intent is to add to this PR with the resolution of the source and destinations after we discuss how we want to resolve them. But figured getting a review of the technical changes first would help in aiding the discussion for how to go about modifying the templates.

PR #951, opened 2020-08-07
Fix indexing and add accuracy logging 

paynesa:
This fixes the issue that our HTML file was beginning indexing at -1; it and the logging now start at 0 

It also adds an optional way of logging accuracy at each iteration of the experiment to create graphs like the ones shown [here](https://www.dropbox.com/scl/fi/n1bgnqr3v4tj35lz0mzw2/Gaze-Ablation-in-Pursuit-Object-Learning.paper?dl=0&rlkey=rijevp5ei2y5orjwodednnr2g).

### DIFF
--- a/adam/experiment/observer.py
+++ b/adam/experiment/observer.py
@@ -101,6 +101,11 @@ class CandidateAccuracyObserver(
     Provide an accuracy score.
     """
     name: str = attrib(validator=instance_of(str))
+
+    # these params allow the accuracy to be written out to a text file at each step which is helpful for graphing for
+    # experiments such as the gaze ablation where we want to compare accuracy
+    accuracy_to_txt: bool = attrib(default=False)
+    txt_path: str = attrib(validator=instance_of(str), default="out.txt")
     _num_predictions: int = attrib(init=False, default=0)
     _num_predictions_with_gold_in_candidates: int = attrib(init=False, default=0)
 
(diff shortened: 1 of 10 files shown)


adam/experiment/observer.py
```diff
@@ -101,6 +101,11 @@ class CandidateAccuracyObserver(
     Provide an accuracy score.
     """
     name: str = attrib(validator=instance_of(str))
+
+    # these params allow the accuracy to be written out to a text file at each step which is helpful for graphing for
+    # experiments such as the gaze ablation where we want to compare accuracy
+    accuracy_to_txt: bool = attrib(default=False)
+    txt_path: str = attrib(validator=instance_of(str), default="out.txt")
```
LichMaster98, line 108: We'd probably prefer this default to add these records into the experiments output folder where the `index.html` is being logged to for consistency in where to find the outputs for an experiment.

parameters/experiments/m13/pursuit_with_noisy_gaze.params
```diff
@@ -5,6 +5,9 @@ _includes:
 experiment: 'pursuit-ablating-gaze'
 curriculum: pursuit
 learner: pursuit-gaze
+accuracy_to_txt : True
+accuracy_logging_path : "out_out.txt"
```
LichMaster98, line 9: Is this supposed to be a full path or just a file name? It would probably be clearer as the accuracy logging file name and programmatically placing it in the directory output where the `index.html` gets placed for the experiment. That way all the outputs from an experiment can be found in the same location by default.

adam/experiment/observer.py
paynesa: Makes sense, I'll update this!

parameters/experiments/m13/pursuit_with_noisy_gaze.params
paynesa: Ah sorry this was for debugging and I forgot to take it out -- but yes, I agree :)

LichMaster98: If its just debug that's fine too! imo, an example param doc showing how to include that file logging probably isn't a bad thing. Can you add it to the `debug.param` ?

paynesa: Yes, will do!

PR #952, opened 2020-08-07
Generate a bunch of gaze ablation params

paynesa:


adam/create_gaze_ablation_params.py
```diff
@@ -0,0 +1,77 @@
+from vistautils.parameters import Parameters
+from vistautils.parameters_only_entrypoint import parameters_only_entry_point
+import numpy as np
+
+
+def create_gaze_ablation_entry_point(params: Parameters) -> None:
+    """This function creates all possible gaze ablation param files within a given range"""
+    # get the parameters directory, which must be non-null
+    parameters_dir = params.optional_creatable_directory("parameters_directory")
```
LichMaster98, line 9: nit: prefer https://github.com/isi-vista/vistautils/blob/master/vistautils/parameters.py#L244 and let vistautils handle requiring this param

adam/create_gaze_ablation_params.py
```diff
@@ -0,0 +1,77 @@
...
+   lexicon_entry_threshold: 0.7
+   smoothing_parameter: .001
+
+pursuit-curriculum-params:
+   num_instances: {num_instances}
+   num_noise_instances: {num_noise_instances}
+   num_objects_in_instance: {num_objects_in_instance}
+   add_gaze : {add_gaze}
+   prob_given : {prob_given}
+   prob_not_given : {prob_not_given}"""
```
LichMaster98, line 70: nit: You could make this string a constant variable at the end of the file and use the `.format()` command to format the f string variables. This can make editing the string in the future easier and allow for reuse in other potential param generation scripts. 

That way if we need to make changes to this 'template' param string we can change it in one place in the future.

PR #955, opened 2020-08-12
Generics learner

denizbeser:


adam/learner/integrated_learner.py
```diff
@@ -217,92 +270,116 @@ def _linguistic_descriptions_from_semantics(
...
+                    return cur_string
+            # get the classifier and add it to the language
+            classifier = get_classifier_for_string(cur_string[-1])
+            if classifier:
+                return tuple(chain((classifier,), cur_string))
+            else:
+                return cur_string
+
+        # handle English determiners
         else:
```
LichMaster98, line 297: Should this instead be an `elif` to check if the language mode is English? I realize we only have two languages at the moment but handling it this way means if someone doesn't fix this specific hack all languages will get this processing when that may not be desired?

LichMaster98 (changes requested):

adam/language_specific/chinese/chinese_language_generator.py
```diff
@@ -1029,7 +1030,7 @@ def add_classifier(
                     IGNORE_HAS_AS_VERB not in self.situation.syntax_hints
                     and not self.situation.is_dynamic
                 ):
-                    return
+                    pass
                     # TODO: we currently return here since we can't handle one possessive node and one not for the third person (i.e. I have my ball, you have your ball
```
spigo900, line 1034: Update comment? Since it's no longer returning. Specifically, now we're passing -- why can we pass here/why do we want to do that instead of returning?

adam/learner/attributes.py
```diff
@@ -118,7 +118,11 @@ def _candidate_templates(
             # later learning of attributes since the learner may consider both the attribute and the object to be objects initially,
             # leading it to try to match two objects with a template that only has one slot
             and not all(
-                (e in ["a", "the"] or isinstance(e, SyntaxSemanticsVariable))
+                (
+                    e in ["a", "the"]
+                    or isinstance(e, SyntaxSemanticsVariable)
+                    or e[:3] == "yi1"
```
spigo900, line 124: `e in DETERMINERS`?

adam/learner/objects.py
```diff
@@ -129,7 +129,7 @@ def _candidate_templates(
             )
             if not language_alignment.token_index_is_aligned(tok_idx)
             # ignore determiners
-            and token not in ["a", "the"]
+            and token not in ["a", "the"] and not token[:3] == "yi1"
```
spigo900, line 132: `token not in DETERMINERS`?

adam/learner/subset.py
```diff
@@ -132,10 +132,10 @@ class AbstractSubsetLearnerNew(AbstractTemplateLearnerNew, ABC):
     _concept_to_hypotheses: Dict[Concept, ImmutableSet[PerceptionGraphTemplate]] = attrib(
         init=False, default=Factory(dict)
     )
-    _concept_to_surface_template: Dict[Concept, SurfaceTemplate] = attrib(
+    concept_to_surface_template: Dict[Concept, SurfaceTemplate] = attrib(
```
spigo900, line 135: Should these be public?

adam/learner/generics.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Actions: E.g dog s walk
+        # Attributes: E.g cookies are brown
+        # Kinds: E.g cookie is a food
+        #  Both of these statements are regular generics as the noun is indefinite. We just need to
+        #  check attributes in addition to actions. We hardwire "are" in order to recognize kinds.
+        # If there is a recognized object node that matches the scene, and a generic action OR attribute, learn!
+        for nodes in [action_nodes, attribute_nodes]:
+            if significant_object_concept and nodes:
+                # Get the slot of object for each recognized action in the statement
```
spigo900, line 99: Wouldn't this be for each recognized action in the scene? (So if there were actions in the scene not mentioned in the statement, those would also be handled here.)

adam/learner/integrated_learner.py
```diff
@@ -136,6 +167,27 @@ def observe(
             if self.functional_learner:
                 self.functional_learner.learn_from(current_learner_state, offset=offset)
 
+        # Engage generics learner if the utterance is indefinite
+        if self.generics_learner and not self.is_definite(current_learner_state):
+            # Lack of definiteness could me marking a generic statement
```
spigo900, line 172:
```suggestion
            # Lack of definiteness could be marking a generic statement
```

adam/learner/integrated_learner.py
spigo900: I agree.

adam/learner/surface_templates.py
```diff
@@ -120,14 +154,12 @@ def match_against_tokens(
...
                     )
-                # if there's a slot without a filler, we just don't learn from this instance instead of raising a runtime error
-                else:
-                    return None
-                    # raise RuntimeError(
-                    #    f"Template contained variable {element}, "
-                    #    f"but it was not found in the mapping of slots to spans: "
-                    #    f"{slots_to_filler_spans}"
-                    # )
+                # else:
```
spigo900, line 157: Why is this commented out? Do we still need it? Is it commented out until some issue can be fixed? Is it no longer relevant?

tests/learner/generics_learner_test.py
```diff
@@ -0,0 +1,165 @@
...
+            ontology=GAILA_PHASE_1_ONTOLOGY, beam_size=5, language_mode=language_mode
+        ),
+        action_learner=SubsetVerbLearnerNew(
+            ontology=GAILA_PHASE_1_ONTOLOGY, beam_size=5, language_mode=language_mode
+        ),
+        generics_learner=SimpleGenericsLearner(),
+    )
+
+
+def run_generics_test(learner, language_mode):
```
spigo900, line 58: Doesn't necessarily need fixing, but to confirm, the test here is just that this doesn't throw any errors?

adam/learner/generics.py
denizbeser: That's right, and in fact that's been helpful because we now also generalize "sit down on" when we hear "sits". We don't typically hypothesize verbs incorrectly, so it's not an issue unless we introduce noise to generics, which is not our plan.

adam/learner/subset.py
denizbeser: maybe not for this case, but accessibility comes in handy for semantics

tests/learner/generics_learner_test.py
denizbeser: Yes, and also the comments contain the outline for generating the output, which we will need for experiments

adam/learner/generics.py
spigo900: Okay. Then I suggest changing the comment:
```suggestion
                # Get the slot of object for each recognized action in the scene
```
(I understood "statement" to mean "the linguistic description.")

PR #957, opened 2020-08-13
Pegasus M13 experiments runner

spigo900:


spigo900:
This should be complete, but I think it makes sense to hold off on merging it until #949 is merged since that makes changes to the M13 runner as well.

adam/experiment/__init__.py
```diff
@@ -143,6 +169,19 @@ def execute_experiment(
...
+            "Cannot both resume from latest logged state and load a specified state file."
+        )
+
+    if resume_from_latest_logged_state and learner_logging_path is None:
+        raise RuntimeError(
+            "Need state logging path to be able to resume from latest learner state."
+        )
+
+    if resume_from_latest_logged_state and starting_point != -1:
+        raise RuntimeError("Starting point should not be specified when")
```
LichMaster98, line 183: This error message looks incomplete.

LichMaster98 (changes requested):

adam/experiment/__init__.py
spigo900: And it is. :)

Thanks for catching this.

spigo900: This should be fixed now.

LichMaster98 (approved):

Issue #958, opened 2020-08-14
Parallel curriculum generation


spigo900:
We'd like to be able to run our curriculum generation code on the cluster using Pegasus so that we can save time when running experiments on fixed curricula.

As part of this, we need a way to store our generated curricula, the obvious way being to use Pickle.

One of the difficulties is that pickling the curriculum causes issues with learning. In particular the subset verb learner unit tests run into failures to learn and crashes when they are modified to pickle and unpickle the curricula before using them.

One specific problem with serializing our curricula is that this does not preserve the identity of `eq=False` objects. This is likely one cause of the above errors, because our code makes assumptions like "there is only one GROUND perception." This assumption is violated when we unpickle a perception graph, since the unpickled GROUND perception is distinct from the one we use everywhere else. The object recognizer code [here][object_recognizer_identity_assumption] is one example where this assumption is used and where it would be violated.

[object_recognizer_identity_assumption]: https://github.com/isi-vista/adam/blob/5a3684d33c482b4bad4234b464110c033af05e08/adam/learner/object_recognizer.py#L296

There's another possible problem, namely that if we create our learner in code while unpickling our curriculum, our learner's ontology and our curriculum's ontology will be out of alignment -- they will no longer be equal. However testing suggests that this is not a problem in practice. First, the ontology graph is unaffected. Second, if we pickle and unpickle the learner together with the curricula, it preserves the alignment between their ontologies, but the same problems occur. I saw the same pattern of unit test failures when pickling and unpickling the learner and curricula together.

Overall I suspect the first problem is most likely to be the source of the problems, in particular due to the unit test failures when the learner and the curricula are pickled together, and due to the specific ways the code fails.

The specific failures are:

- `test_drink`
  - Both the English and Chinese tests crash while replacing objects.
  - The crash is due to attempting to get the successors of a node that isn't in the graph.
  - Specifically, the problem seems to be that it tries to get the successors of the `gravitational-up` axis in the graph.
  - It's unclear why this only happens for `drink`.
  - I'll try debugging to see what the problem is in this specific case.
- `test_fall` and `test_throw`
  - Only the English unit tests fail.
  - Both appear to be due to determiner failures.
  - For `test_fall`, it's a determiner failure, likely caused by the first problem above.
  - For `test_throw`, I suspect the problem is the same, because the descriptions include "a ground" rather than "the ground."
    - However, the failure message doesn't include enough words from the expected vs. actual descriptions to be sure.

spigo900:
[...]

PR #959, opened 2020-08-18
Add AdamPickler/Unpickler which handle persistence of shared world items and GROUND_PERCEPTION

spigo900:
This resolves the pickling problems discussed on #958.

### DIFF
--- a/adam/pickle.py
+++ b/adam/pickle.py
@@ -18,17 +18,17 @@ class AdamPickler(Pickler):
 
     @staticmethod
     def persistent_id(obj):
-        pid = None
+        persistent_id_ = None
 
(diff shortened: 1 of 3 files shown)


spigo900:
I was concerned there might be other problems with pickling and unpickling the curricula, however it appears there aren't, at least for the learners where we have unit tests. I modified `GeneratedFromSituationsInstanceGroup` to pickle and unpickle its perceptual representations using the `AdamPickler` and `AdamUnpickler` before yielding them. While the tests ran somewhat slower this way, I saw no test failures. Thus it looks like the solution as it stands should work fine for pickling and unpickling our curricula.

PR #960, opened 2020-08-20
Curriculum generation

### LINKED ISSUE #958: Parallel curriculum generation

spigo900:
Should allow for parallel curriculum generation (#958), although it does not implement any workflow to do that.

### DIFF
--- a/adam/experiment/curriculum_repository.py
+++ b/adam/experiment/curriculum_repository.py
@@ -1,3 +1,15 @@
+"""
+Code for working with *curriculum repositories*, which are structured directories containing
+saved curricula to be used in experiments.
+
+Curricula are saved and loaded to paths within the repository directory based on the parameters to
+the curriculum or script. We use a fixed parameter ordering to insure that curriculum paths are
+consistent across different runs and scripts.
+
+Since the curriculum parameters are not currently "sandboxed" away from script parameters, we
+include all parameters other than the ones that have been specifically ignored. The user can specify
+additional ignored parameters as appropriate. Unrecognized parameters are an error.
+"""
 from typing import Tuple, AbstractSet
 from pathlib import Path
 import pickle
--- a/adam/experiment/curriculum_repository.py
+++ b/adam/experiment/curriculum_repository.py
@@ -27,8 +27,10 @@ IGNORED_PARAMETERS: ImmutableSet[str] = immutableset(
     [
         "adam_root",
         "adam_experiment_root",
+        "experiment",
         "experiment_name",
         "experiment_group_dir",
+        "learner",
         "hypothesis_log_dir",
         "include_image_links",
         "accuracy_to_txt",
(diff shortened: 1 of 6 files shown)


spigo900:
I am conflicted about the ignore-list way of doing things. On the one hand it achieves the goal of not silently ignoring possible curriculum parameters. On the other hand it seems fragile; any time we add new parameters to the curriculum generation script or the experiment runner script (`log_experiment.py`) we have to add the new parameter to `IGNORED_PARAMETERS`, otherwise it will break either generating or using curricula.

I'm not sure if there's a good solution. However, on thinking about this, I think it would be slightly better thing to be less strict when reading curricula than when writing them. It should probably log the path it's reading from in case it's not the one you expected, but it doesn't need to crash.

LichMaster98 (approved):

PR #961, opened 2020-08-21
Run gaze ablation experiments using Pegasus

spigo900:
The workflow itself seems to function.

However, about 20 of the jobs reach a point where they hang and are eventually terminated. They each hang for at least 40 minutes. There's no obvious pattern in the pursuit parameters that hang. They seem to be hanging during pattern matching or relaxation. Sometimes the last messages before the hang are "blocking match against a non-subobject portion..." and sometimes they're "Relaxation: Last failed pattern node is..."/"Nodes to delete directly: ..."

Depends on the experiment script changes from #957.

### DIFF
--- a/adam/experiment/gaze_ablation.py
+++ b/adam/experiment/gaze_ablation.py
@@ -98,7 +98,6 @@ EXPERIMENT_NAME_FORMAT = (
 FIXED_PARAMETERS = {
     "curriculum": "pursuit",
     "learner": "pursuit-gaze",
-    "accuracy_to_txt": True,
     "pursuit": {
         "learning_factor": 0.05,
         "graph_match_confirmation_threshold": 0.7,
--- a/adam/experiment/gaze_ablation.py
+++ b/adam/experiment/gaze_ablation.py
@@ -85,6 +85,7 @@ def gaze_ablation_runner_entry_point(params: Parameters) -> None:
                                 experiment_name,
                                 log_experiment_entry_point,
                                 experiment_params,
+                                depends_on=[]
                             )
 
     write_workflow_description()
--- a/adam/experiment/gaze_ablation.py
+++ b/adam/experiment/gaze_ablation.py
@@ -11,7 +11,7 @@ import numpy as np
 from adam.experiment.log_experiment import log_experiment_entry_point
 
 
-def create_gaze_ablation_entry_point(params: Parameters) -> None:
+def gaze_ablation_runner_entry_point(params: Parameters) -> None:
     """This function creates all possible gaze ablation param files within a given range"""
     initialize_vista_pegasus_wrapper(params)
 
@@ -108,4 +108,4 @@ FIXED_PARAMETERS = {
 }
 
 if __name__ == "__main__":
-    parameters_only_entry_point(create_gaze_ablation_entry_point)
+    parameters_only_entry_point(gaze_ablation_runner_entry_point)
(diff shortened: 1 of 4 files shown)


spigo900:
I noticed that we're having to do more and more relaxations as we observe more and more examples and our hypothesis gets more general. This seems backwards.

I think it might also be part of our slowness problem since expanding the graph to include subobjects means we have to do a lot more relaxations to update our hypotheses.

I suspect the problem comes in [here](https://github.com/isi-vista/adam/blob/d39bec535791e6f93e1137c1fcf2ccdf13d7c6e3/adam/learner/verbs.py#L271) -- we call `previous_pattern_hypothesis.intersection(current_pattern_hypothesis, ...)`, which looks right, however [the actual intersection code](https://github.com/isi-vista/adam/blob/d39bec535791e6f93e1137c1fcf2ccdf13d7c6e3/adam/perception/perception_graph.py#L1026) uses the `current_pattern_hypothesis` as the pattern graph. What that actually means is that when we get to graph matching we're looking for a subgraph of the more specific graph (the previous pattern hypothesis) that is isomorphic to the bigger graph (the current pattern hypothesis), so we fail a lot and have to do a lot of relaxations.

I think we want to document this behavior in the intersection method's docstring and change the `update` methods to call the intersection method on the current pattern hypothesis instead.

We also could reverse how the intersection method sets up the matching (so that it passes `self` as the pattern) but I suspect it may be equally confusing either way. (@LichMaster98 , do you have an opinion on that?)

LichMaster98:
@spigo900 I believe I prefer the first option where we document the functionality on the `intersection` method's docstring and fix the order in the `update` methods.

Please feel free to rename the parameters to make them more clear. Part of this confusion is from the legacy of having intersections of both `PerceptionGraph` & `PatternGraph` and then needing to support `PatternGraph` to `PatternGraph` intersection as well. It's probably worth mentioning wherever you add the additional docstring that a future user could consider splitting these two distinct intersections into two functions for clarity.

adam/experiment/gaze_ablation.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def gaze_ablation_runner_entry_point(params: Parameters) -> None:
+    """This function creates all possible gaze ablation param files within a given range"""
+    initialize_vista_pegasus_wrapper(params)
+
+    # Get the baseline experiment parameters for gaze ablation -- these are things common to all of
+    # the experiments, like:
+    #
+    #     include_image_links: true
```
paynesa, line 21: possible debugging code to remove?

parameters/experiments/gaze_ablation_runner.params
```diff
@@ -0,0 +1,36 @@
...
+workflow_name: 'gaze_ablation_pegasus'
+workflow_directory: '%adam_experiment_root%/%experiment%'
+site: 'saga'
+namespace: 'saga'
+
+backend: slurm
+partition: ephemeral
+num_cpus: 1
+num_gpus: 0
+memory: '2G'  # not sure how much is needed
```
paynesa, line 34: Guessing you already checked this but if not, it runs in about 2 hours on my laptop with 16 RAM if that helps!

paynesa (approved):

parameters/experiments/gaze_ablation_runner.params
spigo900: That's correct; I'll get rid of the comment. 2G was enough if I recall correctly.

adam/experiment/gaze_ablation.py
spigo900: No, this is just explaining what the `gaze_ablation` namespace is expected to include. Thanks for pointing out the possibility, though.

PR #966, opened 2020-08-31
Add Propose But Verify learner to Objects & simple unit test.

### LINKED ISSUE #545: Test Propose-but-verify model of word learning

LichMaster98:
Closes #545

### DIFF
--- a/adam/learner/propose_but_verify.py
+++ b/adam/learner/propose_but_verify.py
@@ -1,0 +1,222 @@
+import logging
+from abc import ABC, abstractmethod
+
+from immutablecollections import ImmutableSet, immutableset
+from more_itertools import first
+
+from typing import Dict, Optional, AbstractSet, Iterable, Tuple
+
+from attr import attrs, attrib, Factory
+from attr.validators import instance_of, optional
+
+from adam.learner import (
+    SurfaceTemplateBoundToSemanticNodes,
+    LanguagePerceptionSemanticAlignment,
+    SurfaceTemplate,
+)
+from adam.learner.learner_utils import compute_match_ratio, pattern_match_to_semantic_node
+from adam.learner.perception_graph_template import PerceptionGraphTemplate
+from adam.learner.template_learner import AbstractTemplateLearnerNew
+from adam.ontology.ontology import Ontology
+from adam.perception import MatchMode
+from adam.perception.perception_graph import (
+    GraphLogger,
+    DebugCallableType,
+    PerceptionGraph,
+    PerceptionGraphPatternMatch,
+)
+from adam.random_utils import RandomChooser
+from adam.semantics import Concept, SemanticNode
+
+
+@attrs
+class AbstractProposeButVerifyLearner(AbstractTemplateLearnerNew, ABC):
+    """
+    An Abstract Implementation of the Propose but Verify Learning Model
+
+    This model does not retain any information pertaining to
+    disconfirmed hypothesis and is based on a "win-stay, lose-shift" ideology.
+
+    Reference: https://www.ling.upenn.edu/~ycharles/papers/pursuit-final.pdf
+    Section 1 of this paper on pursuit learning includes an introduction to the
+    'propose but verify' learning approach.
+    """
+
+    _ontology: Ontology = attrib(validator=instance_of(Ontology), kw_only=True)
+    _observation_num = attrib(init=False, default=0)
+    _concept_to_hypotheses: Dict[Concept, ImmutableSet[PerceptionGraphTemplate]] = attrib(
+        init=False, default=Factory(dict)
+    )
+    _surface_template_to_concept: Dict[SurfaceTemplate, Concept] = attrib(
+        init=False, default=Factory(dict)
+    )
+    _concept_to_surface_template: Dict[Concept, SurfaceTemplate] = attrib(
+        init=False, default=Factory(dict)
+    )
+    # Internal Params
+    _rng: RandomChooser = attrib(validator=instance_of(RandomChooser))
+    _graph_logger: Optional[GraphLogger] = attrib(
+        validator=optional(instance_of(GraphLogger)), default=None
+    )
+    _debug_callback: Optional[DebugCallableType] = attrib(default=None)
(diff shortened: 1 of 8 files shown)


adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,225 @@
...
+            # We are confident in a hypothesis if we don't have any alternatives.
+            if len(hypotheses) == 1
+        )
+
+    def _fallback_templates(
+        self
+    ) -> Iterable[Tuple[Concept, PerceptionGraphTemplate, float]]:
+        # Alternate hypotheses stored in the beam.
+        return (
+            (concept, hypothesis, 1.0)
```
paynesa, line 175: Is there a reason we default to 1? I would think if these are fallback templates we'd want it to be lower but maybe that's not an issue?

adam/learner/propose_but_verify.py
```diff
@@ -0,0 +1,225 @@
...
+                debug_callback=self._debug_callback,
+            )
+            for match in matcher.matches(use_lookahead_pruning=True):
+                semantic_node_for_match = pattern_match_to_semantic_node(
+                    concept=concept, pattern=pattern, match=match
+                )
+                # A template only has to match once; we don't care about finding additional matches.
+                return match, semantic_node_for_match
+            # We raise an error if we find a partial match but don't manage to match it to the scene
+            raise RuntimeError(
```
paynesa, line 219: To we want to raise an error here or just refuse to learn?

adam/learner/pursuit.py
```diff
@@ -1205,3 +1211,27 @@ def _fallback_templates(
...
+        pattern: PerceptionGraphTemplate,
+        perception_graph: PerceptionGraph,
+    ) -> Optional[Tuple[PerceptionGraphPatternMatch, SemanticNode]]:
+        """
+        Try to match our model of the semantics to the perception graph
+        """
+        matcher = pattern.graph_pattern.matcher(
+            perception_graph,
+            match_mode=MatchMode.NON_OBJECT,
+            # debug_callback=self._debug_callback,
```
paynesa, line 1228: Debugging line commented out; can probably just remove

adam/learner/subset.py
```diff
@@ -359,3 +365,27 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
+        pattern: PerceptionGraphTemplate,
+        perception_graph: PerceptionGraph,
+    ) -> Optional[Tuple[PerceptionGraphPatternMatch, SemanticNode]]:
+        """
+        Try to match our model of the semantics to the perception graph
+        """
+        matcher = pattern.graph_pattern.matcher(
+            perception_graph,
+            match_mode=MatchMode.NON_OBJECT,
+            # debug_callback=self._debug_callback,
```
paynesa, line 382: debugging code can be removed

adam/learner/template_learner.py
```diff
@@ -370,27 +390,24 @@ def _enrich_common(
...
+            # Its possible there is no match for the template in the graph
+            # So we handle a None return here
+            if rtrn:
+                (match, semantic_node_for_match) = rtrn
                 match_to_score.append((semantic_node_for_match, score))
                 # We want to replace object matches with their semantic nodes,
                 # but we don't want to alter the graph while matching it,
                 # so we accumulate these to replace later.
                 if isinstance(concept, ObjectConcept):
                     matched_objects.append((semantic_node_for_match, match))
```
paynesa, line 407: What's the reason for removing the return here?

adam/learner/propose_but_verify.py
LichMaster98: All confidence values are currently 1, it's available so algorithms can take advantage of it but we don't currently use it anywhere.

adam/learner/propose_but_verify.py
LichMaster98: I'd prefer to raise an error here because this case would mean we identified a partial match in the scene which matches our hypothesis enough to say "Yes this is the object we think it is" but we were unable to match the reduced pattern to the perception -- Which would be an error.

adam/learner/pursuit.py
LichMaster98: I'd prefer to leave it in, because we should be able to pass the `debug_callback` into this function (as the learner should have an optional slot for one) but the pursuit learner currently doesn't. Leaving it commented makes it easier for someone to manually enable this debugging feature if needed explicitly until it's properly supported in this learner.

adam/learner/subset.py
LichMaster98: See: Answer to the above in pursuit.

adam/learner/template_learner.py
LichMaster98: I moved to this internal `_match_template` method because the Propose but Verify learner model never modifies its internal hypothesis of what a given concept is other than to completely replace it. So there is a modifiable 'confidence' value in how much of the hypothesized pattern overlaps with the actual pattern in the scene. This ability to say "Oh well 90% of the pattern matches so that's ok" means that because our objects are realized as a complex graph structure rather than a single object we can accept some amount of deviance from the object. 

Previously this function required a 100% exact match so by moving to a function where the implementation of the `TemplateLearner` can decide what counts as "a matched template" gives us more flexibility in implementing other learning algorithms.

adam/learner/pursuit.py
LichMaster98: Perhaps we should track this missing feature in an issue referenced here to explain the commented line?

PR #967, opened 2020-08-31
Add objects for Chinese Classifier learning 

paynesa:


tests/learner/functional_learner_test.py
```diff
@@ -23,15 +24,18 @@ def integrated_learner_factory(language_mode: LanguageMode):
...
+        5, 0, phase2_language_generator(language_mode)
+    )
     sit_test = _make_sit_on_chair_curriculum(
         1, 0, phase2_language_generator(language_mode)
     )
 
     learner = integrated_learner_factory(language_mode)
 
     for (_, linguistic_description, perceptual_representation) in sit_train.instances():
+        print(linguistic_description)
```
denizbeser, line 38: Probably left from debugging - can you remove the print statement?

tests/learner/subset_verb_learner_test.py
```diff
@@ -133,6 +133,7 @@ def run_verb_test(learner, situation_template, language_generator):
         linguistic_description,
         perceptual_representation,
     ) in train_curriculum.instances():
+        print(linguistic_description)
```
denizbeser, line 136: Also print statement

adam/curriculum/phase1_curriculum.py
```diff
@@ -2052,7 +2056,6 @@ def make_take_template(
                 ),
             )
         ],
-        constraining_relations=[bigger_than(agent, theme)],
```
denizbeser, line 2055: Hmm why did we have to remove the constraining relations from these templates?

paynesa: Since we are now filling templates with e.g. bear and cat, which have different relative size to the agents we're currently used to, we aren't able to learn anything unless we remove these.

denizbeser: So e.g both `Mom` and a `bear` can be agents, and there are some objects smaller than `bear` but larger than `mom`? I thought the algorithm, once it matches the objects, only cares about the relations between them so this shouldn't matter? I might be misunderstanding the issue

paynesa: Right, I thought that the algorithm could handle this as well but these constraints cause the learning to completely break and removing them fixes it so I don't think it can.

denizbeser: Okay! Can you make an issue for that?

denizbeser: And I will merge once CI passes

PR #969, opened 2020-09-01
Yi classifiers

paynesa:
This adds the 1 + CLF + noun structure to Chinese singular nouns. However, it also breaks most of our learning since we don't have a reliable way to acquire the classifiers right now which is why CI is broken. 

I believe that this is the behaviour we'd actually expect under our current learning algorithms so I'm unsure exactly how to continue. My original proposed solution to the generics learning problem was to consider the interpretation of the nouns in generic statements to be plural by default so we don't have to deal with this (i.e. we just have to learn `2/many_nouns` vs. `nouns`), but this is quite hackish so I don't know if it's a good fallback or not and would require us to change our generics perceptions to be plural. 

In either case for the generics, this may potentially pose a problem for real learning of Chinese classifiers. I haven't yet tried running the entire learner on it, but if it fails most of the unit tests, I'm not optimistic.

### DIFF
--- a/adam/language_specific/chinese/chinese_language_generator.py
+++ b/adam/language_specific/chinese/chinese_language_generator.py
@@ -1030,7 +1030,7 @@ class SimpleRuleBasedChineseLanguageGenerator(
                     IGNORE_HAS_AS_VERB not in self.situation.syntax_hints
                     and not self.situation.is_dynamic
                 ):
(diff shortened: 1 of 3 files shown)


paynesa:
This PR is now fixed based on a hackish implementation similar to English determiners.

PR #975, opened 2020-09-18
Object Language Ablation Experiments

### LINKED ISSUE #978: Document Configurable Experimentation Parameters
The number of configurable parameters on our experiments has grown to the point where we need a centralized description of the possible configuration changes and their impacts on experiments. 

Additionally, this coverage should document the parameters needed to run experiments via Pegasus Workflow Management.

### LINKED ISSUE #979: Log Experiment Observers state when we log experiment data
Currently a restarted experiment would be unable to continue reporting observer data from where it left off. We should pickle and unpickle our observers or store the necessary information to restore these observers to continue reporting accurate observations of an experiment when restarted. This functionality is especially useful for jobs running via Pegasus on the saga cluster.

LichMaster98:
Closes #978, Closes #979

### DIFF
--- a/adam/curriculum/__init__.py
+++ b/adam/curriculum/__init__.py
@@ -3,12 +3,11 @@ Code to specify what is shown to `LanguageLearner`\ s and in what order.
 """
 from abc import ABC, abstractmethod
 
-from immutablecollections import ImmutableSet
-from typing import Generic, Iterable, Optional, Tuple, List
+from typing import Generic, Iterable, Optional, Tuple
 
 from attr import attrib, attrs
 from attr.validators import instance_of
-from immutablecollections.converter_utils import _to_tuple, _to_immutableset
+from immutablecollections.converter_utils import _to_tuple
 
 from adam.language import LinguisticDescriptionT
 from adam.language.language_generator import LanguageGenerator
@@ -180,13 +179,13 @@ class AblatedLanguageSituationsInstanceGroup(
     """
     The name of the instance group.
     """
-    _instances: List[
+    _instances: Tuple[
         Tuple[
             Optional[SituationT],
             LinguisticDescriptionT,
             PerceptualRepresentation[PerceptionT],
         ]
-    ] = attrib()
+    ] = attrib(converter=_to_tuple)
     """
     Instances already instantiated so we store them as a tuple
     """
--- a/adam/curriculum/__init__.py
+++ b/adam/curriculum/__init__.py
@@ -180,13 +180,13 @@ class AblatedLanguageSituationsInstanceGroup(
     """
     The name of the instance group.
     """
-    _instances: ImmutableSet[
+    _instances: List[
         Tuple[
             Optional[SituationT],
             LinguisticDescriptionT,
             PerceptualRepresentation[PerceptionT],
         ]
-    ] = attrib(converter=_to_immutableset)
+    ] = attrib()
     """
     Instances already instantiated so we store them as a tuple
     """
--- a/adam/experiment/README.md
+++ b/adam/experiment/README.md
@@ -1,0 +1,79 @@
+# Experimentation Setup
+
+Our experiments can be influenced by a number of parameters. This guide is intended to list all potential parameters, their impacts on the relevant portion of the curriculum, and any conflicting settings.
+
+## Generic Parameter Control
+These parameters are independent of any specific experiment.
+### Curriculum Params
+- curriculum: String - The name of the curriculum to generate
+    - Valid options: "m6-deniz", "each-object-by-itself", "pursuit", "m6-preposition", "m9-objects", "m9-attributes", "m9-relations", "m9-events", "m9-debug", "m9-complete", "m13-imprecise-size", "m13-imprecise-temporal", "m13-subtle-verb-distinction", "m13-object-restrictions", "m13-functionally-defined-objects", "m13-generics", "m13-complete", "m13-verbs-with-dynamic-prepositions", "m13-shuffled", "m13-relations", "m15-object-noise-experiments"
+- pursuit-curriculum-params: Namespace - The configuration for the pursuit specific curriculum
+- use-path-instead-of-goal: Boolean (Optional) - If true some curriculums use a path indication rather than a goal
+- num_samples: Integer (Optional) - The number of each specific situation type to instantiate
+- num_noise_objects: Integer (Optional) - The number of noise objects in a situation
+
+### Experiment Params
+- experiment: String - The experiment name
+- debug_log_directory: Path (Optional) - The path to log associated debug information, in this case it enables debug graph logging
+- language_mode: String - The language the experiment should use. This string value converts to an enumerated value
+    - Valid options: "English", "Chinese"
+- experiment_group_dir: Path - The path where the experiment will be logged to
+- learner: String - The name of the learner type to use
+    - Valid options: "pursuit", "object-subset", "preposition-subset", "attribute-subset", "verb-subset", "integrated-learner", "integrated-learner-recognizer", "pursuit-gaze", "integrated-object-only"
+- beam_size: Integer (Optional) - Used to constrain the search space for learner algorithms
+
+- pre_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's output of a situation before the learn process is run on the new input.
+- post_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's output of a situation after the learning process is ran on the new input.
+- test_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's performance on the test situations.
+- hypothesis_log_dir: Path (Optional) - This path is a directory to log internal learner hypotheses into
+- log_hypothesis_every_n_steps: Integer (Optional) - The number of situations to observe before the system logs the internal state of the experiment
+- log_learner_state: Boolean (Optional) - If true, save the internal trained learner state in a pickle.
+- starting_point: Integer (Optional) - Useful when debugging to start training from a specific situation number in the curriculum. Can not be used if `resume_from_latest_logged_state` is true.
+- point_to_log: Integer (Optional) - Useful debugging tool causes a specific log of hypotheses, learner, & observer states (future functionality for Observers)
+- learner_state_path: Path (Optional) - Provide a path to a pickled learner file to load to continue an experiment with. Used commonly in combination with `starting_point`. Can not be used if `resume_from_latest_logged_state` is true.
+- resume_from_latest_logged_state: Boolean (Optional) - Restart an experiment from the most recent logged state that can be located in the directory structure
+- debug_learner_pickling: Boolean (Optional) - If true, log additional debug information for the learner when pickling
+
+### Observer Params
+Currently the accessible observers by these parameters are HTML Observers which can include other sub-observer types
+- include_acc_observer: Boolean (Optional) - Include an accuracy observer
+- accuracy_to_txt: Boolean (Optional) - Log accuracy information to a file
+- accuracy_logging_path: String (Optional) - File path as string to log accuracy information to if `accuracy_to_txt` is true
+- include_pr_observer: Boolean (Optional) - Include a Precision and Recall Observer
+- log_pr: Boolean (Optional) - Log precision and recall information to a file
+- pr_log_path: String (Optional) - File path as string to log precision and recall information to if `log_pr` is true
+
+### Optional: Pegasus Workflow Params
+This section of parameters is specific to using the Pegasus workflow management tool in combination with the ISI Saga cluster.
+See [Pegasus Wrapper Overview][https://github.com/isi-vista/vista-pegasus-wrapper/blob/master/docs/api_overview.rst] for more details on how to use this section
+
+## Learner Parameters
+### Subset
+None
+
+### Propose But Verify
+- graph_match_confirmation_threshold: Float - A value between 0 and 1 which indicates how much a new hypothesis must match a previous pattern to count as a successful match.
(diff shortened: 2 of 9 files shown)


adam/experiment/experiment_utils.py
```diff
@@ -50,6 +55,71 @@
...
+    Restore the output report files to the appropriate number of instances to match the loaded observer
+    """
+    lines_to_write_back: List[str] = []
+    with open(str(path), "r") as file:
+        for num, line in enumerate(file):
+            if num < num_instances:
+                lines_to_write_back.append(line.strip())
+
+    # Remove the old file
+    os.remove(str(path))
```
spigo900, line 94: Does it make sense to use `os.remove` over `path.unlink()` here? If there's a reason, then this should get a comment, otherwise, maybe use the method:
```suggestion
    path.unlink()
```

adam/experiment/experiment_utils.py
```diff
@@ -50,6 +55,71 @@
...
+    path.write_text("\n".join(lines_to_write_back))
+
+
+def restore_html_state(path: Path, num_instance: int) -> None:
+    """
+    Restore an experiment HTML output to a specific number of instances to match the experiment state
+    """
+    lines_to_write_back: List[str] = []
+    count_instances: int = 0
+    with open(str(path), "r") as file:
```
spigo900, line 106:
```suggestion
    with path.open("r") as file:
```

adam/experiment/log_experiment.py
```diff
@@ -106,32 +112,116 @@ def log_experiment_entry_point(params: Parameters) -> None:
...
+        ):
+            try:
+                with observers_state_path.open("rb") as f:
+                    observers_holder = pickle.load(f)
+                    pre_observer = observers_holder.pre_observers
+                    post_observer = observers_holder.post_observers
+                    test_observer = observers_holder.test_observers
+            except OSError:
+                logging.warning(
+                    "Unable to open preobserver state at %s; skipping.",
```
spigo900, line 144:
```suggestion
                    "Unable to open observer state at %s; skipping.",
```

adam/experiment/log_experiment.py
```diff
@@ -106,32 +112,116 @@ def log_experiment_entry_point(params: Parameters) -> None:
...
+                    str(observers_state_path),
+                )
+            except pickle.UnpicklingError:
+                logging.warning(
+                    "Couldn't unpickle preobserver state at %s; skipping.",
```
spigo900, line 149:
```suggestion
                    "Couldn't unpickle observer state at %s; skipping.",
```

adam/experiment/log_experiment.py
```diff
@@ -106,32 +112,116 @@ def log_experiment_entry_point(params: Parameters) -> None:
...
+    elif observers_state:
+        try:
+            with observers_state.open("rb") as f:
+                observers_holder = pickle.load(f)
+                pre_observer = observers_holder.pre_observers
+                post_observer = observers_holder.post_observers
+                test_observer = observers_holder.test_observers
+        except OSError:
+            logging.warning(
+                "Unable to open preobserver state at %s; skipping.", str(observers_state)
```
spigo900, line 185:
```suggestion
                "Unable to open observer state at %s; skipping.", str(observers_state)
```

adam/experiment/log_experiment.py
```diff
@@ -106,32 +112,116 @@ def log_experiment_entry_point(params: Parameters) -> None:
...
+            )
+        except pickle.UnpicklingError:
+            logging.warning(
+                "Couldn't unpickle preobserver state at %s; skipping.",
```
spigo900, line 189:
```suggestion
                "Couldn't unpickle observer state at %s; skipping.",
```

adam/experiment/log_experiment.py
```diff
@@ -356,7 +481,7 @@ def curriculum_from_params(
     ]
 
     num_samples = params.optional_positive_integer("num_samples")
-    num_noise_objects = params.optional_positive_integer("num_noise_objects")
+    num_noise_objects = params.optional_integer("num_noise_objects")
```
spigo900, line 484: Why the change? Do we want to allow zero as an explicit value?

(It's odd, but [apparently](https://github.com/isi-vista/vistautils/blob/ffdc70e900715d7befd5c0e349371f6f58d180cc/vistautils/parameters.py#L564) vistautils doesn't support specifying a range on optional integers as opposed to required ones, otherwise we could use that.)

adam/experiment/observer.py
```diff
@@ -281,35 +302,36 @@ def create_logger(params: Parameters) -> "LearningProgressHtmlLogger":
...
         if include_links_to_images is None:
             include_links_to_images = False
 
         logging.info("Experiment will be logged to %s", output_html_path)
 
-        with open(output_html_path, "w") as outfile:
-            html_dumper = CurriculumToHtmlDumper()
+        html_dumper = CurriculumToHtmlDumper()
+        if not output_html_path.exists():
+            with open(str(output_html_path), "w") as outfile:
```
spigo900, line 314:
```suggestion
            with output_html_path.open("w") as outfile:
```

adam/experiment/observer.py
```diff
@@ -281,35 +302,36 @@ def create_logger(params: Parameters) -> "LearningProgressHtmlLogger":
...
+                      } else {
+                        x.style.display = "none";
+                      }
+                    }
+                    </script>
+                    """
+                )
         return LearningProgressHtmlLogger(
-            outfile_dir=output_html_path,
+            outfile_dir=str(output_html_path),
```
spigo900, line 334: Should this be a string?

adam/experiment/observer.py
```diff
@@ -423,23 +445,26 @@ def test_observer(
             if experiment_group_dir
             else "pr_test_out.txt",
         )
-        return HTMLLoggerPostObserver(
-            name="Test-observer",
-            html_logger=self,
-            candidate_accuracy_observer=CandidateAccuracyObserver(
+        a_ob = None
+        pr_ob = None
```
spigo900, line 449: Less-abbreviated names?

adam/experiment/observer.py
```diff
@@ -611,20 +636,15 @@ class HTMLLoggerPreObserver(  # pragma: no cover
     html_logger: LearningProgressHtmlLogger = attrib(
         init=True, validator=instance_of(LearningProgressHtmlLogger), kw_only=True
     )
-    candidate_accuracy_observer = attrib(  # type: ignore
-        validator=optional(CandidateAccuracyObserver), kw_only=True  # type: ignore
-    )
-    precision_recall_observer = attrib(  # type: ignore
-        validator=optional(PrecisionRecallObserver), kw_only=True  # type: ignore
-    )
+    candidate_accuracy_observer = attrib(kw_only=True)  # type: ignore
```
spigo900, line 639: Why no validators?

adam/experiment/README.md
```diff
@@ -0,0 +1,83 @@
...
+- beam_size: Integer (Optional) - Used to constrain the search space for learner algorithms
+
+- pre_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's output of a situation before the learn process is run on the new input.
+- post_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's output of a situation after the learning process is ran on the new input.
+- test_observer: namespace (Optional) - This namespace parameter is an Observer Params section. This observer reviews the learner's performance on the test situations.
+- hypothesis_log_dir: Path (Optional) - This path is a directory to log internal learner hypotheses into
+- log_hypothesis_every_n_steps: Integer (Optional) - The number of situations to observe before the system logs the internal state of the experiment
+- log_learner_state: Boolean (Optional) - If true, save the internal trained learner state in a pickle.
```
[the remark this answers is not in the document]
LichMaster98: This value defaults to true, but I will add that note as well.  @spigo900 Should I list defaults for any parameters which assume a default here somewhere as well?

adam/experiment/observer.py
LichMaster98: Yes, the `LearningProgresssHtmlLogger` is expecting this "output_dir" parameter to be a string -- It's also a full output path and not a dir so its a confusing parameter name. I'll see if I can refactor it without causing too many issues.

adam/curriculum/__init__.py
```diff
@@ -180,13 +180,13 @@ class AblatedLanguageSituationsInstanceGroup(
     """
     The name of the instance group.
     """
-    _instances: ImmutableSet[
+    _instances: List[
```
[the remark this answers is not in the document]
LichMaster98: @spigo900 Thanks for the catch! `immutablecollections.converter_utils` provides a `_to_tuple` converter for use in attrs which doesn't seem to produce a mypy warning.

adam/experiment/README.md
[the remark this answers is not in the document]
spigo900: @LichMaster98 Yes, I think it makes sense to list the defaults.

adam/experiment/README.md
```diff
@@ -0,0 +1,83 @@
+# Experimentation Setup
+
+Our experiments can be influenced by a number of parameters. This guide is intended to list all potential parameters, their impacts on the relevant portion of the curriculum, and any conflicting settings.
+
+## State Restoration
+As described below, our experiments have a parameter `resume_from_latest_logged_state` which, if true, does attempt to restart an experiment from the latest logged state. However, currently the experiment may restore the observers and the learners to different states IF the 'latest state' doesn't match between these logged folders.
```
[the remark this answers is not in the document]
spigo900: @LichMaster98 True. It's probably fine then.

adam/curriculum/__init__.py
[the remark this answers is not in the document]
spigo900: @LichMaster98 Aha, I wasn't aware of that converter. Thanks! That seems to resolve the problem for me as well.

adam/experiment/log_experiment.py
spigo900: @LichMaster98 I agree, and I think that solution makes sense.

adam/experiment/__init__.py
```diff
@@ -197,7 +223,20 @@ def execute_experiment(
...
+        observer_path = learner_logging_path / "observer_state"
+        if not observer_path.exists():
+            try:
+                observer_path.mkdir()
+            # if we don't have a directory where we can log our observer state, we simply don't log it
+            except OSError:
+                logging.warning("Cannot log observer state to %s", str(observer_path))
+                log_learner_state = False
+                logging.warning(
+                    "Proceeding without logging learner state. The experiment results from the observers won't be valid if this is not logged."
```
spigo900, line 238: Slightly confusing for similar reasons; here I think the "this" is supposed to refer to the observer state? So maybe something like this:
```suggestion
                    "Proceeding without logging learner. The experiment results from the observers won't be valid if the experiment is resumed without a log of the observer state."
```

adam/experiment/observer.py
spigo900: Did this cause issues?

adam/experiment/log_experiment.py
```diff
@@ -294,6 +382,42 @@ def learner_factory_from_params(
...
+                    "lexicon_entry_threshold"
+                ),
+                rng=rng,
+                smoothing_parameter=learner_params.floating_point("smoothing_parameter"),
+                ontology=GAILA_PHASE_2_ONTOLOGY,
+                language_mode=language_mode,
+            )
+        else:
+            raise RuntimeError(f"Invalid Object Learner Type Selected: {learner_type}")
+        return lambda: IntegratedTemplateLearner(object_learner=object_learner)
```
spigo900, line 420: Will we ever call this lambda more than once per execution? I assume not, but if we do then this might cause problems since the object learner will be the same every time.

If we will call it more than once, it might make more sense to `def` the learner-making function in each branch and return the function here. (I'm not sure if Python lets you define functions with the same name in different branches, but I can't see why it wouldn't.)

adam/experiment/__init__.py
```diff
@@ -197,7 +223,20 @@ def execute_experiment(
             except OSError:
                 logging.warning("Cannot log learner state to %s", str(learner_path))
                 log_learner_state = False
-                logging.warning("Proceeding without logging learner state")
+                logging.warning(
+                    "Proceeding without logging learner state. The experiment results from the observers won't be valid if this is not logged."
```
spigo900, line 227: I find this slightly confusing as "this" would seem to refer to the learner state and it's not clear why that would make the observer results invalid except for when the experiment is resumed. Maybe something like this?
```suggestion
                    "Proceeding without logging learner state. The experiment results from the observers won't be valid if the experiment is resumed without a log of the learner state."
```

spigo900 (commented):
Looks good, just a few comments

adam/experiment/log_experiment.py
LichMaster98: @spigo900 I'm unsure if this would ever get called more than once per execution but I was following the templates above for how this function is returning learners as I figured it was best not to change the return.

spigo900: @LichMaster98 Change the return type? It shouldn't require changing the type. Or do you mean changing something else about the return?

Slightly different but largely the same -- you could also `def object_learner_factory()` in each branch and that would let you keep the return expression similar to the template. It would just change `object_learner=object_learner` to `object_learner=object_learner factory()`.

spigo900: To clarify, when I say each branch, I mean each branch of `if object_learner_type == ...`, not of the outer if-statement.

LichMaster98: Oh! That makes sense, I'll make that change. Thanks

spigo900 (approved):
Looks good!

PR #987, opened 2020-10-23
Semantics Experiments

denizbeser:


adam/experiment/observer.py
```diff
@@ -636,13 +633,20 @@ class HTMLLoggerPreObserver(  # pragma: no cover
     html_logger: LearningProgressHtmlLogger = attrib(
         init=True, validator=instance_of(LearningProgressHtmlLogger), kw_only=True
     )
-    candidate_accuracy_observer: Optional[  # type: ignore
-        CandidateAccuracyObserver
-    ] = attrib(kw_only=True, validator=optional(instance_of(CandidateAccuracyObserver)))
+
+    # fmt: off
```
LichMaster98, line 637: What is this control (?) comment doing?

adam/experiment/run_kind_semantics_experiment.py
```diff
@@ -0,0 +1,188 @@
...
+def run_experiment(learner, curricula, experiment_id):
+    # Teach each pretraining curriculum
+    for curriculum in curricula:
+        print("Teaching", curriculum.name())
+        for (
+            _,
+            linguistic_description,
+            perceptual_representation,
+        ) in curriculum.instances():
+            # print(' '.join(linguistic_description.as_token_sequence()))
```
LichMaster98, line 71: 🐛 Debug code?

adam/experiment/run_kind_semantics_experiment.py
```diff
@@ -0,0 +1,188 @@
...
+    results_df.insert(0, "Words", pseudoword_to_kind.keys())
+    print(results_df.to_csv(index=False))
+    learner.log_hypotheses(Path(f"./renders/{experiment_id}"))
+
+    learner.render_semantics_to_file(
+        graph=learner.semantics_graph,
+        graph_name="semantics",
+        output_file=Path(f"./renders/{experiment_id}/semantics.png"),
+    )
+    # nx.write_gpickle(learner.semantics_graph, f"{experiment_id}_graph.gpickle")
```
LichMaster98, line 132: De-🐛 ?

adam/language_specific/english/english_phase_1_lexicon.py
```diff
@@ -127,6 +130,9 @@
         (HEAD, LexiconEntry("head", NOUN, plural_form="heads")),
         (PAPER, LexiconEntry("paper", NOUN, plural_form="papers")),
         (BEAR, LexiconEntry("bear", NOUN, plural_form="bears")),
+        (COW, LexiconEntry("cow", NOUN, plural_form="cows")),
+        (BEEF, LexiconEntry("beef", NOUN, plural_form="beefs")),
+        (CHICKEN, LexiconEntry("chicken", NOUN, plural_form="chickens")),
```
LichMaster98, line 135: These need to be added to the Chinese lexicon or we could hit issue generating curriculum in Chinese.

adam/learner/attributes.py
```diff
@@ -70,45 +70,48 @@ def _candidate_templates(
...
                                 ),
-                                language_mode=self._language_mode,
-                            ),
-                            {SLOT1: object_node},
+                                {SLOT1: object_node},
+                            )
                         )
-                    )
+                except RuntimeError:
+                    continue
```
LichMaster98, line 114: Why are we ignoring a runtime error here?

adam/learner/integrated_learner.py
```diff
@@ -161,14 +173,18 @@ def observe(
...
+                        sub_learner.learn_from(current_learner_state, offset=offset)
+                    current_learner_state = sub_learner.enrich_during_learning(
+                        current_learner_state
+                    )
+                    # Check definiteness after recognizing objects
+                    if sub_learner == self.object_learner:
+                        self.learn_definiteness_markers(current_learner_state)
+                except Exception as e:  # pylint:disable=broad-except
+                    logging.exception(e)
+                    # pass
```
LichMaster98, line 187: Also, why are we broadly ignoring an exception here?

adam/learner/integrated_learner.py
```diff
@@ -186,7 +202,7 @@ def observe(
             descs = self._linguistic_descriptions_from_semantics(
                 current_learner_state.perception_semantic_alignment
             )
-            # If the statement isn't a recognized sentence, run learner
+            # If the statemesnt isn't a recognized sentence, run learner
```
LichMaster98, line 205: Typo

adam/learner/integrated_learner.py
```diff
@@ -620,6 +636,47 @@ def update_concept_semantics(
                     self.semantics_graph.add_edge(
                         obj_con, other_con, slot=slot, weight=1.0
                     )
+                    # if the object is a wug - a new object heard through generics
+                    if not any(
+                        [
+                            obj_con.debug_string == c.debug_string
```
LichMaster98, line 642: Logic shouldn't be based off of the debug strings -- Why are they being referenced here?

adam/learner/object_recognizer.py
```diff
@@ -730,11 +776,22 @@ def _add_relationships_linked_to_root_object_perception(
         output_graph.add_edge(matched_object_node, linked_property_node, label=edge_label)
 
 
+# better name?
```
LichMaster98, line 779: I think this name is fine

adam/learner/objects.py
```diff
@@ -520,7 +512,7 @@ def _keep_hypothesis(  # pylint: disable=arguments-differ
         hypothesis: PerceptionGraphTemplate,
         bound_surface_template: SurfaceTemplateBoundToSemanticNodes,  # pylint:disable=unused-argument
     ) -> bool:
-        if len(hypothesis.graph_pattern) < 2:
+        if len(hypothesis.graph_pattern) < 3:
```
LichMaster98, line 515: Document this change to explain why 3 is now our bound -- The comment seems to imply 2.

adam/learner/objects.py
```diff
@@ -705,7 +697,13 @@ def templates_for_concept(self, concept: Concept) -> ImmutableSet[SurfaceTemplat
...
 
     def log_hypotheses(self, log_output_path: Path) -> None:
-        pass
+        for concept, hypothesis in self.concepts_to_patterns().items():
+            hypothesis.render_to_file(
+                graph_name="perception",
+                output_file=Path(
+                    log_output_path / f"{str(type(self))}-{concept.debug_string}"
+                ),
+            )
```
LichMaster98, line 706: I think this function is designated for abstract implementation given learners could theoretically store internal concepts to patterns which we could also need to display?

adam/learner/objects.py
```diff
@@ -774,7 +772,7 @@ def _keep_hypothesis(  # pylint: disable=arguments-differ
         hypothesis: PerceptionGraphTemplate,
         bound_surface_template: SurfaceTemplateBoundToSemanticNodes,  # pylint:disable=unused-argument
     ) -> bool:
-        if len(hypothesis.graph_pattern) < 2:
+        if len(hypothesis.graph_pattern) < 3:
```
LichMaster98, line 775: Likewise documenting the reason to change to 3?

adam/learner/perception_graph_template.py
```diff
@@ -14,19 +22,12 @@
...
-    ImmutableDict,
-    immutabledict,
-    immutablesetmultidict,
-    ImmutableSetMultiDict,
-)
-from immutablecollections.converter_utils import _to_immutabledict
 
 
-@attrs(frozen=True, slots=True, eq=False)
+@attrs(slots=True, eq=False)
```
LichMaster98, line 30: Why not frozen?

adam/learner/plurals.py
```diff
@@ -82,6 +82,8 @@ def _can_learn_from(
             for s in language_perception_semantic_alignment.perception_semantic_alignment.semantic_nodes
         ]
         counts = collections.Counter(concepts)
+        if not counts.values():
+            return False
         return max(counts.values()) > 1
```
LichMaster98, line 87: nit:
```suggestion
        return max(counts.values()) > 1 if counts.values() else False
```

adam/learner/semantics_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    similarity_matrix = np.zeros((len(nodes_to_embeddings), len(nodes_to_embeddings)))
+    for i, (_, embedding_1) in enumerate(nodes_to_embeddings.items()):
+        for j, (_, embedding_2) in enumerate(nodes_to_embeddings.items()):
+            similarity_matrix[i][j] = cos_sim(embedding_1, embedding_2)
+    names = [n.debug_string for n in nodes_to_embeddings.keys()]
+    df = pd.DataFrame(data=similarity_matrix, index=names, columns=names)
+    plt.rcParams["figure.figsize"] = (20.0, 20.0)
+    plt.rcParams["font.family"] = "serif"
+    # sb.heatmap(df)
+    # sb.clustermap(df)
```
LichMaster98, line 92: de-🐛 ?

adam/learner/semantics_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    sb.clustermap(df, row_cluster=True, col_cluster=True)
+    # cm.ax_row_dendrogram.set_visible(False)
+    # cm.ax_col_dendrogram.set_visible(False)
+    # plt.show()
```
LichMaster98, line 96: de- 🐛 ?

adam/learner/template_learner.py
```diff
@@ -434,12 +449,18 @@ def match_template(
...
         matched_objects.sort(key=by_pattern_complexity, reverse=True)
         already_replaced: Set[ObjectPerception] = set()
         new_nodes: List[SemanticNode] = []
+
         for (matched_object_node, pattern_match) in matched_objects:
+            # perception_graph_after_matching = replace_match_with_object_graph_node(
+            #     matched_object_node=cast(ObjectSemanticNode, matched_object_node),
+            #     current_perception=perception_graph_after_matching,
+            #     pattern_match=pattern_match
+            # )
```
LichMaster98, line 463: de-🐛 ?

adam/learner/verbs.py
```diff
@@ -268,6 +268,23 @@ def _update_hypothesis(
...
+        #         [
+        #             (node2, node1)
+        #             for previous_slot, node1 in previous_pattern_hypothesis.template_variable_to_pattern_node.items()
+        #             for new_slot, node2 in current_pattern_hypothesis.template_variable_to_pattern_node.items()
+        #             if previous_slot == new_slot
+        #         ]
+        #     ),
+        # )
+        # intersection.render_to_file("int", 'intersection')
+        # raise RuntimeError
```
LichMaster98, line 287: de-🐛 ?

adam/ontology/phase2_ontology.py
```diff
@@ -87,7 +87,7 @@
 TWO = OntologyNode("two")
 subtype(TWO, PERCEIVABLE_PROPERTY)
 MANY = OntologyNode("many")
-subtype(TWO, PERCEIVABLE_PROPERTY)
+subtype(MANY, PERCEIVABLE_PROPERTY)
```
LichMaster98, line 90: Huh, I wonder how long this bug has been here - Good catch!

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
```diff
@@ -585,7 +586,8 @@ def _perceive_property_assertions(self) -> None:
                     for r in self._situation.always_relations
                 )
             ):
-                properties_to_perceive.append(TWO_DIMENSIONAL)
+                # properties_to_perceive.append(TWO_DIMENSIONAL)
+                pass
```
LichMaster98, line 590: Why not add this property?

tests/learner/object_learner_test.py
```diff
@@ -280,8 +280,8 @@ def test_subset_learner_subobject():
         for semantic_node in enriched.semantic_nodes
     }
     assert (ObjectSemanticNode, "Mom") in semantic_node_types_and_debug_strings
-    assert (ObjectSemanticNode, "head") in semantic_node_types_and_debug_strings
-    assert (ObjectSemanticNode, "hand") in semantic_node_types_and_debug_strings
+    # assert (ObjectSemanticNode, "head") in semantic_node_types_and_debug_strings
+    # assert (ObjectSemanticNode, "hand") in semantic_node_types_and_debug_strings
```
LichMaster98, line 284: de-🐛 ?

LichMaster98 (changes requested):

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
spigo900: @denizbeser and I ended up having the problem that the subset object learner couldn't recognize juice in a cup because it had only observed it "free-standing," which was causing issues down the line with the semantics experiment. For debugging purposes (?) we simply disabled adding the property.

tests/learner/object_learner_test.py
spigo900: I believe it was. This should probably be marked `xfail` or `skip`.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
LichMaster98: Hmm, it seems like the better solution here is to show the learner a liquid in a container in the training stages rather than removing this property?

spigo900: I think I agree.

adam/experiment/observer.py
denizbeser: It prevents reformatting file, which otherwise breaks the type: ignore. You and I hada conversation about that typing bug a while back.

LichMaster98: Cool! I'm going to make a note about this in the future in case I need that.

adam/learner/attributes.py
denizbeser: Added a comment!

adam/learner/integrated_learner.py
denizbeser: We don't want entire experiments to break because of a single code issue. Exceptions vary a lot - we log it but the comment is necessary for pylint

LichMaster98: Hm, I'd disagree that we wouldn't want the entire experiment to break because of a single code issue. That code issue could invalidate the results of the experiment because it could hide a bug. 

If there's a specific exception we'd potentially expect to get that could be strictly ignored I'd be more ok to say that ignoring the error is fine (Say we have a specific error type we are using with dotgraph viz) but just broadly ignoring exceptions worries me from Ryan's continual guidance of "Problems should raise runtime errors so they don't hide bugs which are harder to debug!"

adam/learner/objects.py
denizbeser: We need this to see what objects look like when we use object recognizer

adam/learner/perception_graph_template.py
denizbeser: Double checking. Likely changed during a potential bug fix

adam/learner/integrated_learner.py
denizbeser: We still log and debug the exception as needed. This is just a safety belt - we don't have any repeating errors, and we have none for semantics actually, but there were rare errors with e.g one instance in plurals with mislearned object representations. I agree that it's better to keep them as runtime and fix, but it's too rare and infrequent that it's impractical to do otherwise given the timeline.

LichMaster98: Sure, my concern would be accidentally introducing a new repeating runtime error which is hidden by this exception rather than any current issues. Ideally even mislearned object representations shouldn't cause a runtime error though I would agree it's impractical to fix the rare error you saw given the timeline.

adam/perception/high_level_semantics_situation_to_developmental_primitive_perception.py
denizbeser: I couldn't find an existing curriculum for that. I suggest we leave it as is given the time constraints.

LichMaster98: Ok, @denizbeser  let's open a git issue to track this change then so it is explained why this occurred and link to it in the code.

LichMaster98 (approved):

PR #988, opened 2020-10-23
Pursuit Integrated Experiment Curriculum

### LINKED ISSUE #983: Curriculum to mirror actual language learner experience
Implement a curriculum generator that mirrors actual child language learner experience.

Description:
These situations consist of one (1) to ten (10) objects in the scene. These objects may have specific spatial arrangements. Objects in the scene may have specific attributes which can be linguistically expressed such as "tall". 

To-Do:
[ ] - Neatly Ordered Curriculum which we believe we _can_ learn
[x] - Topical Order Curriculum - A curriculum which stays on a relative topic. I'm thinking I'm going to implement this by adding a new way of generating situations that follows some probabilities for switching the focal point of the topic along with variable noise control. - Replaced by #986 
[ ] - Initial implementation of Noise elements including talking about situations that are partially or not present.
[ ] - Random Ordered Curriculum - This takes the `neatly ordered` curriculum and presents a larger sample in random order.
[ ] - Test Curriculum

LichMaster98:
Closes #983

### DIFF
--- a/adam/curriculum/preposition_curriculum.py
+++ b/adam/curriculum/preposition_curriculum.py
@@ -1,8 +1,9 @@
 from itertools import chain
-from typing import Iterable, Sequence, Optional
+from typing import Iterable, Sequence, Optional, Any, Tuple
 from immutablecollections import immutableset
 from more_itertools import flatten
 from adam.language.language_generator import LanguageGenerator
+from adam.relation import Relation
 from adam.situation.high_level_semantics_situation import HighLevelSemanticsSituation
 from adam.language.dependency import LinearizedDependencyTree
 from adam.axes import HorizontalAxisOfObject, FacingAddresseeAxis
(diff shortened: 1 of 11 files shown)


adam/curriculum/preposition_curriculum.py
```diff
@@ -113,19 +120,20 @@ def _under_template(
...
     is_training: bool,
     is_distal: bool,
-    syntax_hints: Iterable[str] = [],
+    syntax_hints: Iterable[str] = immutableset(),
+    background_relations: Iterable[TemplateObjectVariable] = immutableset(),
 ) -> Phase1SituationTemplate:
     handle = "training" if is_training else "testing"
-    # TODO: currently this hack keeps old implementation for English that hasn't solved https://github.com/isi-vista/adam/issues/802
-    # and returns new implementation for Chinese that does solve this
+    relations = [
```
spigo900, line 127: Was the subject of the comment resolved here?

adam/curriculum/preposition_curriculum.py
```diff
@@ -139,19 +147,20 @@ def _over_template(
...
```
spigo900, line 154: Commented problem resolved?

spigo900 (approved):
Looks good to me, just a few minor comments.

adam/curriculum/preposition_curriculum.py
LichMaster98: Yes, at least according to the git issue.

spigo900: Right. What I was confused about was whether this reference meant that "this is the problem that needs to be fixed" (and since the issue is, the problem is) or "this problem was mostly fixed, which is why the issue is resolved, but this specific case was not fixed, and here is the issue which gives more context about why/the way in which it was not fixed." Though if it were the latter case then there should have been a separate issue, so this is probably resolved.

Issue #984, opened 2020-10-14
Investigate implementation of an Attribute Pursuit Learner


LichMaster98:

denizbeser:
Just looked at the code again - I don't why this wouldn't be possible. All we need to combine `AbstractPursuitLearnerNew` and `AbstractAttributeTemplateLearnerNew` to create a "PursuitAttributeLearnerNew". For requirements, we can check the existing object version called `PursuitObjectLearnerNew`, and the subset version for attributes called `SubsetAttributeLearnerNew`. I couldn't spot any methods that would cause issues, but my guess is that anything non-object would run much slower since the hypothesis space is much larger compared to just objects (explicitly extracted with `extract_candidate_objects`) or the convergent learning method of intersecting/subsetting hypotheses; I suspect the search space is identical to learning verbs.

LichMaster98:
Ok, @denizbeser Do you have time to do the implementation of this learner or should I factor that into the time it will take me to get this experiment running?

Additionally, If the search space does prove to be an issue we could write an `extract_candidate_attributes` where we explicitly limit the search space (documented) to some restricted set correct? We could artificially say "We will only consider attributes which can be explored with X jumps away from the candidate object being described"?

denizbeser:
@LichMaster98 I'm busy with a couple of semantics experiments - can you work on it and I can jump in to help if you run into any issues? 
Yes, we should indeed trim the space with a specific extraction function, just like we did for obhects. We would need to discuss it's requirements though. Let's slack about meeting to discuss that.

LichMaster98:
Sounds good. Thanks for looking into this for me.

PR #995, opened 2020-11-12
Pursuit attribute learner

spigo900:
Related to #984.

### DIFF
--- a/adam/learner/attributes.py
+++ b/adam/learner/attributes.py
@@ -475,8 +475,11 @@ def _extract_candidate_attributes(
             if label == HAS_PROPERTY_LABEL
         ]
     )
-    # For testing purposes, use the much smaller set of candidate attributes
+    # Furthermore, we limit the search space to the even smaller set of hypotheses
     # where we consider only single properties as possible attributes.
+    # Otherwise there are too many hypotheses for the pursuit learner to search through
+    # and it's unlikely to converge on the correct hypothesis
+    # in any reasonable amount of time or number of examples.
     candidate_attribute_subgraph_node_sets = [
         immutableset([object_with_attribute, property]) for property in properties
     ]
--- a/adam/learner/attributes.py
+++ b/adam/learner/attributes.py
@@ -447,6 +447,20 @@ class PursuitAttributeLearnerNew(  # should it be called new? there is no old on
                     log_output_path / f"{concept.debug_string}.{i}.score_{score}",
                 )
 
+        logging.info(
+            "Logging %s lexicon items to %s",
+            len(self._concept_to_hypotheses_and_scores),
+            log_output_path,
+        )
+        for (
+            surface_template,
+            meaning_as_perception_graph_template,
+        ) in self._lexicon.items():
+            meaning_as_perception_graph_template.render_to_file(
+                surface_template.to_short_string(),
+                log_output_path / f"lexicon-{surface_template.to_short_string()}",
+            )
+
 
 def _extract_candidate_attributes(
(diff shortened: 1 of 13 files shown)


spigo900:
@LichMaster98 This seems to be working now. Some caveats:

1. The candidate attribute extraction code heavily restricts the search space down to only hypotheses that look like `object HAS_PROPERTY property`. (Attributes that cover multiple properties are not included in the space of possible hypotheses.)
2. Despite this it still seems pretty fragile, which makes sense as every object should have at least two "noise" attributes (`animate`/`inanimate` and `biggerThan`/`smallerThan`/`aboutTheSameSizeAsLearner`).
3. I haven't tested it except on some very simple and specific situations. I will see about setting up something more realistic tomorrow.

spigo900:
@LichMaster98 I set up a more realistic experiment and the pursuit attribute learner seems to handle it fine. The experiment runs the pursuit attribute learner (using an object recognizer) on the `m9-attributes` curriculum. The experiment runs without any runtime errors, and the pursuit learner learns four of the ~seven colors (green, white, light brown, dark brown) correctly, so this seems to be working. Results:

- It is still fragile, only achieving about 40% accuracy overall.
- One quirk is that it outputs "the my ground" and "the your ground" because it incorrectly learns that "my" and "your" just mean "inanimate."
- It also learns that red and black mean "inanimate," though the language there is fine after I fixed "the red the ground." (It's now generating "the red ground.")

I"ll look into a hack to get rid of the "the your ground" quirk.

spigo900:
@LichMaster98 @denizbeser This seems to be working as it "should."

- Does this candidate extraction method make sense? Is it trimming things too much?
- Separately, is it a problem that the learner is learning (and lexicalizing) the wrong meanings for "red," "black, "your" and "my"?

adam/learner/attributes.py
```diff
@@ -297,3 +304,192 @@ def _update_hypothesis(
             ),
             trim_after_match=pattern_remove_incomplete_region_or_spatial_path,
         )
+
+
+@attrs
+class PursuitAttributeLearnerNew(  # should it be called new? there is no old one as far as I know
```
LichMaster98, line 310: I'd suggest leaving it called new given the rest of the current learners we are using are called 'new'. Before we release the software if we have time we should probably clean this up.

adam/learner/attributes.py
```diff
@@ -297,3 +304,192 @@ def _update_hypothesis(
...
+        immutableset([object_with_attribute, property]) for property in properties
+    ]
+    return immutableset(
+        [
+            whole_scene_perception_graph.subgraph_by_nodes(
+                candidate_attribute_subgraph_nodes
+            )
+            for candidate_attribute_subgraph_nodes in candidate_attribute_subgraph_node_sets
+        ]
+    )
```
LichMaster98, line 495: This is the section of code we need to be sure to fully disclaim in any writeup which uses this learner. We know that this is a 'hack' to substantially cut down on the search space.

LichMaster98 (commented):
I've left a few comments on here mostly for future consideration when writing the write-ups of experiments using this code.

adam/learner/attributes.py
spigo900: @LichMaster98 So just to confirm, do we want to keep this hack then?

LichMaster98: Yes @spigo900 otherwise the search space for pursuit could grow too large.

spigo900: @LichMaster98 Definitely, just thought I'd confirm. Thanks!

LichMaster98 (approved):

PR #996, opened 2020-11-16
Add Pursuit Relations Learner

### LINKED ISSUE #990: Implement TemplateLearner version of pursuit relation learner

### LINKED ISSUE #997: Learners attempt to match each template only once
Our template learners (excepting for the object learner in #987) only attempt to match each template once. This means we get at most one semantic node (and one description) per template.

We need to fix this so that we can handle things like the relation "on." In the scene "a ball on a table" it's implicit that the table is on the ground, so the learner might also match its `slot1_on_slot2` template with "a table on the ground" instead of "a ball on a table." Ideally we'd like the relation learner to be able to identify both of these relations, but currently it can only identify one or the other.

This might also be needed for experiments with attribute/relation noise.

## Solutions?
We don't have a general fix for this yet. @denizbeser and I fixed this for the object learner in #987, however the fix depends on the fact that the object learner replaces its matches. Other learners don't and shouldn't do this, because they would remove information later learners need. For example, the relation learner might identify the relation "a dog on the ground" and replace the dog, which prevents the action learner from later recognizing "a dog eats a cookie."

One idea is to simply match the templates more than once. The problem with this is that we will may end up with extra matches (and so semantic nodes) for the "same things" (attribute, relation, verb) where the differences are trivial. For example, we might get extra matches for a verb template where the only difference is how we matched up the predicates for the world's horizontal axes differently. These extra matches won't get thrown away when we take the set of `immutable_new_nodes` because attribute, relation, and action semantic nodes are `eq=False`. And because they won't get thrown away, they might slow down learners later in the learning order.

We could fix the problems with that idea by changing the semantic nodes to be `eq=True`. I can only see two minor problems with that:

1. Unless we add some information to these nodes, we won't be able to recognize multiple distinct instances of the same action in the same scene (if, for example, we had a situation where Mom sits down, then gets up, then sits down again).
    - However, we currently don't have any such situations and aren't likely to need them in the time we have left, so I think this is fine.
2. It might slow things down since we will be looking for every match, not just the first.
    - I think this is probably fine as well.

LichMaster98:
Closes #990, Closes #997

### DIFF
--- a/adam/learner/relations.py
+++ b/adam/learner/relations.py
@@ -1,16 +1,38 @@
-from abc import ABC
+import logging
+from adam.utils.networkx_utils import digraph_with_nodes_sorted_by, subgraph
 
-from typing import AbstractSet, Optional, Iterable, Tuple
+from attr.validators import instance_of, optional
+from pathlib import Path
+from networkx import all_shortest_paths
+
+from abc import ABC
+from adam.learner.pursuit import AbstractPursuitLearnerNew
+
+from typing import AbstractSet, Optional, Iterable, Tuple, Mapping, Sequence
 import itertools
-from adam.learner import LanguagePerceptionSemanticAlignment, PerceptionSemanticAlignment
+from adam.learner import (
+    LanguagePerceptionSemanticAlignment,
+    PerceptionSemanticAlignment,
+    get_largest_matching_pattern,
+)
 from adam.learner.perception_graph_template import PerceptionGraphTemplate
 from adam.learner.subset import AbstractTemplateSubsetLearnerNew
-from adam.learner.surface_templates import SurfaceTemplateBoundToSemanticNodes
-from attr import attrs
+from adam.learner.surface_templates import (
+    SurfaceTemplateBoundToSemanticNodes,
+    SLOT1,
+    SLOT2,
+)
+from attr import attrs, attrib
 from adam.learner.template_learner import AbstractTemplateLearnerNew
-from adam.perception import MatchMode
-from adam.perception.perception_graph import PerceptionGraph
-from adam.semantics import RelationConcept, SemanticNode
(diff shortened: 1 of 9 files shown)


adam/learner/relations.py
```diff
@@ -133,3 +155,218 @@ def _update_hypothesis(
...
+            surface_template,
+            meaning_as_perception_graph_template,
+        ) in self._lexicon.items():
+            meaning_as_perception_graph_template.render_to_file(
+                surface_template.to_short_string(),
+                log_output_path / f"lexicon-{surface_template.to_short_string()}",
+            )
+
+
+class ImmutableSet(object):
```
spigo900, line 322: Why is this here?

adam/learner/relations.py
```diff
@@ -133,3 +155,218 @@ def _update_hypothesis(
...
+    # along the shortest path between the two objects involved in the perception graph.
+    for hypothesis_spine_nodes in all_shortest_paths(
+        perception_graph_undirected, relation_object_2, relation_object_1
+    ):
+        # Along the core of our hypothesis we also want to collect the predecessors and successors
+        hypothesis_nodes_mutable = []
+        for node in hypothesis_spine_nodes:
+            if node not in {relation_object_1, relation_object_2}:
+                for successor in perception_digraph.successors(node):
+                    if not isinstance(successor, ObjectPerception):
```
spigo900, line 352: Do we also want to exclude `ObjectSemanticNode` s along the path and explicitly include the relation objects?

adam/learner/relations.py
LichMaster98: Oh, pycharm import mistake with the `alt-enter` action

spigo900 (approved):
lgtm

PR #1005, opened 2020-12-10
User-facing documentation

### LINKED ISSUE #1001: User-facing documentation on how to use our work
@marjorief mentioned thinking about things we still need to document for potential open-source users of ADAM.

Assuming these potential users want information on how to extend our work (without changing much of the internals, like how perceptions are generated), I think we still need documentation for the following things. Note that a caret (^) indicates "bonus content" -- things that would be nice to have but seem nontrivial to document and not strictly necessary (I think) for ADAM to be useful.

- Assumptions of our framework?
  - (I think these have already been written up somewhere. I am not sure where they are/whether they are somewhere that potential users would see them.)
- Difference between object learner and object recognizer?
  - (This might already be documented somewhere visible. If not, it's worth pointing out for new users. The purpose of the object recognizer is to experiment with non-object learners under the assumption of perfect object recognition, whether for experiment speed or for isolation-of-faults.)
- Extending the existing framework with
  - a new object.
  - a new property.
  - a new relation.
  - a new action.
  - a new curriculum.
  - ^a new learner.
    - ^Background for implementing new algorithms: the general assumptions our template learner framework makes
        - (Somewhat more detailed than the general assumptions. May also be written up somewhere already.)
        - (I think that a proper tutorial is impossible given there are lots of different algorithms you might want to implement, and implementing e.g. subset is very different from pursuit which is very different from the cross-situational learner. So this would mostly be general assumptions that are implicit in the design so that you can figure out how to adapt your algorithm to our framework. The API documentation should be able to fill in the details.)
    - ^ Implementing new variants of existing learners (ex: pursuit for actions, cross-situational for relations, propose-but-verify for attributes)
      - ^one for each algorithm? subset, pursuit, cross-situational, propose-but-verify
      - (I think it would make sense to include pointers to the different object learner implementations, but not to go into detail about each one.)
  - ^a new language
    - (Would need to consult with @paynesa on this as I haven't worked with the language-specific code enough to be confident I know all the steps in defining a new language.)
    - (This might be messy to document given the determiner hacks and other language-specific stuff spread throughout the codebase. Those things would need to be updated if a user wanted to add a new language.)
- Examples for almost every extension
  - None for adding a completely new learning algorithm
- Running your experiment
  - Changing `log_experiment.py` to include your new curriculum (if applicable)
  - Pointer to Jacob's excellent documentation on the various experiment parameters

I think it should take me a few days at most to produce the documentation without "adding a new learner," without "adding a new language," and without examples. I think it should take a little over a week to produce examples that go along with each non-bonus section, assuming example generation goes poorly. I expect adding an object, adding a new curriculum, and "running your experiment" to be the simplest three to generate examples for, while (of the non-bonus content) adding a property, relation or action might be more complicated.

spigo900:
Closes #1001.


spigo900:
I added a mention of the tutorials in the README, though I'm not sure if they will actually show up in the Read the Docs build after this is merged. It [looks like](https://readthedocs.org/projects/adam-language-learner/builds/12560491/) Read the Docs is repeatedly trying to install `six` (as a several-steps-removed dependency of pylint) and failing because it's already installed. It looks like this has been happening for about 2 months. I'll look into the issue and see if I can figure out why that's happening.

spigo900:
I've now checked and fixed the instructions in "Running experiments" and "Adding a curriculum." I checked "Advanced curricula" as well, and the curriculum part seems to work, but if I run an experiment on it, the action learner runs into alignment errors. For now I'm going to move on to checking the other curricula but this seemed worth noting.

spigo900:
As an update, I've now tested all of the tutorials' instructions/code. Everything seems to work, except for two strange unit test failures. These failures seem to be spurious however and unrelated to adding the toy block object (which is what triggered them).

The failures are

1. describing "your box" simply as "a box" in `test_your_attribute_learner`, where the problem is probably that the new object changed the exact curriculum to something that breaks the subset learner, and
2. failing to correctly describe "a chicken puts a block in a car" in `test_put`, where it seems to be failing to recognize the block for some reason.

LichMaster98:
@spigo900 I don't have an immediate thought on the first failure you're seeing however for the second instance I know that the unit tests use the object recognizer which is built from a specific list of Phase1 objects which I assume you didn't add the box to?

spigo900:
@LichMaster98 You're correct, I did not. I didn't realize the list existed. Thanks for pointing that out; I'll add it to the tutorial.

spigo900:
@LichMaster98 To confirm, that did fix the problem. I also had to change the ontology node's debug string to match the English I wanted the recognizer to generate; that's going in the tutorial as well.

spigo900:
I confirmed the cause of the first problem. It is spurious and happens because adding the object changed the exact curricula used.

The learner is seeing "your cup," "your hat," "your beef," "your cookie," and "your block" in training, and all of those objects are smaller than the learner. A box is bigger than the learner. So subset doesn't get to subset away the `smallerThan` property so it can't match `your SLOT1` to the box.

LichMaster98 (approved):

PR #1020, opened 2021-08-20
Remove old learner code which is no longer used

lichtefeld:


adam/experiment/log_experiment.py
```diff
@@ -51,17 +51,13 @@
     integrated_experiment_language_generator,
 )
 from adam.language_specific.english import ENGLISH_DETERMINERS
-from adam.learner.attributes import (
-    SubsetAttributeLearner,
-    SubsetAttributeLearnerNew,
-    PursuitAttributeLearnerNew,
-)
+from adam.learner.attributes import SubsetAttributeLearnerNew, PursuitAttributeLearnerNew
```
denizbeser, line 54: Shouldn't we also refactor the learners so their names no longer say "New"?

lichtefeld: That's a good point. I can add that into this refactor.

denizbeser (approved):

PR #1037, opened 2021-09-20
Bring precommit checks to Angular App

lichtefeld:
This PR brings the prettier and angular linting checks to the angular app.

Travis should also run the checks, will verify once tests are complete.


gayatri1205:
Travis's build failed too. Looking into this to check what caused the issue

lichtefeld:
It seems like Travis doesn't like the `*.md` wildcard operator while my local machine handles that fine. I can just list every file there which is a `.md` file...

lichtefeld:
Travis currently isn't running automated checks on the angular portion of the code. I'll look into having it do so but on a lower priority than some other tasks.

danielnapierski (approved):
All these changes look good.

Issue #1022, opened 2021-09-03
Define expected features from Perception Modules


lichtefeld:
Create new types of perception graph nodes in our perception graph to account for new types of visual features.

lichtefeld:
In looking to implement these further considerations:

For color do we want to work with categorical space where the perception component just gives a color or work with the continuous space of RGB values and deal with the 'similarity' problem within the graph structure? - My preference is categorical because that matches P1/2 and as attributes are not explicitly a target of learned language categorical simplifies the matching process to "do the categories match" rather than us dealing with "is this new RGB value close enough to previous ones".

Texture / Material & Rough Shape also both seem like categorical features.

Distances between objects will become continuous -- Contact is a boolean.

Temporal pace is continuous --This should probably strictly be time as frame rate is camera dependent. The traversal path is probably a mix of some continuous and some boolean descriptions?

lichtefeld:
@blakeharrison-ai discussed "Optical Flow" being a potential feature to pass through for between two still images in the flip-book view to help capture motion. We'd need to determine how to pass and deal with this feature within ADAM's graph representation.

lichtefeld:
@shengcheng -- As I believe you're the one who's going to be responsible for generating the output of Perception would you mind proposing an output format you believe is reasonable for the information requested above?

My preference is either a YAML or JSON format (YAML if we'd prefer the output to be more human-readable) but the exact layout of the keys I think should come from you as you'll be generating these outputs. If you have any more questions for @denizbeser or myself before proposing a format feel free to ask here so we can track the discussion!

PR #1041, opened 2021-09-23
Add new nodes and predicates for perception graphs

lichtefeld:
Needed for perception features as discussed in #1022

### DIFF
--- a/adam/perception/perception_graph.py
+++ b/adam/perception/perception_graph.py
@@ -56,7 +56,6 @@ from adam.ontology.phase1_ontology import (
 from adam.ontology import IS_SPEAKER, IS_ADDRESSEE
 from adam.ontology.phase1_spatial_relations import (
     Direction,
-    Distance,
     PathOperator,
     Region,
     SpatialPath,
@@ -73,6 +72,31 @@ from adam.perception.developmental_primitive_perception import (
 from adam.perception.high_level_semantics_situation_to_developmental_primitive_perception import (
     HighLevelSemanticsSituationToDevelopmentalPrimitivePerceptionGenerator,
 )
+from adam.perception.perception_graph_nodes import (
+    PerceptionGraphNode,
+    UnwrappedPerceptionGraphNode,
+    ObjectNode,
+    CategoricalNode,
+    ContinuousNode,
+    RgbColorNode,
+)
+from adam.perception.perception_graph_predicates import (
+    NodePredicate,
+    GeonPredicate,
+    CrossSectionPredicate,
+    RegionPredicate,
+    AxisPredicate,
+    AnyObjectPerception,
+    IsOntologyNodePredicate,
+    IsColorNodePredicate,
+    ObjectSemanticNodePerceptionPredicate,
+    IsPathPredicate,
+    PathOperatorPredicate,
+    AnyObjectPredicate,
+    CategoricalPredicate,
+    ContinuousPredicate,
+    RgbColorPredicate,
+)
 from adam.random_utils import RandomChooser
 from adam.relation import Relation
 from adam.semantics import ObjectSemanticNode
@@ -81,7 +105,7 @@ from adam.situation.high_level_semantics_situation import HighLevelSemanticsSitu
 from adam.utilities import sign
 from adam.utils.networkx_utils import copy_digraph, digraph_with_nodes_sorted_by, subgraph
 from attr import attrib, attrs
-from attr.validators import deep_iterable, instance_of, optional
+from attr.validators import deep_iterable, instance_of
 from immutablecollections import (
     ImmutableDict,
     ImmutableSet,
@@ -93,7 +117,6 @@ from immutablecollections import (
 from immutablecollections.converter_utils import (
     _to_immutabledict,
     _to_immutableset,
-    _to_tuple,
 )
 from vistautils.misc_utils import str_list_limited
 from vistautils.preconditions import check_arg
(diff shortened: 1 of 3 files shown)


adam/perception/perception_graph_nodes.py
```diff
@@ -0,0 +1,86 @@
...
+    PathOperator,
+    Distance,
+)
+from adam.perception import ObjectPerception
+from adam.semantics import ObjectSemanticNode
+
+
+# Perception graph nodes
+class GraphNode(ABC):
+    """Super-class for all perception graph nodes, useful for types."""
```
denizbeser, line 22: It could be useful to include optional X, Y information here. Although we won't necessarily use that for patterns, it would be useful when passed on to the UI later.

adam/perception/perception_graph_nodes.py
```diff
@@ -0,0 +1,86 @@
...
+
```
denizbeser, line 23: Actually, also include a weight field of course

adam/perception/perception_graph_predicates.py
```diff
@@ -0,0 +1,644 @@
...
+        return isinstance(other, CategoricalPredicate)
+
+    def matches_predicate(self, predicate_node: NodePredicate) -> bool:
+        if isinstance(predicate_node, CategoricalPredicate):
+            return predicate_node.value == self.value
+        return False
+
+
+@attrs(frozen=True, slots=True, eq=False)
+class ContinuousPredicate(NodePredicate):
```
denizbeser, line 136: Not sure what the distinction between Categorical and Continuous Node Predicates is. Does this have to do with continuous representations/weights?

adam/perception/perception_graph_predicates.py
```diff
@@ -0,0 +1,644 @@
...
+            if self.aligned_to_gravitational != predicate_node.aligned_to_gravitational:
+                return False
+            else:
+                return True
+        else:
+            return False
+
+
+@attrs(frozen=True, slots=True, eq=False)
+class GeonPredicate(NodePredicate):
```
denizbeser, line 351: Most of the code like this class is moved from perception_graph.py, right?

adam/perception/perception_graph_predicates.py
lichtefeld: Yes that's correct. For any feature which is a continuous value we would use the continuous node but for categorical values (e.g. material) we'd use the category node. I'm unsure we currently plan to have a continuous feature but it was easy enough to add now just so its available in the future.

adam/perception/perception_graph_predicates.py
lichtefeld: Correct, I just relocated these predicates into the new file.

adam/perception/perception_graph_nodes.py
lichtefeld: @denizbeser -- Would you think storing one (X,Y) pair as the center of the feature of interest is reasonable enough for now? We may need this to become more robust in the future but I think future robustness is going to be a question of what information we can get about where the features are located from @shengcheng ?

denizbeser: Yes, I think so, center should be enough!

adam/perception/perception_graph_predicates.py
denizbeser: I see. Hmm, it seems that all of our features (and corresponding predicates) would be continuous though, no? I thought even something "categorical" would actually have a weight associated with it (like in the weight field from the comment above)

lichtefeld: We've discussed binning color by just arbitrarily picking the breakpoints in RGB space rather than trying to explore the problem of "defining what 'red' is in RGB" and dealing with different lighting conditions, etc. So in that case color would become a categorical definition rather than continuous correct? For an example that's not color I suspect the feature of "material" will come to use as a set of discrete categories which we may end up referring to as 0, 1, 2, etc rather than by a human name but the pertained system probably only has a finite list of materials it can categorize to.

Yes I agree these nodes will also have a weight, I was under the impression this "weight" could be the confidence of the perception system that this feature is accurately detected or alternatively in the graph patterns the weight of importance for "if we are missing these feature can we still say this object matches this pattern" E.g. if an apple is matte rather than shiny is it still an Apple ? but we don't loose the information that maybe this was an outlier and actually most apples are shiny.

denizbeser: Don't we have a specific predicate for colors though? Either way, these will be specified more as we go. I was just confused about what `ContinuousNodePredicate` was supposed to capture

denizbeser (approved):

adam/perception/perception_graph_predicates.py
```diff
@@ -0,0 +1,644 @@
...
+        if isinstance(other, RegionPredicate):
+            if self.distance is None and other.distance is None:
+                return True
+            elif self.distance is not None and other.distance is not None:
+                return self.distance.name == other.distance.name
+        return False
+
+    def matches_predicate(self, predicate_node: "NodePredicate") -> bool:
+        if isinstance(predicate_node, RegionPredicate):
+            return predicate_node.distance == self.distance
```
danielnapierski, line 493: readability note, self. was listed as first argument in most comparisons in this file, but here predicate_node. is first.

danielnapierski (approved):
Looks good.  I made a few notes.

adam/perception/perception_graph.py
```diff
@@ -73,6 +73,31 @@
...
+    IsOntologyNodePredicate,
+    IsColorNodePredicate,
+    ObjectSemanticNodePerceptionPredicate,
+    IsPathPredicate,
+    PathOperatorPredicate,
+    AnyObjectPredicate,
+    CategoricalPredicate,
+    ContinuousPredicate,
+    RgbColorPredicate,
+)
```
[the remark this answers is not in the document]
lichtefeld: Agreed, for some reason I recall the auto sorting of imports on this repository not working. If I get some free time to address this I will but otherwise I feel inclined to not worry about this for the moment.

Issue #1029, opened 2021-09-14
Learner Experiment Output


lichtefeld:
Currently when running an experiment we output a sequential HTML file with the results of an experiment -- This format will not suffice for the new viewer. With this in mind here is the propose directory layout along with the format for the output of ADAM's learning results.

### Experiment Directory Layout
This directory stores the entire results from running an experiment with a given learner and curriculum (both defined in the `config.yaml` file. Two additional directories exist to store the outputs of the learner and the impacts on a per-scene basis.
```
- config.yaml
- scenes
  - scene_1
    - description.yaml
    - perception.yaml
    - pre_learn_decode.yaml
    - post_learn_decode.yaml
  - scene_2
- learner
  - pickled_states # This says the same layout as currently used
    - 50_scenes_learner.pickle
    - ...
  - patterns # This stays the same layout as currently used -- These are mostly debugging ?
    - 50
      - objects
      - actions
    - ...
```

#### config.yaml
```yaml
curriculum: /path/to/curriculum/dir
start_scene_img: /path/to/scene/image.png
end_scene_img: /path/to/scene/end.png
learner_config:
  object_learner: 
    object_learner_type: "subset" # This can also be "pursuit"
    # Other learner configurations by learner type go here
  action_learner:
    # Same layout as the object learner type. This is really a mirror of the configuration files for the integrated learner loading
    # We just want to dump the full configuration here so it can be displayed
```

### Scenes
This directory stores the information for each scene being a directory containing the following files:

#### description.yaml
This yaml file provides hooks needed to map an object segment (or the collection of strokes from @shengcheng ) from the perception to a location in the 2D object so we can track the gold-standard information for both automatic analysis and display. Exact format is TBD given the information it may need to provide.

#### perception.yaml
This stores the presentation of the scene as extracted by ASU as the input to ADAM. Format will be discussed in #1022 

#### *_decode.yaml
This is the output of the learner for the given scene. The pre_learning and post_learning tags are useful for comparing the pre and post decode when a new language input is presented to the learner. This can be useful for comparing how new inputs change the patterns of objects & actions. This file should be designed to help the viewer so the language output is a triplet of the output text, a confidence, and the collection of features to describe the object. Additionally we need a way to map this language to a portion of the scene which is probably a map to a cluster in 3D space which can be translated into a square outline on the scene image.

lichtefeld:
Proposed output for the `*_decode.yaml` file:

```yaml
scene_num: 1
output_language: 
[...]

PR #1043, opened 2021-09-28
Data Loader for UI

denizbeser:
sample learner output directory included. #1029

### DIFF
--- a/adam/experiment/experiment_data_loader.py
+++ b/adam/experiment/experiment_data_loader.py
@@ -1,0 +1,80 @@
+import os
+from pathlib import Path
+from typing import List
+import yaml
+
+
+def get_train_curricula(output_root: Path) -> List[str]:
+    r"""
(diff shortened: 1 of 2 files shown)


lichtefeld (approved):
@denizbeser This code looks great thanks. @gayatri1205 -- I'm inclined to merge this into master and I'll get the flask integration setup next week for you. Does that plan work for you? That way you can consider how the UI can handle the list of output language using this test example.

For now just using the original scene image from Blake is a good example image still as there is a table in the image. Not that the working example matters for if all the information makes sense as we just want to evaluate the layout and make changes as needed.

gayatri1205:
This looks good to me! I'll work on making the UI receive this output and displaying it in a separate result component. You can get me the flask set up in your own time. Thank you so much! :)

gayatri1205 (approved):

PR #1044, opened 2021-10-07
Add Flask App

### LINKED ISSUE #1040: Minimum Viable Flask API Endpoints
Below lists the minimum Flask API endpoints as needed by the UI along with any expected input values and the output format. 
These are all `GET` requests as we don't currently have a need to `POST` data.

This issue references data formats as discussed in #1026, #1027, and #1029.

**/api/learners**
Output: 
A list of dicts which for now only have the "name" of the learner. This is future proofing for easier addition of more learner details to display if needed.
```json
{
  "learner_types": [
    {
       "name": "integrated_complete",
    },
  ] 
}
```

**/api/training_curriculum**
Output:
The name of the available training data sets in a list.
```json
{
  "training_curriculum": [
    "objects_single",
    "objects_two_in_scene",
    "etc"
  ] 
}
```

**/api/testing_curriculum**
Same as `api/training_curriculum` except returns the list of test curriculums.

**/api/load_scene**
Input:
```json
{
  "curriculum_name": "objects_test",
  "scene_number": 1
}
```
Comments:
I think this API endpoint should be able to load a scene from the training set and by default we'd probably want the display to show the information after loading. So I think we should, by convention, use the `post_learning` field for the decode on a the test curriculum. Thoughts? I think the format in these fields can just be the output of the `output_language` in this [yaml description](https://github.com/isi-vista/adam/issues/1029#issuecomment-921155932).

Output:
The entire collection of data needed to display a scene on the UI. Including the scene images, the gold standard information, and the learner's output.
```json
{
  "curriculum_name": "objects_test",
  "scene_number": 1,
  "scene_images": [
    "/path/to/scene/img.png",
    "/path/to/scene/img2.png",
  ],
  "post_learning": {},
  "pre_learning": {},
  "gold_standard": {
    "language_with_scene": "a chair",
  }
}
```
--
@gayatri1205 How does this match your current expectations of available end points? Do these endpoints provided the data you're expecting

lichtefeld:
Closes #1040 -- This PR updates the dependencies of ADAM and introduces a Flask App for delivering data to an angular front end component.

### DIFF
--- a/adam/flask_backend/app.py
+++ b/adam/flask_backend/app.py
@@ -1,0 +1,108 @@
+from glob import glob
+from http import HTTPStatus
+from typing import Any
+
+import yaml
+from flask import Flask, abort, request
+from flask_cors import CORS
+
+from adam.paths import (
+    DATA_DIR,
+    LEARNERS_DIR,
+    SITUATION_DIR_NAME,
+    POST_LEARN_FILE_NAME,
+    EXPERIMENTS_DIR_NAME,
+    EXPERIMENTS_TESTING_DIR_NAME,
+)
+from adam.experiment.experiment_data_loader import (
+    get_learners,
+    get_train_curricula,
+    get_test_curricula,
+)
+
+app = Flask(__name__)
+cors = CORS(app)
+app.config["CORS_HEADERS"] = "Content-Type"
+app.config["JSON_SORT_KEYS"] = False
+
+
+@app.route("/api/learners", methods=["GET"])
+def get_all_learners() -> Any:
+    """Get all learner configurations which are prepared."""
+    learners = get_learners(DATA_DIR)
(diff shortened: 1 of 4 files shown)


adam/flask_backend/app.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@app.route("/api/training_curriculum", methods=["GET"])
+def get_all_train_curriculum() -> Any:
+    """Get all possible training curriculum."""
+    train_cur = get_train_curricula(DATA_DIR)
+    return {"training_curriculum": train_cur}
+
+
+@app.route("/api/training_curriculum", methods=["GET"])
```
gayatri1205, line 43: Is this a duplicate route? There already seems to be a route for training_curriculum, should this be testing_curriculum?

Issue #1045, opened 2021-10-08
Observer to generate P3 output format


lichtefeld:
in the [observers.py](https://github.com/isi-vista/adam/blob/master/adam/experiment/observer.py) we need to implement a new observer to generate the output format discussed in #1029. This will be a new `DescriptionObserver` which writes out a file in the `observe` method. This file write would be the decode yaml discussed in #1029.

Additional tools may be needed to easily extract the list of features which correspond to the linguistic token. This list will need to be extracted from a pattern graph in a way that makes logical sense to read and can be linked back to the source image where applicable. 

This observer should be designed to be scene-independent and shouldn't need to track any internal state from scene-to-scene. This will reduce the chance of #1018 appearing in this new implementation. 

As a sanity check against #1018 we should not overwrite an existing decode file if one exists and instead just log a warning that we attempted to overwrite a decode file.

PR #1046, opened 2021-10-12
Initial P3 Observer

lichtefeld:
Works on #1045 

This is not a complete implementation of the desired output of the P3 Observer as the information to extract features correlated with object clusters in the scene is not currently available to the observer. I believe the simplest change at this point will be to retrieve this information from the perceptual representation created from the visual sensors & processing output. Once this format has been established I'll resume extracting the feature information for this observer.

Going to consider the expansion of this observer a combined experiment PR after all individual portions of the changes have been merged.

### DIFF
--- a/adam/experiment/observer.py
+++ b/adam/experiment/observer.py
@@ -1,7 +1,9 @@
 import logging
+import yaml
+
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Generic, Mapping, Optional, Tuple
+from typing import Any, Generic, Mapping, MutableMapping, Optional, Tuple
 
 from attr import attrib, attrs
 from attr.validators import instance_of, optional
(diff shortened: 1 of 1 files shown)


adam/experiment/observer.py
```diff
@@ -751,6 +754,50 @@ def report(self) -> None:
...
+        self,
+        situation: Optional[SituationT],
+        true_description: LinguisticDescription,
+        perceptual_representation: PerceptualRepresentation[PerceptionT],
+        predicted_descriptions: Mapping[LinguisticDescription, float],
+    ) -> None:
+        output_dict: MutableMapping[str, Any] = dict()
+        output_dict["gold_language"] = true_description.as_token_string()
+        output_dict["output_language"] = []
+
```
danielnapierski, line 773: Is it worth defining strings for `output_language` and `gold_language` that are reused?

danielnapierski (approved):
I only had a couple of small notes/questions.

adam/experiment/observer.py
lichtefeld: Yes this makes sense.

gayatri1205 (approved):

PR #1052, opened 2021-10-26
Load perception frames and enable conversion to PerceptionGraphs

### LINKED ISSUE #1030: Translate ASU's visual representation graph to our internal perception graph representation
Need to write a translation method, similar to `from_frame()` un `perception_graph.py` line 345. It would take in the ASU representation and return a PerceptionGraph
 https://github.com/isi-vista/adam/blob/91f392f2529a98cd50c095a18769ae4b55ce4292/adam/perception/perception_graph.py#L345
 
Newt things to keep in mind/include:
- New node types as we progress.
- Source process (i.e the feature extractor used for marking this feature)
- Weight of the detection, i.e continuous features, will also be used for confidence.

lichtefeld:
Closes #1030 

This PR involves two reasonable sized refactors the first to how the IntegratedLearner is derived by inheritance so that different IntegratedLearners can share the same base code and only need to implement specific methods to deal with handling the different type of frames. (Frames being what holds the perceptual description from a system before it is converted into a PerceptionGraph). This refactor converted the `IntegratedLearner` into a `SymbolicIntegratedLearner` which passes unit tests as a replacement for the original.

The second refactor is with the `_FrameTranslation` class which provides utility functions to convert a `PerceptionFrame` into a `PerceptionGraph`. This was refactored to share code and ensure future `_FrameTranslation` classes include the appropriate shared functions. 

Finally the `VisualPerceptionFrame` and the requisite `_VisualFrameTranslation` class were implemented to convert perception output into the perception graph. Two simple unit tests are provided one of which tests that the JSON loader of the perception frame does so without error. The second ensures a `VisualPerceptionFrame` can be translated into a `PerceptionGraph` which when immediately transformed into a `PerceptionGraphPattern` matches against itself (as it should).

More complex tests can come as I have more perception source data to work with and convert into a test -- In most cases more complex tests will derive from experimentation producing an error case we want to debug and ensure we don't reintroduce :)


adam/perception/perception_graph.py
```diff
@@ -33,6 +33,7 @@
     TypeVar,
     Union,
     cast,
+    Generic,
```
danielnapierski, line 36: Consider placing this in the list aphabetically.

adam/perception/perception_graph.py
```diff
@@ -2629,6 +2696,62 @@ def _map_relation(
...
+            "Can only create a DynamicPerceptionGraph from exactly two frames, "
+            "but got %s",
+            (len(perceptual_representation.frames),),
+        )
+
+        raise NotImplementedError(
+            "Multiple frame translation is not yet supported for Visual Perception Frames."
+        )
+
+
```
danielnapierski, line 2754: This is a large file, just calling attention to the length of the file.

danielnapierski (approved):
Lots of content.  A few comments.

PR #1062, opened 2021-11-07
updated parsing method of post-decode file

gayatri1205:
Updated parsing method to handle the case where no sub-objects are present for an object


lichtefeld (commented):
These changes look good to me. To accommodate `make precommit` not working well for @gayatri1205 yet I've run it locally and fixed any concerns. If we wait until after #1060 is merged and then rebase on that branch travis will check this for us automatically. 

Rather than approve this after just fixing precommit changes I'm going to have @gayatri1205 confirm functionally is still intact after my fixes in addition to my validation so that hopefully both of us validating functionality ensures these changes are good to go.

lichtefeld (approved):

Issue #1088, opened 2022-01-10
If object has a color property, display a swatch of that color in the properties list


lichtefeld:
Currently features for ADAM's discovered concepts are detailed in a 'simple' list. It would be ideal for this display to include a small span of ~40-50px long after the line which is a hex code. This span should be colored in the color indicated by the hex. This will allow for a better human understanding of the color attribute.

<img width="1241" alt="Screen Shot 2022-01-10 at 9 35 11 AM" src="https://user-images.githubusercontent.com/8626833/148786394-b95e78e2-39dd-4df7-a34c-cbb381393215.png">

To accomplish this, the method of generating the feature list will need to be updated to allow for some inspection of the string type to identify the color feature. A simple match would be to check for the `#` sign at the start of the feature as this is probably the only color feature we'd be including however a more robust match would be to regex match `#` + 6 hex digits as the only text in the line.

lichtefeld:
Closed by #1091

PR #1091, opened 2022-01-21
If object has a color property, display a swatch of that color in the properties list #1088

gracemcclurg:
I added a boolean attribute `isColor` to the property object which uses regex to determine whether the color swatch should be displayed in HTML on the corresponding line. The color is displayed on a square whose font color matches the property color.

### DIFF
--- a/angular-viewer/adam-angular-demo/src/app/classes/features.ts
+++ b/angular-viewer/adam-angular-demo/src/app/classes/features.ts
@@ -1,3 +1,4 @@
 export class Features {
   name: string;
+  isColor: boolean;
 }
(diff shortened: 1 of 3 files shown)


lichtefeld (changes requested):
It looks like you need to run `make precommit` to run the angular code formatters. I don't have a working script to run this from the root directory (yet) so from `angular-viewer/adam-angular-demo` if you run `make precommit` the angular precommit checks will run. Please fix any errors from this set of tools then I'll rereview. 

Can you also attach a screenshot of the color bar into this issue so we can see it working?

gracemcclurg:
Here is an attached screenshot showing the responding color: image

lichtefeld (approved):
Looks good!

PR #1097, opened 2022-02-08
Update ADAM to Python 3.7

lichtefeld:
This PR updates ADAM to use Python 3.7 now that 3.6 is end of life.


tests/language_specific/english/test_english_language_generator.py
```diff
@@ -1148,108 +1148,90 @@ def test_in_region_as_goal():
...
+            goal_object,
         )
-        == ("a", "dog", "goes", "over", "a", "table")
-    )
+    ) == ("a", "dog", "goes", "over", "a", "table")
 
 
 def test_under_region_as_goal():
     goal_object = situation_object(TABLE)
     # Over
```
danielnapierski, line 1228: 1243 better as "Under"

tests/language_specific/chinese/test_chinese_language_generator.py
```diff
@@ -2202,94 +2200,79 @@ def test_in_region_as_goal():
...
 def test_under_region_as_goal():
     goal_object = situation_object(TABLE)
     # Over
-    assert (
-        generated_tokens(
-            region_as_goal_situation(
-                Region(goal_object, distance=PROXIMAL, direction=GRAVITATIONAL_DOWN),
-                goal_object,
-            )
+    assert generated_tokens(
```
danielnapierski, line 2270: line 2269 "# Under"

requirements_lock.txt
```diff
@@ -1,63 +1,60 @@
...
-charset-normalizer==2.0.6
-click==8.0.1
+black==22.1.0
+certifi==2021.10.8
+charset-normalizer==2.0.11
+click==8.0.3
 click-default-group==1.2.2
 contexttimer==0.3.3
-coverage==6.0.1
-dataclasses==0.8
```
danielnapierski, line 12: was datac;asses unused?

Makefile
```diff
@@ -2,7 +2,7 @@ default:
 	@echo "an explicit target is required"
 
 # easier to test python2 vs. python3
-PYTHON=python3
+PYTHON=pypy3
```
danielnapierski, line 5: Can you comment on the benefits of this change?

danielnapierski (approved):
Looks good.  Added a couple of questions.

Makefile
lichtefeld: We were previously using pypy for tests as it runs the entire test suite in ~10ish minutes on my laptop as compared to ~20+ for regular python. A large number of the tests involve our graph matching algorithm, which while designed to be efficient, is solving an NP-Complete problem of sub-graph isomorphism matching. I'd run into problems getting the local install of pypy working so I had changed this back to python3 so I could still test even if slower but with the update to 3.7 I got pypy working again locally.

requirements_lock.txt
lichtefeld: dataclasses are built-in to python 3.7 so the backwards compatibility package can be dropped.

PR #1109, opened 2022-03-02
Update ID tracking to use ID from graph nodes if one exists

### LINKED ISSUE #1104: Text and Cluster ID Mismatch
The objects' descriptions do not correspond to the appropriate IDs. In the example below, the book, which has ID 0, is unproperly labeled with ID 1 in the description pane.
image

lichtefeld:
Closes #1104 by tracking an original node ID via the detected semantic_node when applicable.

### DIFF
--- a/adam/experiment/observer.py
+++ b/adam/experiment/observer.py
@@ -880,9 +880,9 @@ class YAMLLogger(DescriptionObserver[SituationT, LinguisticDescriptionT, Percept
                     # to ensure the ID is unique, so we assume all semantic nodes in the scene could have an original ID
                     # and 'count' from there. This uniqueness guarantee could be improved, but it's not a problem
                     # to potentially have non-linear unique IDs.
-                    semantic_node.original_node_id
+                    semantic_node.original_node_id  # type: ignore
                     if semantic_node.original_node_id is not None
-                    else idx + len(predicted_scene_description.semantics_to_descriptions),  # type: ignore
+                    else idx + len(predicted_scene_description.semantics_to_descriptions),
                     semantic_node,
                     linguistic_description,
                     predicted_scene_description,
--- a/adam/experiment/observer.py
+++ b/adam/experiment/observer.py
@@ -874,7 +874,15 @@ class YAMLLogger(DescriptionObserver[SituationT, LinguisticDescriptionT, Percept
             semantic_node, linguistic_description = semantic_to_description
             output_dict[OUTPUT_LANGUAGE].append(
                 self._convert_to_output_format(
-                    semantic_node.original_node_id if semantic_node.original_node_id is not None else idx + len(predicted_scene_description.semantics_to_descriptions),  # type: ignore
+                    # If the semantic node has an alignment to an original perception node in the graph (e.g. an object
+                    # perception root) we want to use the original ID value to enable a matching between the produced
+                    # linguistic utterance and the identified object cluster. In the case no original ID exists we want
+                    # to ensure the ID is unique, so we assume all semantic nodes in the scene could have an original ID
+                    # and 'count' from there. This uniqueness guarantee could be improved, but it's not a problem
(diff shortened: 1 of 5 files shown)


gracemcclurg (changes requested):
Great job on the formatting of your code. However, the ID mismatch issue continues to persist on my end when testing the branch. This may be because the ID overlay code was merged between your testing of the code and my review. Please let me know if the code still appears to work on your end and I will look more into my testing. Thanks! : )

gracemcclurg (changes requested):
The ID mismatch issue continues to persist it seems, specifically with scene 7 as an example.

gracemcclurg (approved):
With your changes, it appears in testing on my end that all scenes' IDs in the linguistic description 
are matched appropriately to the correct objects.

adam/experiment/observer.py
```diff
@@ -874,7 +874,7 @@ def observe(  # pylint: disable=unused-argument
             semantic_node, linguistic_description = semantic_to_description
             output_dict[OUTPUT_LANGUAGE].append(
                 self._convert_to_output_format(
-                    idx,
+                    semantic_node.original_node_id if semantic_node.original_node_id is not None else idx + len(predicted_scene_description.semantics_to_descriptions),  # type: ignore
```
spigo900, line 877: This is a complicated enough expression that it probably deserves either (1) a comment or (2) a separate function. Using the original node ID makes sense, but I'm not sure I understand the fallback (at least not from the immediate diff context).

adam/learner/learner_utils.py
```diff
@@ -781,3 +790,31 @@ def get_slot_from_semantic_node(
         if object_node.concept == object_concept:
             return slot_var.name
     return slot
+
+
+def get_root_object_cluster_perception(
+    graph: DiGraph, matched_subgraph_nodes: AbstractSet[PerceptionGraphNode]
+) -> Optional[ObjectClusterNode]:
+    """Get the root object cluster node from a graph given a subset of nodes which represent the object.
```
spigo900, line 798: The function name and description are a bit confusing here, and I'd suggest changing them for clarity. It looks like this function is designed (as it should be) so that it doesn't crash when the subset of nodes is not actually a match for an object pattern match but say an action. So the name and docstring should reflect that.

(Re: the docstring, I think it's fine to leave this as a note below the single-line description since the description is already longish. Your call how to format it though.)

At the moment, the match should always be an object, but seems worth being clear that this should still work in the non-object case.

spigo900 (changes requested):
Looks good, just two suggestions for readability's sake.

adam/learner/learner_utils.py
spigo900: For a name, maybe something like `get_root_if_perception_is_object_cluster`?

adam/experiment/observer.py
spigo900: The new comment looks good to me. Thanks!

adam/learner/learner_utils.py
```diff
@@ -781,3 +790,33 @@ def get_slot_from_semantic_node(
...
             return slot_var.name
     return slot
+
+
+def get_root_if_perception_is_object_cluster(
+    graph: DiGraph, matched_subgraph_nodes: AbstractSet[PerceptionGraphNode]
+) -> Optional[ObjectClusterNode]:
+    """Get the root object cluster node from a graph given a subset of nodes which represent the object.
+
+    If the root object cluster does not exist this function will return none.
```
spigo900, line 800: I think it's worth being explicit here about why the root object cluster might not exist, because otherwise the name and docstring don't obviously make sense together. Something like:

```suggestion
    If a root object cluster node does not exist this function will return none. For example, this happens when the passed nodes represent an action and not an object. It could also happen if the nodes represent two separate objects.
```

spigo900 (changes requested):

PR #1112, opened 2022-03-14
Updated Action Ontology

### LINKED ISSUE #1066: Create M6 Actions Core Curriculum
The M6 Core Curriculum will contain a small set of smilier types of actions the final set needs to be agreed upon but I'd like to suggest: Open, Close, Stack, Take, Put as these actions are an agent interacting with an object with the potential to express relationships between the objects at the start or end of the action. Additionally the 'open' and 'close' action provides a discovery of an affordance difference between `Cube` and `Box`.

Step 1 is making sure we all agree with this curriculum selection then:

- [ ] YAML Specification of M6 Core Curriculum - @lichtefeld - 11/24
- [ ] Creation of Simulated Output for M6 Core Curriculum - @blakeharrison-ai - 12/14
- [ ] Perception Processing over Simulated Output for M6 Core Curriculum - @shengcheng - 12/21 (Or the last day before the Holiday break?)

gracemcclurg:
This pull request closes #1066. Added 6 new actions to Phase 3 Ontology and imported 14 actions from Phase 1 Ontology into Phase 3 Ontology. The new actions are highlighted in blue in the following table.
image


adam/ontology/integrated_learner_experiement_ontology.py
```diff
@@ -116,12 +116,24 @@
 subtype(GLIM, INANIMATE_OBJECT)
 ZUP = OntologyNode(
     "zup",
-    [CAN_FILL_TEMPLATE_SLOT, PERSON_CAN_HAVE, ROLLABLE, RED, INTEGRATED_EXPERIMENT_PROP],
+    [
```
lichtefeld, line 119: Did black reformat this code? If its needed to pass the code-formatter now we can keep this change but otherwise I don't see a reason for this change to occur in a non-related ontology.

adam/ontology/phase3_ontology.py
```diff
@@ -197,6 +273,38 @@
     ),
 )
 
+_NULL_ACTION_DESCRIPTION = ActionDescription(
+    frame=ActionDescriptionFrame({AGENT: NULL, THEME: NULL, GOAL: NULL}),
```
lichtefeld, line 277: Rather than using the `NULL` from `asynic.io` (which isn't applicable here) can you make a `_NULL_ACTION_VARIABLE = ActionDescriptionVariable(NULL_NODE)` to use here?

lichtefeld (changes requested):
One other code organizational note. This is mostly for readability and to mirror `phase1_ontology` but can you move all the new additions for actions below the `GAILA_PHASE_3_STRUCUTRAL_SCHEMATA` dictionary? It will make find parts of the ontology definition easier in the future. Python doesn't care but a human reader will :)

Also your commit is failing CI so please check the travis failure to resolve it as well.

adam/ontology/integrated_learner_experiement_ontology.py
gracemcclurg: It

lichtefeld (approved):
lgtm!

lichtefeld:
Rebased for linear merge into master

PR #1120, opened 2022-04-18
Implement affordance learner

### LINKED ISSUE #1111: Affordance Learner
For Phases 1 & 2 ADAM's learner used proto-semantic roles to determine which objects could fill the `SLOT` of an action pattern. While there was previously some work on extracting this set of proto-roles from video that doesn't align well with Phase 3's goal of feature discovery. While we may still need some notion of these 'proto-roles' to distinguish between the 'agent' and 'theme' of an action I'm setting aside this portion of the feature to look at action affordances. That is the ability of a ball to be rolled (e.g. rollable) when observed in the 'roll' action and the ability to determine what about the object's features enables this affordance. 

The remainder of this issue takes two parts the first explains the high-level learning process I'm currently envisioning for this process (and has been loosely talked about before in presentations/meetings) and the second covers some more technical details of the implementation. 

## Affordance Learning
As described above an affordance for purposes of this learner can be thought of as an ability of an object to participate in the given action. For clarity 'participate' in the action can mean in multiple different semantic roles. E.g. both a person and a ball are 'rollable' if we want to also encode the semantic e.g. "rollable" vs "can rollable" we in theory could do this but I'm unsure immediately how we'd have ADAM discover the semantic difference without some kind of proto-roll information available as a feature. We can't rely on the agent being `slot_0` and the theme `slot_1` as a representation hack either so I'm open to potential ideas here. For the moment I'm going to stick with the idea of learning affordances of the form 'rollable'.

### What does the learner look like
I'm envisioning the affordance learner as a new composable learner as the goal of the affordances is to find a collecting of existing features that enables the affordance. For the example 'rollable' affordance would ideally identify features within the stroke-graph that enable this functionality. This affordance concept maps to a feature node which is a categorical feature with the category "affordance" and the values are an expanding set along all actions observed. 

### How and where in the pipeline will learning occur
Learning affordance, if the learner is enabled, will occur after the action learner has decoded any actions for the scene. The affordance learner will take in the action matched and generate a feature for this action. This feature will be an incremental counter of 'affordance_X' where X is the number of actions seen so far. The affordance for a given action will always map to the same affordance ID. Learning for an affordance will then take the object pattern for the object aligned to a given slot in an action and replace the object cluster with a `SLOT` filler. The description of the object will then be `subset` across all examples of objects which participate in the same slot of the given action. This subsetted pattern can then be match in the original perception graph before action matching occurs to indicate that other objects share this same affordance even if we've never observed a novel object being rolled before.

Ideally the learning of this feature can be propagate into the concept for the given action.

### Where will decode take place
Decode will take place after initial object recognition as we'll need an `ObjectSemanticNode` to fill the `SLOT` created in the affordance pattern with. In practice this will probably occur just before action learning for simplicity. This process will change the perception graph by adding in a new type of node than we've previously added during the process. Rather than just adding a semantic node we'll be adding and node designed as a potential identifying feature. 

### Adding to Integrated Learner
I'm not thrilled to be adding into the integrated learner but I think this is going to be the best route for this module. We need these affordances available within the learning process of other learners. 

### Debug Vs. real affordance IDs
I'm considering implementing the unique value for affordances differently between a debug mode and a 'real' mode. In debug mode the affordance id would be generated in the form of "able to X" where X is the action that triggered the creation of this affordance. While in the 'real' mode the affordance ID would take the form of "affordance_Y" where Y is a running count of the actions seen. In theory this provides a way for replacing "affordance_Y" with the actual name (e.g. 'rollable') in the future. While I'm not using these names as anything other than ID I suspect if I try to present to DARPA with "able to X" as the affordance value they'll complain I'm leaking data into the system even though ADAM doesn't use that value past a string comparison. Therefor sticking with the "affordance_Y" approach for real data would be preferred in demo situations.

### Potential issues
* Pattern graph immutability is likewise impacted by this proposal. It's likely that the solution used in #1110 will get reused here.

### LINKED ISSUE #1114: Categorical nodes don't have a `label` attribute
>> (Categoricals don't yet carry a label.) 
>
>I suppose that's not a bug strictly speaking (so long as all values across all categories are unique) but given the uniqueness guarantee isn't enforced anywhere it seems like a bug waiting to happen. I'll open a new issue to discuss this because there's a reasonable argument for not adding a category label 
>
> _Originally posted by @lichtefeld in https://github.com/isi-vista/adam/issues/1102#issuecomment-1069176872_

This isn't strictly a bug for the reasons I listed above however introducing a `label` attribute would be reasonable for a few reasons:
* Makes category values less likely to clash with values from other categories
* Places the responsibility for uniqueness on the component/sensor/algorithm generating the category values rather than making ADAM need to ensure uniqueness across all category values and raise an error if that is not kept

lichtefeld:
Closes #1111 and Closes #1114

This implements the affordance learner for the simulated integrated learner. While in theory this could be used in the symbolic learner aside from some minor testing I didn't do any verification that this learner will work with the symbolic system.


adam/learner/affordances.py
```diff
@@ -0,0 +1,195 @@
...
+                affordance_property = CategoricalNode(
+                    label="affordance", value=node.concept.debug_string, weight=1.0
+                )
+                digraph.add_node(affordance_property)
+                # This edge might not handle temporal encoding correctly?
+                digraph.add_edge(
+                    list(node.slot_fillings.values())[0],
+                    affordance_property,
+                    label=HAS_PROPERTY_LABEL,
+                )
```
spigo900, line 93: Right, for dynamic graphs we have to wrap this in a temporally scoped edge label, assuming we're continuing with the old representation. This code always uses an unwrapped label which breaks edge label consistency for that approach to dynamic scenes.

I don't see a good way of handling scope while keeping things in scope (ha), so maybe we want to use either (the equivalents of) `BEFORE` if we're being conservative or `ENTIRE_SCENE` otherwise. `BEFORE` seems like a safe bet -- the affordance we're learning has to hold `BEFORE` because otherwise the action couldn't have taken place in the scenes we learned it from. `ENTIRE_SCENE` is stronger but may not make sense for some affordances. Say we have an `edible`/`slot2_slot1_eat_slot2` affordance, in a scene `a person eats an apple` -- the affordance no longer logically applies after the eating. In the symbolic setting the object exists both before and after, if I recall correctly, which means we also can't couple it to the object's scope (though I don't think nodes even have temporal scope?).

I see approaches that might be better but I think they're out of scope. One better approach might be to learn the affordance patterns on the `BEFORE` scope, then match on each temporal scope separately. But that sounds complicated enough to be out of scope. An extension would be to learn patterns for each scope separately, but that seems even more complicated. I think we'd have to deal with chained actions (e.g. "I put the box on the table, then I put the box on the floor") to learn each scope separately, which is definitely out of scope.

On a related note, I think we should create an issue about how to handle temporal scopes for simulated action samples and link to that from here. We can plan to address this piece of code once we know how we will represent the samples as temporally scoped graphs. I think (correct me if I'm wrong) we haven't yet defined how we will translate ASU's output (which might? provide many frames of features, more than 2 or 3) into temporally scoped graphs, which complicates this. I think this will be clearer once we action samples to work with, so ideally we would wait on that.

spigo900 (changes requested):
Overall looks good, but I think I noticed a hole in our plan. I'm posing this as a question:

Can the affordance learner match its patterns to static scenes? I don't think it can as things stand, and this is something we need for contrastive learning, and probably for other things. I see two possible fixes.

One would be to change `match_template` inside `template_learner` so it can handle matching dynamic patterns to non-dynamic graphs. I don't think this will cause problems if all the temporal scopes are the same, but I'm not sure what we want to do for "inherently dynamic" patterns.  Would have to think about that more.

Another would be to revise the affordance learner to learn non-dynamic patterns from dynamic scenes. But this might cause weirdness for dynamic scenes since we default to an `ENTIRE_SCENE` temporal scope on edges for non-dynamic patterns and I suspect this is not appropriate for matching affordances to action scenes.

I think there was a second issue, but I don't remember what it was 🙃 so I'll see if I can remember after lunch. ETA (after lunch): I still don't remember. Probably nothing.

adam/learner/affordances.py
```diff
@@ -0,0 +1,195 @@
...
+            raise ValueError(
+                f"Affordance learner can not extract features when more than one object node is a "
+                f"slot filler. Offending template: {bound_surface_template}"
+            )
+        return immutableset(
+            PerceptionGraphTemplate.from_graph(
+                perception_graph=candidate_attribute_meaning,
+                template_variable_to_matched_object_node=bound_surface_template.slot_to_semantic_node,
+            )
+            for candidate_attribute_meaning in _extract_candidate_attributes(
```
spigo900, line 157: Does this work? It looks like `_extract_candidate_attributes()` looks at properties, but it looks for them by checking the label against `HAS_PROPERTY_LABEL` directly. In a dynamic graph the edge label should be wrapped in a scope, which might break this?

spigo900: Actually, that function also limits itself to single-property hypotheses, which might cause problems for learning affordances. I expect affordances may need to include multiple relationships which may or may not be properties.

lichtefeld: Ah you make a fair point, I was trying to avoid rewriting the logic to stay internal to an object's influence on the perception graph but I _do_ want multiple properties returned as well as the stroke information. I'll just implement an extraction of affordance attributes borrowing from the attribute logic to stay internal to the object.

adam/learner/affordances.py
lichtefeld: So I'd think in general the affordances should be able to be identified regardless of temporal scope, they are essentially 'not temporally determined' (or well shouldn't be) except for when used as a property. (e.g. an 'apple' is only edible before the 'eats' action because afterwards the 'apple' objects ceases to exist (if it still existed it would still be edible). The concern here with adding the node in should probably be safe to use `ENTIRE_SCENE` in our current configuration but I'm fine being conservative with `BEFORE`.  Ideally we'd just want to mirror the temporal scope used for 'object permanence' but I'm not quite sure we represent that.

I'll follow up with Temporal Scope in an issue.

adam/learner/affordances.py
```diff
@@ -0,0 +1,195 @@
...
+        for action_node in semantic_nodes:
+            if isinstance(action_node, ActionSemanticNode):
+                for slot, object_node in action_node.slot_fillings.items():
+                    rtr.append(
+                        SurfaceTemplateBoundToSemanticNodes(
+                            surface_template=SurfaceTemplate(
+                                elements=(slot.name, action_node.concept.debug_string),
+                                language_mode=self._language_mode,
+                            ),
+                            slot_to_semantic_node={SLOT1: object_node},
```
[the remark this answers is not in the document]
lichtefeld: Right, this 'it's not really from the text but also it's a template', should technically be valid. It's a hack around needing a 'surface template' for the learner when probably a learner like the AffordanceLearner (property-learner?) shouldn't need to align to a surface template. I'll add a comment to explain why its being used.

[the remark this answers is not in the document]
spigo900: Right

adam/learner/affordances.py
```diff
@@ -0,0 +1,195 @@
...
+            len(
+                language_perception_semantic_alignment.perception_semantic_alignment.semantic_nodes
+            )
+            > 1
+        )
+
+    def _preprocess_scene(
+        self, perception_semantic_alignment: PerceptionSemanticAlignment
+    ) -> PerceptionSemanticAlignment:
+        return perception_semantic_alignment
```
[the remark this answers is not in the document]
lichtefeld: > Do we we want to change this to filter out all the AffordanceSemanticNode s? If we don't, the affordance learner may learn patterns that refer to its own output, because we enrich with the affordance learner before learning with it.

It's reasonable that we shouldn't be able to learn from the self-affordance node (with the way subset works this shouldn't be possible anyway but I'll see if I can easily add a check for it). It seems reasonable to me that some affordances could be indicated by other affordances but agreed that maybe we just shouldn't include affordance properties in the current iteration. I need to double check if I can make that type of change in the preprocessing or if that will trickle down and remove the semantic nodes from the `perception_semantic_alignment` permanently.

> Separate issue: We probably need to handle temporal scope here to make things work. Namely, we have to learn on dynamic perception graphs because we're learning affordances from observing actions, but I think we want to apply the affordance learner to static perception graphs.

Ooh right, I know the  static -> dynamic is 'solved' because the object learners can handle it but the other direction is less clear. One ugly hack would be to just say that any 'static' scene after the actions learners is just really dynamic so that the edge conversion takes place :) (I don't actually hate this hack because it's still right to say 'ENTIRE_SCENE' temporally for a static instance)

adam/learner/affordances.py
spigo900: Right, I think I agree about not being temporally determined.

I weakly favor `BEFORE` because of `eats` but either scope works fine.

adam/learner/affordances.py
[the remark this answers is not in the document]
spigo900: > I need to double check

Good idea to check, I'm not sure on this one either.

> ugly hack

Not sure I understand. Is the idea here to convert static graphs into dynamic graphs by wrapping every edge in `ENTIRE_SCENE` scopes? I don't hate this idea either, it seems valid.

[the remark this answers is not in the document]
lichtefeld: Correct, just convert a static scene into a dynamic scene (which in effect should wrap every edge with an `ENTIRE_SCENE` temporal scope))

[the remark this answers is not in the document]
spigo900: For the other learners to incorporate affordances, we would have to always do this conversion, right? Or would we do it only under certain circumstances? I didn't follow that part.

This approach seems fine however I suspect this would require some changes to the object learner, which I suspect makes similar assumptions to the attribute learner.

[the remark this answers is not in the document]
lichtefeld: > For the other learners to incorporate affordances, we would have to always do this conversion, right? Or would we do it only under certain circumstances? I didn't follow that part.

We could always make this conversion for the test curriculum, let me thinking about it a little more for the learning curriculum.

lichtefeld:
> Can the affordance learner match its patterns to static scenes? I don't think it can as things stand, and this is something we need for contrastive learning, and probably for other things. I see two possible fixes.
>
> One would be to change match_template inside template_learner so it can handle matching dynamic patterns to non-dynamic graphs. I don't think this will cause problems if all the temporal scopes are the same, but I'm not sure what we want to do for "inherently dynamic" patterns. Would have to think about that more.

I still need to consider this concern more completely. I think the solution will be when going to match the template is converted from temporal scope into a graph without a temporal scope for non-dynamic scenes.

spigo900 (commented):
The review changes look good. I posted one minor thing regarding a comment.

I think we haven't yet resolved the static/dynamic mismatch entirely. Unfortunately I'm not able to reply on those threads from the web UI -- the reply box is gone. We had discussed the solution of converting the graph to a dynamic one before running the affordance learner. This works at test time but causes problems if we do this naively at learning time. Doing this while learning makes the affordance learner happy but causes problems for static learners like the object learner, because they can't learn from dynamic graphs. I think that is where we left the discussion, correct me if I'm wrong. You planned to think more about what to do at learn time?

Is it still part of the plan to modify other learners' patterns? We discussed modifying patterns at some point (in Slack?) but I don't recall the outcome of that. The diff deleting the "add node to pattern" function reminded me of this.

lichtefeld:
> I think we haven't yet resolved the static/dynamic mismatch entirely. Unfortunately I'm not able to reply on those threads from the web UI -- the reply box is gone. We had discussed the solution of converting the graph to a dynamic one before running the affordance learner. This works at test time but causes problems if we do this naively at learning time. Doing this while learning makes the affordance learner happy but causes problems for static learners like the object learner, because they can't learn from dynamic graphs. I think that is where we left the discussion, correct me if I'm wrong. You planned to think more about what to do at learn time?

I think the solution is the patterns should return either temporally scoped or non-temporally scoped based on the perception graph's settings. For now I'm going to naively implement just removing temporal scope from all edges and see if that works. If it doesn't we'll either enforce 'all temporal scopes must be the same' or we'll need some other method of handling it... Hopefully the 'just convert to non-temporal scope' will be sufficient.

> Is it still part of the plan to modify other learners' patterns? We discussed modifying patterns at some point (in Slack?) but I don't recall the outcome of that. The diff deleting the "add node to pattern" function reminded me of this.

Oh right! That's why that utility function was there, I just noticed I didn't end up using it so I removed it. I'll restore it and add that logic in.

adam/learner/affordances.py
```diff
@@ -0,0 +1,225 @@
...
+                # Available to us here and refactoring to have that information is out of scope
+                affordance_property = CategoricalNode(
+                    label="affordance", value=node.concept.debug_string, weight=1.0
+                )
+                perception_graph_after_matching = add_node_connected_to_perception_graph(
+                    perception_graph_after_matching,
+                    list(node.slot_fillings.values())[0],
+                    affordance_property,
+                    temporal_scope=TemporalScope.BEFORE,
+                )
```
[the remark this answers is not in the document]
lichtefeld: After learning (which is where this post-process function runs) we're guaranteed to have a temporal graph because we can only learn from actions.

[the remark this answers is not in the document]
lichtefeld: > This method is called after learning but it's also called at test time. It's called in _enrich_common() which is also called in enrich_during_description(). ETA: So we are guaranteed to have a temporal graph if we assume the translate-to-dynamic hack.

I'll add in the check you suggested. It's probably a stronger guarantee than relying on assumptions.

> Do we assert anywhere that the input graph is dynamic? We probably should given that, if this assumption is violated, things will subtly break.

We do :) -- This is handled in the integrated learner code.

lichtefeld:
@spigo900 Yes that is correct.

adam/learner/perception_graph_template.py
```diff
@@ -222,3 +225,55 @@ def copy_with_temporal_scopes(
...
+        return PerceptionGraphTemplate(
+            graph_pattern=add_predicate_to_pattern_graph(
+                self.graph_pattern,
+                root_node,
+                predicate_node_to_add,
+                temporal_scope=TemporalScope.BEFORE
+                if self.graph_pattern.dynamic
+                else None,
+            )
+        )
```
spigo900, line 279: This method is specific enough to the object learner's affordance processing (the specific scope; the assumption that we're dealing with a single-object pattern) that I think it should live somewhere closer to the learner. It doesn't require any protected access to the template, so moving it out should work fine.

I suggest putting this in `learner_utils.py` near the two other object-related functions, `get_objects_from_perception()` and `candidate_object_hypotheses()`. The name should probably change with the move to (for example) `add_predicate_node_to_object_template(...)`. Also, this could use a docstring.

adam/perception/perception_graph.py
```diff
@@ -931,11 +931,44 @@ def copy_with_temporal_scopes(
 
         return PerceptionGraphPattern(dynamic=True, graph=wrapped_graph)
 
+    def copy_remove_temporal_scopes(self) -> "PerceptionGraphPattern":
```
spigo900, line 934: Same nit as for templates:

```suggestion
    def copy_removing_temporal_scopes(self) -> "PerceptionGraphPattern":
```

adam/perception/perception_graph.py
```diff
@@ -931,11 +931,44 @@ def copy_with_temporal_scopes(
...
+        """Produces a copy of this perception graph pattern
+        where all edge predicates have their temporal scope removed.
+
+        The new pattern will be static and MAY be nonsensical."""
```
spigo900, line 938: 😄

spigo900 (changes requested):
Mostly looks good. I suggest one refactoring. I marked four nits.

spigo900 (approved):
Looks good to me. All that's left is the typo. I trust you to fix that and run/fix precommit/tests, plus clean up history.

lichtefeld:
@spigo900 I'm now passing precommit and tests but I've requested a re-review because I'd like to make sure I integrated the changes from the continuous value matching correctly into the affordance learner.

adam/learner/affordances.py
```diff
@@ -0,0 +1,231 @@
...
+            match_mode=MatchMode.NON_OBJECT,
+            allowed_matches=immutablesetmultidict(
+                [
+                    (node2, node1)
+                    for previous_slot, node1 in previous_pattern_hypothesis.template_variable_to_pattern_node.items()
+                    for new_slot, node2 in current_pattern_hypothesis.template_variable_to_pattern_node.items()
+                    if previous_slot == new_slot
+                ]
+            ),
+        )
```
spigo900, line 154: This need to be updated similarly to how the other subset object learners were updated. See 2ad58ac5ff9e9e9bbd86aeeb7f36cb158e82ca02.

spigo900 (changes requested):
Looks good. One change needed to support continuous feature learning.

spigo900 (approved):

PR #1125, opened 2022-05-02
Add multi-frame VisualPerceptionRepresentation

lichtefeld:
Add ability to read and store multi-frame scenes in ADAM's phase 3 curriculum.

### DIFF
--- a/adam/curriculum/curriculum_from_files.py
+++ b/adam/curriculum/curriculum_from_files.py
@@ -15,7 +15,6 @@ from adam.paths import (
     CURRICULUM_INFO_FILE,
     SITUATION_DIR_NAME,
     SITUATION_DESCRIPTION_FILE,
-    FEATURE_YAML,
 )
 from adam.perception.visual_perception import (
     VisualPerceptionFrame,
@@ -58,6 +57,7 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
         if curriculum_type == TRAINING_CUR
         else PHASE_3_TESTING_CURRICULUM_OPTIONS,
     )
+    color_is_rgb = params.boolean("color_is_rgb", default=False)
 
     if curriculum_type == TRAINING_CUR:
         root_dir = TRAINING_CURRICULUM_DIR
@@ -92,6 +92,7 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
             ) as situation_description_file:
                 situation_description = yaml.safe_load(situation_description_file)
             language_tuple = tuple(situation_description["language"].split(" "))
+        feature_yamls = sorted(situation_dir.glob("feature_*"))
         situation = SimulationSituation(
             language=language_tuple,
             scene_images_png=sorted(situation_dir.glob("rgb_*")),
@@ -99,15 +100,28 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
             depth_pngs=sorted(situation_dir.glob("depth_*")),
             pdc_semantic_plys=sorted(situation_dir.glob("pdc_semantic_*")),
             semantic_pngs=sorted(situation_dir.glob("semantic_*")),
-            features=sorted(situation_dir.glob("feature_*")),
+            features=feature_yamls,
             strokes=sorted(situation_dir.glob("stroke_[0-9]*_[0-9]*.png")),
             stroke_graphs=sorted(situation_dir.glob("stroke_graph_*")),
+            actions=sorted(situation_dir.glob("actions_*")),
         )
         language = TokenSequenceLinguisticDescription(tokens=language_tuple)
-        perception = VisualPerceptionRepresentation.single_frame(
-            VisualPerceptionFrame.from_yaml(
-                situation_dir / FEATURE_YAML,
-                color_is_rgb=params.boolean("color_is_rgb", default=False),
+        perception = (
+            VisualPerceptionRepresentation.single_frame(
+                VisualPerceptionFrame.from_yaml(
+                    situation_dir / feature_yamls[0],
+                    color_is_rgb=color_is_rgb,
+                )
+            )
+            if len(feature_yamls) == 1
+            else VisualPerceptionRepresentation.multi_frame(
+                frames=[
+                    VisualPerceptionFrame.from_yaml(
+                        situation_dir / feature_file, color_is_rgb=color_is_rgb
+                    )
+                    for feature_file in feature_yamls
+                ],
+                action_feature=situation_dir / "action.yaml",
             )
         )
         instances.append((situation, language, perception))  # type: ignore
--- a/adam/curriculum/curriculum_from_files.py
+++ b/adam/curriculum/curriculum_from_files.py
@@ -103,27 +103,30 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
             features=feature_yamls,
             strokes=sorted(situation_dir.glob("stroke_[0-9]*_[0-9]*.png")),
             stroke_graphs=sorted(situation_dir.glob("stroke_graph_*")),
-            actions=sorted(situation_dir.glob("actions_*")),
+            actions=sorted(situation_dir.glob("action_*")),
         )
         language = TokenSequenceLinguisticDescription(tokens=language_tuple)
-        perception = (
-            VisualPerceptionRepresentation.single_frame(
+        if len(feature_yamls) == 1:
+            perception = VisualPerceptionRepresentation.single_frame(
                 VisualPerceptionFrame.from_yaml(
                     situation_dir / feature_yamls[0],
                     color_is_rgb=color_is_rgb,
                 )
             )
-            if len(feature_yamls) == 1
-            else VisualPerceptionRepresentation.multi_frame(
+        else:
+            # If we have more than one feature yaml there is also some information in a
+            # separate file relating to action feature extraction
+            with open(situation_dir / "action.yaml", encoding="utf-8") as action_yaml:
+                action_features = yaml.safe_load(action_yaml)
+            perception = VisualPerceptionRepresentation.multi_frame(
                 frames=[
                     VisualPerceptionFrame.from_yaml(
                         situation_dir / feature_file, color_is_rgb=color_is_rgb
                     )
                     for feature_file in feature_yamls
                 ],
-                action_feature=situation_dir / "action.yaml",
+                action_features=action_features,
             )
-        )
         instances.append((situation, language, perception))  # type: ignore
 
     return [
--- a/adam/curriculum/curriculum_from_files.py
+++ b/adam/curriculum/curriculum_from_files.py
@@ -103,7 +103,7 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
             features=feature_yamls,
             strokes=sorted(situation_dir.glob("stroke_[0-9]*_[0-9]*.png")),
             stroke_graphs=sorted(situation_dir.glob("stroke_graph_*")),
-            actions=sorted(situation_dir.glob("action_*")),
+            actions=sorted(situation_dir.glob("action*")),
         )
         language = TokenSequenceLinguisticDescription(tokens=language_tuple)
         if len(feature_yamls) == 1:
--- a/adam/math_3d.py
+++ b/adam/math_3d.py
@@ -22,3 +22,27 @@ class Point:
     z: float = attrib(  # pylint:disable=invalid-name
         validator=instance_of(float), converter=float
     )
+
+
+@attrs(frozen=True)
+class DepthPoint:
+    """
+    A point in 3D space represented by a depth value for the z direction.
+    """
+
+    x_coord: float = attrib(validator=instance_of(float), converter=float)
+    y_coord: float = attrib(validator=instance_of(float), converter=float)
+    depth: float = attrib(validator=instance_of(float), converter=float)
+
+    @property
+    def x(self) -> float:  # pylint:disable=invalid-name
+        return self.x_coord
+
+    @property
+    def y(self) -> float:  # pylint:disable=invalid-name
+        return self.y_coord
+
+    @property
+    def z(self) -> float:  # pylint:disable=invalid-name
+        """Provided for ease of accessing the Depth value."""
+        return self.depth
--- a/adam/math_3d.py
+++ b/adam/math_3d.py
@@ -28,6 +28,10 @@ class Point:
 class DepthPoint:
     """
     A point in 3D space represented by a depth value for the z direction.
+
+    Points are view agnostic by implementation but are assumed to exist
+    from the space view-point when compared. Our system may use multiple views
+    at different points which are not 1:1 comparable.
     """
 
     x_coord: float = attrib(validator=instance_of(float), converter=float)
@@ -43,6 +47,5 @@ class DepthPoint:
         return self.y_coord
 
     @property
-    def z(self) -> float:  # pylint:disable=invalid-name
-        """Provided for ease of accessing the Depth value."""
+    def d(self) -> float:  # pylint:disable=invalid-name
         return self.depth
--- a/adam/math_3d.py
+++ b/adam/math_3d.py
@@ -30,8 +30,8 @@ class DepthPoint:
     A point in 3D space represented by a depth value for the z direction.
 
     Points are view agnostic by implementation but are assumed to exist
-    from the space view-point when compared. Our system may use multiple views
-    at different points which are not 1:1 comparable.
+    from some specific view-point in space when compared. Our system may use
+    multiple views at different points which are not 1:1 comparable.
     """
 
     x_coord: float = attrib(validator=instance_of(float), converter=float)
--- a/adam/perception/__init__.py
+++ b/adam/perception/__init__.py
@@ -4,6 +4,7 @@ used to describe `Situation`\ s from the point-of-view of `TopLevelLanguageLearn
 """
 from abc import ABC, abstractmethod
 from enum import Enum
+from pathlib import Path
 from typing import Generic, Optional, Tuple, TypeVar
 
 from attr import attrib, attrs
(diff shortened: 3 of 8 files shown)


adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+
+            joint_point_nodes = []
+            for i, joint_feature in enumerate(
+                zip(
+                    action_stroke_graph["joint_points_world_coords"],
+                    action_stroke_graph["joint_points_xyd_coords"],
+                    action_stroke_graph["joint_points_confidence_score"],
+                )
+            ):
+                world_coord, xyd_coord, conf = joint_feature
```
spigo900, line 3046: Since we don't need the joint feature itself, you can make this slightly more compact:
```suggestion
            for i, (world_coord, xyd_coord, conf) in enumerate(
                zip(
                    action_stroke_graph["joint_points_world_coords"],
                    action_stroke_graph["joint_points_xyd_coords"],
                    action_stroke_graph["joint_points_confidence_score"],
                )
            ):
```

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+                    JointPointNode(
+                        scene_xyd_coord=DepthPoint(
+                            xyd_coord[0], xyd_coord[1], xyd_coord[2]
+                        ),
+                        world_coord=Point(world_coord[0], world_coord[1], world_coord[2]),
+                        confidence=conf,
+                        weight=conf,
+                        temporal_index=i,
+                    )
+                )
```
spigo900, line 3057: This step could be its own (static)method: Loading the joint point nodes.

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+                for object_node in involved_object_nodes
+            ]
+            edges_to_add.extend(
+                (action_decode_node, joint_node, JOINT_POINT_TRAVERSAL)
+                for joint_node in joint_point_nodes
+            )
+            edges_to_add.extend(
+                (agent_node, joint_node, HAS_PROPERTY_LABEL)
+                for joint_node in joint_point_nodes[1:-1]
+            )
```
spigo900, line 3071: This section of code reads weird to me, I think because it's grouping several mostly-unrelated things. There are lots of edges we add that don't belong in this list, and these aren't obviously related to each other -- why these? The answer is "they have the same scope" but that requires reading the later code and isn't very satisfying.

I think it would be clearer to define three separate lists, say `object_involvement_edges`, `joint_point_traversal_edges`, and `agent_has_joints_edges`, and chain the lists together at the point where we create the `DURING` edges.

This sort of ties in with breaking this section up into smaller helpers (see comment 🍌).

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+
+            for source, target, label in edges_to_add:
+                _dynamic_digraph.add_edge(
+                    source,
+                    target,
+                    label=TemporallyScopedEdgeLabel.for_dynamic_perception(
+                        label, TemporalScope.DURING
+                    ),
+                )
```
spigo900, line 3107: (:banana:) It would be nice to break the different categories of edges into smaller helper functions. Say, for example, `_link_action_to_objects()`, `_link_action_to_agent()`, and `_link_agent_to_joint_points()`.

adam/perception/perception_graph_nodes.py
```diff
@@ -136,8 +136,25 @@ class StrokeGNNRecognitionNode(GraphNode):
     def dot_label(self):
         return (
             f"StrokeGNNRecognitionNode(object_recognized={self.object_recognized}, "
-            f"confidence={self.confidence})"
+            f"confidence={self.confidence:.4f})"
```
spigo900, line 139: Good idea to limit this to 4 places. 🙂

adam/perception/perception_graph_nodes.py
```diff
@@ -136,8 +136,25 @@ class StrokeGNNRecognitionNode(GraphNode):
...
+    confidence: float = attrib(validator=instance_of(float))
+
+    def dot_label(self) -> str:
+        return (
+            f"JointPointNode(world_coord={self.world_coord}, scene_xyd_coord={self.scene_xyd_coord},"
+            f"confidence={self.confidence:.4f}, temporal_index={self.temporal_index})"
+        )
+
+    def __str__(self) -> str:
+        return f"JointPointNode([{self.temporal_index}], world_coord={self.world_coord}, scene_zyd+coord={self.scene_xyd_coord} ({self.confidence:.2f}))"
```
spigo900, line 160: Fix typo:
```suggestion
        return f"JointPointNode([{self.temporal_index}], world_coord={self.world_coord}, scene_xyd_coord={self.scene_xyd_coord} ({self.confidence:.2f}))"
```

adam/math_3d.py
```diff
@@ -22,3 +22,27 @@ class Point:
...
+    @property
+    def x(self) -> float:  # pylint:disable=invalid-name
+        return self.x_coord
+
+    @property
+    def y(self) -> float:  # pylint:disable=invalid-name
+        return self.y_coord
+
+    @property
+    def z(self) -> float:  # pylint:disable=invalid-name
```
spigo900, line 46: I'm inclined to call this `d` to emphasize the difference from world coordinates, but `z` is fine too I guess.

spigo900 (changes requested):
This looks alright. Most of the comments are me grappling with the frame->graph translation code. I think that could use some refactoring for readability. The other comments are for a typo and various nits.

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+                        temporal_specifiers=chain(after_label.temporal_specifiers),
+                    )
+                    logging.warning(
+                        f"We currently don't know how to handle a change in label "
+                        f"on an edge between the before frame and the after frame."
+                        f"Source={source}; Target={target}; "
+                        f"before label={before_label.attribute}; "
+                        f"after label={after_label.attribute}."
+                        f"As a hack, we delete the 'before' and preserve the 'after'."
+                        f"See https://github.com/isi-vista/adam/issues/666"
```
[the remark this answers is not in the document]
lichtefeld: Yeah, a hack from phase 1/2 continues to haunt us... but switching to multi-digraphs is out of scope :)

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+                decoded_object_to_cluster_node[object_name]
+                for object_name in action_features["objects"]
+            ]
+            involved_object_nodes.append(agent_node)
+
+            action_decode_node = CategoricalNode(
+                label="action_gnn",
+                value=action_stroke_graph["action_name"].lower(),
+                weight=action_stroke_graph["confidence_score"],
+            )
```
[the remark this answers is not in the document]
lichtefeld: Hmm, good point, I may wrap this node like I did the actions node so there can be a unique, unchanging, confidence value and a separate weight value.

adam/perception/perception_graph.py
lichtefeld: I need to commit this to memory for how you unpack a tuple with enumeration. Thanks!

adam/perception/perception_graph.py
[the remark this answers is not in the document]
spigo900: Indeed.

adam/perception/perception_graph.py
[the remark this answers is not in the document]
spigo900: That makes sense to me.

adam/perception/perception_graph.py
spigo900: Welcome!

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+
+        before_graph_frame = PerceptionGraph.add_temporal_scopes_to_edges(
+            self._translate_frame(perceptual_representation.frames[0]),
+            TemporalScope.BEFORE,
+        )
+
+        after_graph_frame = PerceptionGraph.add_temporal_scopes_to_edges(
+            self._translate_frame(perceptual_representation.frames[1]),
+            TemporalScope.AFTER,
```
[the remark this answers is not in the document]
lichtefeld: There is already a check for this.

https://github.com/isi-vista/adam/blob/actions-loader/adam/perception/perception_graph.py#L2942-L2947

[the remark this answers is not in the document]
spigo900: I see, thanks for pointing that out. Somehow I missed that.

adam/math_3d.py
```diff
@@ -22,3 +22,30 @@ class Point:
...
+
+
+@attrs(frozen=True)
+class DepthPoint:
+    """
+    A point in 3D space represented by a depth value for the z direction.
+
+    Points are view agnostic by implementation but are assumed to exist
+    from the space view-point when compared. Our system may use multiple views
+    at different points which are not 1:1 comparable.
```
spigo900, line 34: Nit: "the space viewpoint" sounds to me like there is necessarily just one. So:
```suggestion
    Points are view agnostic by implementation but are assumed to exist
    from some specific view-point in space when compared. Our system may use multiple views
    at different points which are not 1:1 comparable.
```

I am happy with the second sentence here though, that gets the main point across.

adam/perception/__init__.py
```diff
@@ -4,10 +4,10 @@
 """
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Generic, Optional, Tuple, TypeVar
+from typing import Generic, Optional, Tuple, TypeVar, Mapping, Any
 
 from attr import attrib, attrs
-from attr.validators import instance_of, optional
+from attr.validators import instance_of, optional, deep_mapping
```
spigo900, line 10: Was `deep_mapping` needed for something else besides action features? It looks like an unused import.

adam/perception/perception_graph_nodes.py
```diff
@@ -136,8 +136,25 @@ class StrokeGNNRecognitionNode(GraphNode):
...
+    confidence: float = attrib(validator=instance_of(float))
+
+    def dot_label(self) -> str:
+        return (
+            f"JointPointNode(world_coord={self.world_coord}, scene_xyd_coord={self.scene_xyd_coord},"
+            f"confidence={self.confidence:.4f}, temporal_index={self.temporal_index})"
+        )
+
+    def __str__(self) -> str:
+        return f"JointPointNode([{self.temporal_index}], world_coord={self.world_coord}, scene_zyd_coord={self.scene_xyd_coord} ({self.confidence:.2f}))"
```
spigo900, line 160: Ah, similar typo here, missed it the first time.
```suggestion
        return f"JointPointNode([{self.temporal_index}], world_coord={self.world_coord}, scene_xyd_coord={self.scene_xyd_coord} ({self.confidence:.2f}))"
```

adam/perception/perception_graph.py
```diff
@@ -2932,9 +2876,225 @@ def _translate_frames(
...
+        )
+        self._link_action_to_agent(_dynamic_digraph, agent_node)
+        self._link_agent_and_action_to_joint_points(
+            _dynamic_digraph, agent_node, action_decode_node, joint_point_nodes
+        )
+
+        return _dynamic_digraph
+
+    @staticmethod
+    def _link_actions_to_objects(
```
spigo900, line 2943: Typo:
```suggestion
    def _link_action_to_objects(
```

adam/perception/perception_graph.py
```diff
@@ -2932,9 +2876,225 @@ def _translate_frames(
...
+                temporal_index=i,
+            )
+            for i, (world_coord, xyd_coord, conf) in enumerate(
+                zip(
+                    action_stroke_graph["joint_points_world_coords"],
+                    action_stroke_graph["joint_points_xyd_coords"],
+                    action_stroke_graph["joint_points_confidence_score"],
+                )
+            )
+        ]
```
spigo900, line 2930: Hmm, this code expects to get a sequence of 3 floats, a sequence of 3 floats, and a single float out of iterating over this zip, but when I try that locally with the `walk.yaml` data (not running all the surrounding code, but just loading the action features and stroke graph manually in the interpreter) I'm getting *lists of lists*.

Thinking of these as arrays, the xyd and world coords for the joint points should be of shape T x K x 3. When we zip them we're getting (I think) K x 3 slices. (For confidence we're getting a K-size slice/list.) I think what we want here is another zip:
```suggestion
        joint_point_nodes = [
            JointPointNode(
                scene_xyd_coord=DepthPoint(xyd_coord[0], xyd_coord[1], xyd_coord[2]),
                world_coord=Point(world_coord[0], world_coord[1], world_coord[2]),
                confidence=conf,
                weight=conf,
                temporal_index=i,
            )
            for i, (world_coords, xyd_coords, confs) in enumerate(
                zip(
                    action_stroke_graph["joint_points_world_coords"],
                    action_stroke_graph["joint_points_xyd_coords"],
                    action_stroke_graph["joint_points_confidence_score"],
                )
            )
            for world_coord, xyd_coord, conf in zip(
                world_coords, xyd_coords, confs
            )
        ]
```

spigo900 (changes requested):
Looks good, a few comments.

adam/curriculum/curriculum_from_files.py
```diff
@@ -92,22 +92,36 @@ def phase3_load_from_disk(  # pylint: disable=unused-argument
...
             scene_images_png=sorted(situation_dir.glob("rgb_*")),
             scene_point_cloud=tuple(situation_dir.glob("pdc_rgb_*")),
             depth_pngs=sorted(situation_dir.glob("depth_*")),
             pdc_semantic_plys=sorted(situation_dir.glob("pdc_semantic_*")),
             semantic_pngs=sorted(situation_dir.glob("semantic_*")),
-            features=sorted(situation_dir.glob("feature_*")),
+            features=feature_yamls,
             strokes=sorted(situation_dir.glob("stroke_[0-9]*_[0-9]*.png")),
             stroke_graphs=sorted(situation_dir.glob("stroke_graph_*")),
+            actions=sorted(situation_dir.glob("actions_*")),
```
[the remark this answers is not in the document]
spigo900: Unless I'm mistaken one would still need to add `action.yaml` separately -- the change to globbing for `action_*` wouldn't include it?

adam/perception/perception_graph.py
lichtefeld: 🤔  That's right there are K joint points we are tracking so we need to iterate over those too. I may be getting close to agreeing with you that this should maybe be a static function even though its just a generator... this is a very complicated generator. I'm thinking of doing the following and tracking a `node_index` as well as a `temporal_index`. This means I should probably check with ASU that the joints are always in the same order between decode outputs.

```py
joint_point_nodes = [
    JointPointNode(
        scene_xyd_coord=DepthPoint(xyd_coord[0], xyd_coord[1], xyd_coord[2]),
        world_coord=Point(world_coord[0], world_coord[1], world_coord[2]),
        confidence=conf,
        weight=conf,
        temporal_index=temporal_index,
        joint_index=joint_index,
    )
    for joint_index, (world_coords, xyd_coords, confs) in enumerate(
        zip(
            action_stroke_graph["joint_points_world_coords"],
            action_stroke_graph["joint_points_xyd_coords"],
            action_stroke_graph["joint_points_confidence_score"],
        )
    )
    for temporal_index, (world_coord, xyd_coord, conf) in enumerate(
        zip(
            world_coords, xyd_coords, conf
        )
    )
]
```

Adjacent to this topic, how should i go about matching these nodes?

spigo900: The single-loop generator looked okay -- I agree though, with the two loops it's looking pretty complicated and probably worth a staticmethod. That code makes sense to me -- it probably is worth tracking the joint index as well for debugging purposes.

Yes, I think that's a good idea to check with ASU about joint ordering. Another question to check is, is the time dimension first or the joint dimension? i.e. is it T x K x 3 or K x T x 3. The above code would use the latter -- I thought time came first but I'm not confident in that so probably worth checking.

adam/perception/__init__.py
lichtefeld: Huh this is, pylint didn't catch it but pycharm did!

spigo900: Strange!

adam/perception/perception_graph.py
spigo900: One could even go further and rewrite the method as building up a list using loops -- more like a generator function -- which makes the control flow slightly more straightforward (top to bottom) but I'm not going to request that. Thanks for extracting it.

adam/curriculum/curriculum_from_files.py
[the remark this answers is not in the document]
spigo900: @lichtefeld As far as I can tell this hasn't been addressed?

adam/perception/perception_graph.py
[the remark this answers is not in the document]
spigo900: @lichtefeld Has this been addressed? It looks like there's still only a weight.

adam/perception/perception_graph.py
```diff
@@ -2932,10 +2946,168 @@ def _translate_frames(
...
+            _dynamic_digraph.add_edge(
+                agent_node,
+                joint_point_nodes[-1],
+                label=TemporallyScopedEdgeLabel.for_dynamic_perception(
+                    HAS_PROPERTY_LABEL, TemporalScope.AFTER
+                ),
+            )
+
+            # Also to help clarify the agent we mark them as 'VOLITIONAL' from the original symbolic system
+            # This is mostly a way of representing the structured knowledge in the feature file.
```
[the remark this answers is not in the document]
spigo900: AFAICT the joint point work is restricted to humans so I think it's reasonable to assume the object will never be encoded in the `action_stroke_graph`, and we can forget about this for now unless a problem comes up. I am wondering now though how simple interactions get encoded.

spigo900 (commented):
Two minor issues left to address.

adam/curriculum/curriculum_from_files.py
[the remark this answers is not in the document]
lichtefeld: Oops! I missed your follow up comment here. Sorry

spigo900 (approved):
Looks good to me.

PR #1133, opened 2022-05-23
Add M5 Action Experiments

lichtefeld:
M5 action experiments from M5.

Pending:
* Debugging an issue where non-person objects cause identification issues for Actions


scripts/run_experiment.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Starting run at: $(date)"
+echo "Job ID: $SLURM_JOB_ID"
+
+# Run GPT-2 script
+echo
+time PYTHONPATH=. python adam/experiment/log_experiment.py "$3"
+echo
+
+# Finish up the job
+echo "Job finished with exit code $? at: $(date)"
```
spigo900, line 31: I think this is subtly not what you would expect, for two reasons:

1. The blank echo clobbers the return code -- yes, even though it's a builtin. 🙃
2. `set -e` means Slurm will fail after `time` if the Python script exits nonzero. So the exit code echo will only ever print "exit code 0". You can fix this by wrapping in an if statement (to override the nonzero exit behavior) and grabbing the exit code ahead of time (see below). Unfortunately, you can't just do `|| true` which does prevent the script crashing but also clobbers the exit code.

(Optionally see [this guide](https://github.com/anordal/shellharden/blob/master/how_to_do_things_safely_in_bash.md) for some more Bash quirks)

```suggestion
if time PYTHONPATH=. python adam/experiment/log_experiment.py "$3"; then
    EXITCODE=0
else
    EXITCODE=$?
fi
echo

# Finish up the job
echo "Job finished with exit code $EXITCODE at: $(date)"
```

scripts/run_experiment.sh
```diff
@@ -0,0 +1,31 @@
...
+#SBATCH --output=data/slurm_logs/%x-%j.out    # %x-%j means JOB_NAME-JOB_ID.
+#SBATCH --requeue
+#SBATCH --exclude="saga01,saga02,saga03,saga04,saga05,saga06,saga07,saga08,saga10,saga11,saga12,saga13,saga14,saga15,saga16,saga17,saga18,saga19,saga20,saga21,saga22,saga23,saga24,saga25,saga26,gaia01,gaia02"
+# The excluded nodes above have slower CPUs on the cluster. This list targets our execution at the CPUs with the highest
+# base & boost frequency
+
+set -euo pipefail
+
+source "$1"/etc/profile.d/conda.sh
+conda activate "$2"
```
spigo900, line 18: I'm not sure we strictly need these lines? The Slurm job inherits the conda setup from your bash session when you run this using sbatch. That includes the currently activated environment.

scripts/run_experiment.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+#SBATCH --partition=ephemeral
+#SBATCH --qos=ephemeral
+#SBATCH --ntasks=1
+#SBATCH --mem-per-cpu=8G
+#SBATCH --cpus-per-task=1
+#SBATCH --job-name=ADAM_EXPERIMENT
+#SBATCH --output=data/slurm_logs/%x-%j.out    # %x-%j means JOB_NAME-JOB_ID.
+#SBATCH --requeue
```
spigo900, line 10: TIL `--requeue` is an option you can specify instead of just manually queueing up a bunch of dependent jobs... 🙃

scripts/run_experiment.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
spigo900, line 1: Good idea making a Slurm script.

scripts/organize_action_curriculum.py
```diff
@@ -0,0 +1,208 @@
...
+                with open(action_feature_file, encoding="utf-8") as action_feature_file:
+                    action_features = yaml.safe_load(action_feature_file)
+                action_features["object_name"] = "0"
+                action_features["objects"] = [str(value+1) for value in range(len(objects_to_append.get(f"{args.input_split}_{action_debug_name}_{ex}_{cam}", [])))] if action_features["objects"] is None else action_features["objects"]
+                with open(output_situation / "action.yaml", "w", encoding="utf-8") as action_feature_out:
+                    yaml.dump(action_features, action_feature_out)
+
+                # Description file
+                with open(output_situation / "description.yaml", "w", encoding="utf-8") as description_file:
+                    yaml.dump({"language": f"a person {ACTION_DEBUG_NAME_TO_NAME.get(action_debug_name, action_debug_name)}"}, description_file)
```
spigo900, line 195: No change needed. This is probably for the best given we're using subset on actions where the other objects may not be visible. This does seem to mean we can't learn affordances with the current dataset?

scripts/organize_action_curriculum.py
```diff
@@ -0,0 +1,208 @@
...
+                ]}
+                with open(output_situation / "feature_0.yaml", "w", encoding="utf-8") as feature_obj_0:
+                    yaml.dump(object_features, feature_obj_0)
+                with open(output_situation / "feature_1.yaml", "w", encoding="utf-8") as feature_obj_1:
+                    yaml.dump(object_features, feature_obj_1)
+                # Action Feature copy
+                with open(action_feature_file, encoding="utf-8") as action_feature_file:
+                    action_features = yaml.safe_load(action_feature_file)
+                action_features["object_name"] = "0"
+                action_features["objects"] = [str(value+1) for value in range(len(objects_to_append.get(f"{args.input_split}_{action_debug_name}_{ex}_{cam}", [])))] if action_features["objects"] is None else action_features["objects"]
```
spigo900, line 189: This is a bit dense. I suggest splitting it up more. Like:

```suggestion
                annotated_objects = objects_to_append.get(f"{args.input_split}_{action_debug_name}_{ex}_{cam}", [])
                action_features["objects"] = [
                    str(value+1) for value in range(len(annotated_objects))
                ] if action_features["objects"] is None else action_features["objects"]
```

parameters/experiments/p3/m5_actions.person_only.params
```diff
@@ -0,0 +1,64 @@
...
+    curriculum_type: "training"
+    curriculum: "m5_actions_person_only"
+    color_is_rgb: True
+test_curriculum:
+    curriculum_type: "testing"
+    curriculum: "m5_actions_person_only"
+    color_is_rgb: True
+
+# Experiment Configuration
+experiment: "m4_core"
```
spigo900, line 41: I think this is only used for logging, but for clarity:

```suggestion
experiment: "m5_actions_person_only"
```

parameters/experiments/p3/m5_actions.params
```diff
@@ -0,0 +1,64 @@
...
+    curriculum_type: "training"
+    curriculum: "m5_actions"
+    color_is_rgb: True
+test_curriculum:
+    curriculum_type: "testing"
+    curriculum: "m5_actions"
+    color_is_rgb: True
+
+# Experiment Configuration
+experiment: "m4_core"
```
spigo900, line 41: I think this is only used for logging, but for clarity:

```suggestion
experiment: "m5_actions"
```

spigo900 (changes requested):
Looks good. I think the only proper change requests I have are a few minor typos and a few changes to the Slurm and reorg scripts.

adam/perception/perception_graph.py
```diff
@@ -603,6 +603,10 @@ def _to_dot_node(
             label = f"Object Cluster {unwrapped_perception_node.cluster_id} | View: {unwrapped_perception_node.viewpoint_id}"
         elif isinstance(unwrapped_perception_node, ObjectStroke):
             label = f"Stroke: [{', '.join(str(point) for point in unwrapped_perception_node.normalized_coordinates)}]"
+        elif isinstance(unwrapped_perception_node, TrajectoryRecognitionNode):
+            label = f"Trajectory Recognition: {unwrapped_perception_node.action_recognized} ({unwrapped_perception_node.confidence})"
+        elif isinstance(unwrapped_perception_node, JointPointNode):
+            label = f"JointPointNode ({unwrapped_perception_node.joint_index}.{unwrapped_perception_node.temporal_index}) world: {unwrapped_perception_node.world_coord}"
```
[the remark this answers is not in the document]
lichtefeld: I did too! But I hit a crash on this `elif` ladder so I added them here. I should probably open an issue to try and remove this ladder and use the existing methods. I suspect its one of the ~multiple ways we can print debug graphs that's not consistent...

scripts/organize_action_curriculum.py
lichtefeld: Ah! Good catch -- Actually the `m5_action` dataset was a version of this which encoded the first object in the annotated set (if it existed) as the object eg. `a person throws a ball` -- But I had just manually altered this generation. I should set this via a flag.

adam/curriculum/curriculum_from_files.py
```diff
@@ -36,6 +38,8 @@
     "m5_objects_eval",
     "m5_objects_v0_eval",
     "m5_objects_v0_with_mugs_eval",
+    "m5_actions",
+    "m5_actions_person_only",
```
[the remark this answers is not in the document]
lichtefeld: I didn't post-pend my test curriculums with `_eval` -- I probably should have done so to make the curriculums more clear.

adam/perception/perception_graph.py
```diff
@@ -1767,7 +1771,7 @@ def gather_nodes_to_excise(focus_node: NodePredicate) -> None:
                     gather_nodes_to_excise(predecessor)
 
         last_failed_node = match_failure.last_failed_pattern_node
-        logging.info("Relaxation: last failed pattern node is %s", last_failed_node)
+        logging.debug("Relaxation: last failed pattern node is %s", last_failed_node)
```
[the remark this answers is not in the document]
lichtefeld: They are _situationally useful_. I finally remembered to make this change, it's been on my local list of wants for a while.

scripts/run_experiment.sh
lichtefeld: Huh, this probably means the scripts from the source of this bash script are also wrong.

scripts/run_experiment.sh
lichtefeld: You're correct -- I'm used to configuring the jobs for within Pegasus mostly where this is strictly necessary. I'll remove it from this script.

adam/curriculum/curriculum_from_files.py
```diff
@@ -28,6 +28,8 @@
     "m5_objects",
     "m5_objects_v0",
     "m5_objects_v0_with_mugs",
+    "m5_actions",
```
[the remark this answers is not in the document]
lichtefeld: It does the postfix `person_only` is a reference to the language generation not the actions contained in the set.

scripts/run_experiment.sh
lichtefeld: I believe `--requeue` will just automatically requeue the job if its pre-empted by another job because of paritions. It's probably not needed for this script given the execution time is under an hour and ephemeral guarantees at least that long once a job starts.

spigo900 (changes requested):
Looks good to me. Just a few typos.

scripts/organize_action_curriculum.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+# These must also match the ontology node labels due to how the GNN recognizer works.
+ACTION_DEBUG_NAME_TO_NAME = {
+    "running": "run",
+    "walking": "walk",
+    "falling": "fall",
+    "drinking": "drink",
+    "writing": "write",
+}
```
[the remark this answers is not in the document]
lichtefeld: Functionally it doesn't matter for our purpose. The language generation aspect of the current phase is mostly descriptive so we know what's been decoded, proper grammar rules are irrelevant. If I could have removed articles from phase 3 easily I would have (given a polite request in a PI meeting about it)

spigo900 (approved):
Looks good to me. I'll leave it up to you whether to rename the test curricula. I'd like to have the names consistent but I acknowledge that it might be (ETA: too much of) a pain.

PR #1141, opened 2022-06-03
Implement mapping affordance learner

lichtefeld:
@spigo900 Here's the initial implementation of the mapping affordance learner so you can take a look at the interface to return the affordance concepts given an object concept. Let me know if you foresee any issues with the integration with the contrastive learner.


adam/experiment/experiment_utils.py
```diff
@@ -459,6 +459,11 @@ def build_affordance_learner_factory(
             language_mode=language_mode,
             min_continuous_feature_match_score=min_continuous_feature_match_score,
         )
+    elif learner_type == "mapping":
```
spigo900, line 462: So you can't use both this learner and the pattern-based affordance learner?

lichtefeld: You can, I've noticed that this push doesn't have the added logic to handle loading this second type of affordance learner at the same time. I have that fixed on my local branch.

spigo900:
@lichtefeld But we can't propagate that way, can we? The problem in my mind is that the mapped affordance nodes are based on us recognizing an object. If we don't recognize the object, we can't add the mapped affordance. If we add the affordance to the object pattern, then to recognize the affordance we need to recognize the object but to recognize the object we need to recognize the affordance. Does this make sense?

lichtefeld:
Ah right 🙃 I've made a cycle. I suppose the only weights/confidence we could track for the affordance contrastively is not related to the graph pattern for the object concept but rather the weight of the affordance appearing... (which should always be 100%). 

I think the best we can do with these affordances is provide differences/similarities between the two object concepts then.

spigo900:
@lichtefeld I'm thinking about how to produce string descriptions in post_decode for the UI. I think we are missing some methods we would want to be able show descriptions.

We will probably want to show something like `can be SLOT2 in "SLOT1 eats SLOT2"`, right? I am thinking the integrated learner has to generate this string. So I think we want the integrated learner to be able to get (a) the affordance slot, and (b) the action concept so it can look up (c) the template for the action. It doesn't look like the learner stores these things. Does that sound right? Am I missing any public method/etc. that gives the info we need?

adam/learner/affordances.py
```diff
@@ -233,3 +248,94 @@ def _fallback_templates(
...
+
+    def _new_concept(self, debug_string: str) -> Concept:
+        return AffordanceConcept(debug_string)
+
+    def _learning_step(
+        self,
+        language_perception_semantic_alignment: LanguagePerceptionSemanticAlignment,
+        bound_surface_template: SurfaceTemplateBoundToSemanticNodes,
+    ) -> None:
+        for _, object_semantic in bound_surface_template.slot_to_semantic_node.items():
```
spigo900, line 282: Maybe specify `object_semantic_node` (not just `object_semantic`)?

Also, maybe worth a comment: There should only ever be one `object_semantic` here. Or maybe this should use `only()` instead of a loop?

adam/learner/affordances.py
```diff
@@ -233,3 +248,94 @@ def _fallback_templates(
...
+            if debug_name in self.debug_to_concept:
+                concept = self.debug_to_concept[debug_name]
+            else:
+                concept = self._new_concept(debug_name)
+                self.debug_to_concept[debug_name] = concept
+
+            if object_semantic.concept not in self.concept_to_affordances:
+                self.concept_to_affordances[object_semantic.concept] = set()
+            if concept not in self.concept_to_affordances[object_semantic.concept]:
+                self.concept_to_affordances[object_semantic.concept].add(concept)
```
spigo900, line 293: I think you can write this slightly more concisely as:

```suggestion
            self.concept_to_affordances.setdefault(
                object_semantic.concept, set()
            ).add(concept)
```

adam/perception/perception_graph.py
```diff
@@ -2984,8 +2984,8 @@ def _link_action_to_objects(
     ) -> None:
         for object_node in object_nodes:
             perception_digraph.add_edge(
-                object_node,
                 action_node,
+                object_node,
```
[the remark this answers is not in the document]
spigo900: Makes sense.

adam/learner/affordances.py
lichtefeld: TIL: Dictionaries have a `setdefault` method.

adam/learner/affordances.py
```diff
@@ -233,3 +248,95 @@ def _fallback_templates(
...
+
+    def _new_concept(self, debug_string: str) -> Concept:
+        return AffordanceConcept(debug_string)
+
+    def _learning_step(
+        self,
+        language_perception_semantic_alignment: LanguagePerceptionSemanticAlignment,
+        bound_surface_template: SurfaceTemplateBoundToSemanticNodes,
+    ) -> None:
+        # This loop should only execute once in standard cases
```
spigo900, line 282: I think it's worth pointing out why -- an affordance has exactly one associated slot and so there should only be one slot/slot filler pair, which is the object involved in the affordance.

spigo900 (approved):
Looks good. I suggest elaborating a little more on this comment but everything else looks good.

adam/learner/affordances.py
```diff
@@ -233,3 +248,97 @@ def _fallback_templates(
...
+        # As a normal affordance only has one slot filler,
+        # the object the affordance affects.
```
spigo900, line 284: This scanned a little weird to me, just suggesting some minor changes. Otherwise looks good, thanks for adding.

```suggestion
        # This loop should only execute once in standard cases
        # because a normal affordance only has one slot filler, namely
        # the object that has the affordance.
```

spigo900 (approved):

sidharth-sundar (approved):
Looks good, I don't have anything too add from my end.

PR #1142, opened 2022-06-06
1113 rgbcolornodepredicate are slightly bugged

sidharth-sundar:
Fairly quick fix, just altered upper limit of ranges in RGB Nodes/Predicates to 256 instead of 255.

### DIFF
--- a/tests/perception/perception_graph_test.py
+++ b/tests/perception/perception_graph_test.py
@@ -60,7 +60,10 @@ from adam.perception.perception_graph import (
     RelationTypeIsPredicate,
     HAS_PROPERTY_LABEL,
 )
-from adam.perception.perception_graph_nodes import ContinuousNode
+from adam.perception.perception_graph_nodes import (
+    ContinuousNode,
+    RgbColorNode
+)
 from adam.perception.perception_graph_predicates import (
     DistributionalContinuousPredicate,
     AnyObjectPredicate,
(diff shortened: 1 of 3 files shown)


tests/perception/perception_graph_test.py
```diff
@@ -919,6 +920,21 @@ def test_simulated_one_object_graph():
...
+def test_rgb_color_predicate():
+    assert RgbColorPredicate(red=0, blue=0, green=0, weight=1.0)
+    assert RgbColorPredicate(red=255, blue=255, green=255, weight=0.0)
+
+
+def test_rgb_color_perception():
+    assert RgbColorPerception(red=0, blue=0, green=0)
+    assert RgbColorPerception(red=255, blue=255, green=255)
+
+
```
spigo900, line 937: Minor changes:

1. Since these are boundary tests, put that in the name? `test_rgb_color_{name}_accepts_valid_values` say
2. No need to assert the objects -- `attrs` will raise an error when validation fails leading to test failure. The real thing we're testing is whether these raise an error, so you can just call the constructors without the assert. If the pre-commit linters complain, disable them for these lines.a

spigo900 (changes requested):
Looks good, just two nits on the tests.

tests/perception/perception_graph_test.py
sidharth-sundar: Thanks for the tips! I've updated the tests accordingly.

spigo900 (approved):
Looks good to me. You can merge this after the continuous integration passes.

PR #1143, opened 2022-06-07
Set observers to only log with log_learner_state=True

sidharth-sundar:
Just added an indent to check if `log_learner_state=True` before logging observers


spigo900 (approved):
Looks good to me.

Issue #1140, opened 2022-06-03
Contrastive learner matching/updates are slightly non-deterministic


spigo900:
The contrastive learner's graph matching logic, thus its updates, are slightly non-deterministic in that they arbitrarily take the first available match. That means the exact updates to be applied in some edge cases could be nondeterministic. In practice I think this isn't a problem.

Here's an example of the issue.. Suppose pattern P_A has a feature `alpha` that could match to either of two things in graph A, say observed features `x` and `y`. If `x` is in the difference of graph A against graph B, and `y` is not, then whether we update the node weight for `alpha` depends entirely on whether the graph matching algorithm "decides" to match it to `x` or `y` which in turn probably depends on an arbitrary relative ordering of those nodes. 

In practice this scenario isn't a problem for two reasons. First, we only have one object in each scene. Second, so far an object won't have multiple of the same feature type.

This could become more of a problem however if we were to try and generalize using "super-nodes" (#1136), e.g. "red OR green," "table OR chair" etc.

spigo900:
Wrote this up because I noticed this [TODO](https://github.com/isi-vista/adam/blob/bf8916c441b185ec92ad70ec039e3e45acd72875/adam/learner/contrastive_learner.py#L191). That should get linked here eventually.

PR #1144, opened 2022-06-09
1138 contrastive no primary template

### LINKED ISSUE #1138: Teaching contrastive learner fails to retrieve meaning when apprentice has no primary template for a scene
When its apprentice has no primary template for a scene, the contrastive learner fails to retrieve a meaning for the scene leading to a crash.

This is a problem because it happens with the subset learner in the simulated setting. I had it happen with contrastive object learning on the M5 objects curriculum. The "problem object" is `sofa`. We end up with two hypotheses that differ only in their stroke graph. Specifically, one has a single-point stroke graph and the other has a two-linked-points stroke graph. This is stable -- the two-points hypothesis do not seem to be contradicted/intersected out. Now, abstractly, when you ask the subset learner for its concept to template map, it includes only those hypotheses it's sure of, i.e. where it has only one. So in this situation the contrastive learner tries to get a hypothesis for `sofa` and match/update that to the `sofa` scene, but the subset learner won't return any such hypothesis because we have two for `sofa`.

It is simple enough to work around this by setting `beam_size: 1`, [which is what I did](https://github.com/isi-vista/adam/blob/bf8916c441b185ec92ad70ec039e3e45acd72875/parameters/experiments/p3/m5_contrastive_object_learning.params#L6). This seems a bit hacky, however.

# Fix: Change the contrastive learner to handle multiple hypothesis
The obvious fix is to return _some hypothesis_ for these scenes. However, that complicates contrastive learning and the method we use to propose pattern updates. Both currently assume each concept of interest has only one candidate meaning.

So, the contrastive learner would have to change. We would have to contrast each candidate meaning vs. each candidate meaning instead of assuming we have 1 pattern for each concept. I suspect there would be some plumbing to do to make contrastive learning work this way.

And we would have to change how we propose updated hypotheses. Currently, we propose updates by concept, and demand that each concept to be updated have only one hypothesis. To fix this, maybe instead of proposing a simple one-level dict, we could propose a nested dict: `Mapping[Concept, Mapping[PerceptionGraphPattern, PerceptionGraphPattern]]`. Patterns are hashable so this should work, though it is a little ugly.

Returning to the obvious part -- we should return some hypothesis -- we would want to add a new method to template learners at the same level as `concepts_to_patterns`/`self._primary_templates`. Here are two possible variations on what to return:

1. return the set of all the hypotheses, `self._concept_to_hypotheses[concept]`, or
2. return only the "leading" hypothesis, `self._concept_to_hypotheses[concept][0]`.

I'm inclined to do things as in (1) i.e. return all the hypotheses. I don't see a strong reason not to update everything -- we shouldn't have so many patterns that speed is a problem.

sidharth-sundar:
Precommit and tests pass. I mainly just added a new function for retrieving a list of concepts, a function for retrieving multiple hypotheses from a learner, and updated the preexisting functions to work around multiple hypotheses rather than a single one.

Closes #1138

### DIFF
--- a/adam/learner/__init__.py
+++ b/adam/learner/__init__.py
@@ -351,3 +351,8 @@ class ApprenticeLearner(Protocol):
         """
         Retrieve top n hypotheses for input concept.
         """
+
+    def get_concepts(self) -> ImmutableSet[Concept]:
+        """
+        Retrtieve all concepts from learner
+        """
--- a/adam/learner/__init__.py
+++ b/adam/learner/__init__.py
@@ -22,6 +22,7 @@ from immutablecollections import (
     immutabledict,
     ImmutableSetMultiDict,
     immutablesetmultidict,
+    ImmutableSet,
 )
 from more_itertools import first
 from networkx import isolates
@@ -341,4 +342,12 @@ class ApprenticeLearner(Protocol):
         is a `PerceptionGraphPattern`.
 
         The learner may do with these as it will.
+        """
+
+    @abstractmethod
+    def concept_to_hypotheses(
+        self, concept: Concept, top_n: Optional[int]
+    ) -> ImmutableSet[PerceptionGraphTemplate]:
+        """
+        Retrieve top n hypotheses for input concept.
         """
--- a/adam/learner/__init__.py
+++ b/adam/learner/__init__.py
@@ -346,7 +346,7 @@ class ApprenticeLearner(Protocol):
 
     @abstractmethod
     def concept_to_hypotheses(
-        self, concept: Concept, top_n: Optional[int]
+        self, concept: Concept
     ) -> ImmutableSet[PerceptionGraphTemplate]:
         """
         Retrieve top n hypotheses for input concept.
--- a/adam/learner/__init__.py
+++ b/adam/learner/__init__.py
@@ -33,6 +33,7 @@ from adam.learner.alignments import (
     PerceptionSemanticAlignment,
 )
 from adam.learner.language_mode import LanguageMode
+from adam.learner.perception_graph_template import PerceptionGraphTemplate
 from adam.learner.surface_templates import (
     STANDARD_SLOT_VARIABLES,
     SurfaceTemplate,
@@ -327,10 +328,17 @@ class ApprenticeLearner(Protocol):
 
     @abstractmethod
     def propose_updated_hypotheses(
-        self, concept_to_updated_patterns: Dict[Concept, PerceptionGraphPattern]
+        self,
+        concept_to_hypothesis_updates: Dict[
+            Concept, Dict[PerceptionGraphTemplate, PerceptionGraphPattern]
+        ],
     ) -> None:
         """
         Propose new/updated hypotheses to the learner.
+
+        This expects hypothesis updates to be given as mappings of the form `old_hypothesis ->
+        new_hypothesis` where `old_hypothesis` is a `PerceptionGraphTemplate` and `new_hypothesis`
+        is a `PerceptionGraphPattern`.
 
         The learner may do with these as it will.
         """
--- a/adam/learner/contrastive_learner.py
+++ b/adam/learner/contrastive_learner.py
@@ -30,6 +30,7 @@ from adam.learner import (
     SurfaceTemplate,
 )
 from adam.learner.objects import SubsetObjectLearner
+from adam.learner.perception_graph_template import PerceptionGraphTemplate
 from adam.ontology import OntologyNode
 from adam.ontology.ontology import Ontology
 from adam.perception import MatchMode
@@ -186,7 +187,7 @@ class TeachingContrastiveObjectLearner(ContrastiveLearner):
 
     def _match_concept_pattern_to_multiple_graphs(
         self, concept: Concept, graph: PerceptionGraph, top_n: Optional[int] = None
-    ) -> Optional[Mapping[PerceptionGraphPatternMatch, PerceptionGraphPatternMatch]]:
+    ) -> Mapping[PerceptionGraphPatternMatch, PerceptionGraphPatternMatch]:
         # Arbitrarily take the first match per hypothesis.
         #
         # Because there may be more than one match per hypothesis, this results in non-deterministic behavior in
@@ -194,7 +195,7 @@ class TeachingContrastiveObjectLearner(ContrastiveLearner):
         # https://github.com/isi-vista/adam/issues/1140
         # TODO issue
         return {
-            apprentice_concept.graph_pattern: apprentice_concept.graph_pattern.matcher(
+            apprentice_concept: apprentice_concept.graph_pattern.matcher(
                 graph,
                 match_mode=MatchMode.OBJECT,
             ).relax_pattern_until_it_matches_getting_match(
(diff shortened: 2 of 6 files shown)


adam/learner/contrastive_learner.py
```diff
@@ -146,7 +146,9 @@ class TeachingContrastiveObjectLearner(ContrastiveLearner):
         validator=instance_of(Counter), factory=Counter, init=False
     )
 
-    def learn_from(self, matching: LanguagePerceptionSemanticContrast) -> None:
+    def learn_from(
+        self, matching: LanguagePerceptionSemanticContrast, top_n: Optional[int] = None
```
spigo900, line 150: I prefer to make `top_n` an attribute/property here. I think `top_n` may not make sense for all contrastive learners. It's not required, so this doesn't change the interface, but the script has to pass it in at the point we call learn_from. That means that we have to special case any contrastive learner that doesn't accept this parameter. So it's sort of de facto adding to the contrastive learning interface.

If we use a property it is slightly cleaner to slot a different contrastive learner into the contrastive learning script. We still have to deal with setting `top_n`, but we can do it in a factory function for the contrastive learner where we have to do special casing anyway. See `log_experiment.py`/`build_object_learner_factory()` for example. That contains the specialcasing to our setup so we don't have to worry about those details for the rest of the experiment script.

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +265,83 @@ def log_hypotheses(self, log_output_path: Path) -> None:
     def _update_counts(
         self,
         concept: Concept,
-        pattern_to_graph_match: PerceptionGraphPatternMatch,
+        pattern_to_graph_matches: Mapping[
```
spigo900, line 268: Since we only use the values here, I prefer to make this `Sequence[Optional[PerceptionGraphPatternMatch]]` Or better, `Sequence[PerceptionGraphPatternMatch]` and just filter out the `None` s at the call site.

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +265,83 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
-        new_pattern_digraph = DiGraph()
-
-        for node, data in old_pattern_digraph.nodes(data=True):
-            old_to_new_node[node] = self._re_weighted_node(concept, node)
-            new_pattern_digraph.add_node(old_to_new_node[node], **data)
-
-        for old_u, old_v, data in old_pattern_digraph.edges(data=True):
-            new_pattern_digraph.add_edge(
-                old_to_new_node[old_u], old_to_new_node[old_v], **data
+            new_pattern = PerceptionGraphPattern(
```
spigo900, line 340: Might as well inline `new_pattern` as it's fairly simple.

adam/learner/contrastive_learner.py
```diff
@@ -179,9 +183,26 @@ def learn_from(self, matching: LanguagePerceptionSemanticContrast) -> None:
...
+            ).relax_pattern_until_it_matches_getting_match(
+                ontology=self._ontology, trim_after_match=None
+            )
+            for apprentice_concept in self.apprentice.concept_to_hypotheses(
+                concept, top_n
+            )
+        }
 
-    def _match_concept_pattern_to_graph(
+    def _match_concept_pattern_to_primary_graph(
```
spigo900, line 205: It's fine to just delete this since it's a private method with no callers left.

spigo900 (changes requested):
Looks good overall. Some nitpicking in the details. :)

parameters/experiments/p3/m5_contrastive_object_learning.params
```diff
@@ -3,7 +3,7 @@ _includes:
 
 # Learner Configuration
 learner: simulated-integrated-learner-params
-beam_size: 1
+beam_size: 10
```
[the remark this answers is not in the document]
sidharth-sundar: Got it, I've shifted the parameters to their own updated file now and restored the older one.

adam/learner/subset.py
```diff
@@ -257,54 +256,79 @@ def _fallback_templates(
...
         )
 
+    def concept_to_hypotheses(
+        self, concept: Concept, top_n: Optional[int] = None
+    ) -> ImmutableSet[PerceptionGraphTemplate]:
+        if concept in self._concept_to_hypotheses:
+            return immutableset(
+                self._concept_to_hypotheses[concept]
+                if not top_n
+                else self._concept_to_hypotheses[concept][:top_n]
```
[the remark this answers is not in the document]
sidharth-sundar: Ah, I didn't know that was how that works. This makes it simpler!

adam/learner/contrastive_learner.py
sidharth-sundar: Ok, I've shifted top_n to be an attribute instead of a function parameter.

parameters/experiments/p3/m5_contrastive_object_learning_without_logging.params
```diff
@@ -0,0 +1,58 @@
+_includes:
+   - "../../root.params"
+
+# Learner Configuration
+learner: simulated-integrated-learner-params
+beam_size: 10
```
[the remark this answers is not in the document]
sidharth-sundar: Got it! I've moved the parameter to corresponding m6 files.

adam/learner/contrastive_learner.py
sidharth-sundar: Ok, without altering the `_match_concept_pattern_to_multiple_graphs` function, I've extracted the values into a new variable and parsed the `None`s out beforehand.

adam/learner/__init__.py
```diff
@@ -322,10 +324,26 @@ def concepts_to_patterns(self) -> Dict[Concept, PerceptionGraphPattern]:
 
     @abstractmethod
     def propose_updated_hypotheses(
-        self, concept_to_updated_patterns: Dict[Concept, PerceptionGraphPattern]
+        self,
+        concept_to_updated_patterns: Dict[
```
spigo900, line 328: Just occurred to me but this should probably change now that the values are themselves mappings:

```suggestion
        concept_to_hypothesis_updates: Dict[
```

adam/learner/__init__.py
```diff
@@ -322,10 +324,26 @@ def concepts_to_patterns(self) -> Dict[Concept, PerceptionGraphPattern]:
...
+            Concept, Dict[PerceptionGraphTemplate, PerceptionGraphPattern]
+        ],
     ) -> None:
         """
         Propose new/updated hypotheses to the learner.
```
spigo900, line 334: A bit more documentation for good measure

```suggestion

    This expects hypothesis updates to be given as mappings of the form `old_hypothesis -> new_hypothesis` where `old_hypothesis` is a `PerceptionGraphTemplate` and `new_hypothesis` is a `PerceptionGraphPattern`.

```

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +251,78 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
-                    pattern_to_graph_match.pattern_node_to_matched_graph_node[
-                        pattern_node
-                    ]
-                    in difference_nodes
-                ):
-                    self._categorical_values_present_in_difference[
-                        concept, pattern_node.value
-                    ] += 1
+        for pattern_to_graph_match in pattern_to_graph_matches:
+            if pattern_to_graph_match:
```
spigo900, line 258: The `if` is no longer needed here since the matches are guaranteed non-`None`.

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +251,78 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
+                                pattern_node
+                            ]
+                            in difference_nodes
+                        ):
+                            self._ontology_node_present_in_difference[
+                                concept, pattern_node.property_value
+                            ] += 1
+                    # Otherwise, if it's categorical, count the value observed
+                    elif isinstance(pattern_node, CategoricalPredicate):
+                        self._categorical_values_present[concept, pattern_node.value] += 1
```
spigo900, line 292: Unrelated fix, but small enough that it seems worth including: Now that categoricals also have features, this should also index by `pattern_node.label`:

```suggestion
                        self._categorical_values_present[concept, pattern_node.label, pattern_node.value] += 1
```

This also requires updating the type annotation for `_categorical_values_present`.

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +251,78 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
-
-        for node, data in old_pattern_digraph.nodes(data=True):
-            old_to_new_node[node] = self._re_weighted_node(concept, node)
-            new_pattern_digraph.add_node(old_to_new_node[node], **data)
-
-        for old_u, old_v, data in old_pattern_digraph.edges(data=True):
-            new_pattern_digraph.add_edge(
-                old_to_new_node[old_u], old_to_new_node[old_v], **data
-            )
+        patterns = self.apprentice.concept_to_hypotheses(concept, self._top_n)
```
spigo900, line 305: I suggest renaming this `hypotheses` since now they're templates.

adam/learner/contrastive_learner.py
```diff
@@ -244,69 +251,78 @@ def log_hypotheses(self, log_output_path: Path) -> None:
...
+        updated_hypotheses = dict()
+        for pattern_template in patterns:
```
spigo900, line 307: for hypothesis in hypotheses

adam/learner/contrastive_learner.py
```diff
@@ -181,24 +190,22 @@ def learn_from(self, matching: LanguagePerceptionSemanticContrast) -> None:
...
-        # general. We would like to do something smarter. See the GitHub issue:
-        # TODO issue
-        return (
-            self._get_apprentice_pattern(concept)
-            .matcher(
+    ) -> Optional[
+        Mapping[PerceptionGraphTemplate, Optional[PerceptionGraphPatternMatch]]
+    ]:
+        return {
+            apprentice_concept: apprentice_concept.graph_pattern.matcher(
```
spigo900, line 199: Ah, missed this earlier -- please retain the previous comment about arbitrarily taking the first match and replace `TODO issue` with a link to #1140. That comment is about an issue unrelated to #1138, one that is more to do with how graph matching works.

sidharth-sundar: On this, I think the issue changes a bit when handling multiple hypotheses. If there is more than one hypothesis, then the first `top_n` are parsed. These `top_n` still face the same arbitrary issues as prior (taking `_concept_to_hypotheses[:top_n]`), instead of just `_concept_to_hypotheses[0]`)

spigo900: The problem isn't with which or how many arbitrary hypotheses we take. The problem is that the matching process itself is not deterministic. Specifically, we don't have any guarantees that two seemingly identical hypothesis won't be matched differently to the same perception, or the same hypothesis won't be matched differently to two (nearly) identical perceptions.

spigo900: Feel free to update the comment to make this more clear.

sidharth-sundar: Ah, got it.

spigo900 (approved):
Looks good, just one last typo. Also need to fix pre-commit but that should be simple enough. Other than the typo no further changes needed.

Also, change in process. I want you to rebase manually because I realized I was wrong about how GitHub does rebases -- it doesn't include a merge commit whereas manual rebase -> merge does. Please rebase this on master manually. Also, please simplify/clean up the GIt history (via interactive rebase). Once you've done that and you're ready to merge, ping me with a comment here saying you're ready to merge and I'll merge the pull request.

(Generally cleanup means getting rid of "noisy" commits like (a) merges, (b) commits fixing typos in previous commits in the same PR, (c) adding a default value to a method created in a previous commit, etc.)

If you haven't done interactive rebases before, here is some info. First, [link about interactive rebase in general](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d), and [link about interactive rebase in PyCharm](https://www.jetbrains.com/help/pycharm/edit-project-history.html#interactive-rebase). Also, some suggestions from experience:
1. if you get a conflict and you made multiple history changes, stop, abort, and switch to making just one change (move+maybe a fixup/squash) at a time.
2. It often causes conflicts when you change the relative order of commits to the same file, so I find it easier to do those reordering changes last when they are needed.
3. You may (I don't know) have a big commit you want to split up to separate different changes -- either to make the history more readable or because it makes reordering/squashing things easier. You can do that by marking the  for editing, then "undoing" the commit (on commandline, soft reset to previous), then committing each change separately.
    1. If you need to split up changes to a single file, you'll have to do interactive staging. You can do this with `git add -i [file]` on the command line or pressing `command+D`/`win+D` while the file is selected in the commit pane in PyCharm. (For PyCharm see [here](https://www.jetbrains.com/help/pycharm/comparing-file-versions.html#compare-modified-file). Also [here](https://www.jetbrains.com/help/pycharm/viewing-changes-information.html#comparing_local_changes) which is more specific -- has pictures. Note you use the checkboxes to pick which changes to commit.)

sidharth-sundar:
@spigo900 Once the travis check completes, this should be good to merge into `master`. I've rebased it onto the most recent version of master (immediately after #1141 was merged), so it's ready to merge.

PR #1146, opened 2022-06-14
1145 display for affordances

### LINKED ISSUE #1145: Output and display for affordances
We need to be able to display the affordances produced by #1111 and #1129 in the UI. Specifically, I think we want a list of text descriptions like 'can be SLOT2 in "SLOT1 throws SLOT2"', 'can be SLOT2 in "SLOT1 eats SLOT2"', et cetera. This requires some changes:

1. On the ADAM system side:
   1. To `IntegratedTemplateLearner`, generally in the area of `describe_common()`
   2. To `TopLevelLanguageLearnerDescribeReturn` to include the new features.
      1. Add a semantic node -> feature mapping giving the text descriptions, similar to what we have for features. The keys are objects and the values are sequences (lists/tuples) of descriptions for affordances that the object has.
      2. Populate this mapping in one of the helper methods for `describe_common()`.
      3. Generating the description output (as described at the top of this issue) requires knowing something about the template that the affordance comes from and the slot the affordance is focused on (SLOT2 in this case).
   4. To `YAMLLogger`, so that it outputs the affordances for an object. These should be minor, and similar to what's already there for (extracted visual) features.
3. On the UI side:
   1. (_No change: The Flask backend can stay the same, since it simply passes along the `output_language` key._)
   2. To the UI, so it displays the affordances. This is a few small changes -- the most annoying part is how to display things. See the section below.

See #1141 for the "mapping affordances" API used to get the affordances.

# Angular UI changes
Four small changes needed:

1. Purely for typing reasons: The interface/s in `selector-parent.component.ts`, specifically `LanguageResponse` needs a new attribute, say `derived_features`, which should be a new interface `DerivedFeatureResponse` that is basically the same as `FeatureResponse`.
    1. The actual code doesn't need to change at all since (like the Flask backend) we're just passing along the output lanugage.
2. `object-results.component.ts` needs to change to accommodate the new type of features, specifically `makeLinguisticObject`. This should be straightforward.
5. `object-results.component.html` needs to change to implement the display. This should be easy -- almost a copy of the "features" section.
   1. We do want slightly different formatting to distinguish the extracted features from the things (affordances) we inferred based on those extracted features. The details of the change should go in the CSS, though.
6. `object-results.component.css` probably should change to implement the formatting changes. (Italics? underline?).

The visual changes the least certain part. I think we will want to produce say four different demos of what the display could look like, review/discuss those, and go from there.

sidharth-sundar:
Closes #1145

Since #1141 handles adding affordances to the decode, I've added everything required from the angular side to display affordances in the front end.

### DIFF
--- a/angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
+++ b/angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
@@ -1,4 +1,5 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Affordances } from '../../classes/affordances';
 import { Features } from '../../classes/features';
 import { LinguisticOutput } from '../../classes/linguistic-output';
 
@@ -26,6 +27,23 @@ export class ObjectResultsComponent implements OnChanges {
       features.push(feat);
     });
 
+    const affordances: Affordances[] = [];
+    if (entry.affordances) {
+      entry.affordances.forEach((element) => {
+        const element_split = element.split('_');
+        let feat = { name: element };
+        // Formatting the affordances is already handled in the back end, so this is just here to be safe
+        if (element_split.length > 2) {
+          feat = {
+            name: `can be ${element_split[0]} in "${element_split
+              .slice(1)
+              .join(' ')}"`,
+          };
+        }
+        affordances.push(feat);
+      });
+    }
+
     const sub_objects: LinguisticOutput[] = [];
 
     if (entry.sub_objects) {
@@ -40,6 +58,7 @@ export class ObjectResultsComponent implements OnChanges {
       confidence: entry.confidence,
       type: entry.type,
       features: features,
+      affordances: affordances,
       sub_objects: sub_objects,
     };
   }
--- a/angular-viewer/adam-angular-demo/src/app/classes/affordances.spec.ts
+++ b/angular-viewer/adam-angular-demo/src/app/classes/affordances.spec.ts
@@ -1,0 +1,7 @@
+import { Affordances } from './affordances';
+
+describe('Affordances', () => {
+  it('should create an instance', () => {
+    expect(new Affordances()).toBeTruthy();
+  });
+});
--- a/angular-viewer/adam-angular-demo/src/app/classes/affordances.ts
+++ b/angular-viewer/adam-angular-demo/src/app/classes/affordances.ts
@@ -1,0 +1,3 @@
+export class Affordances {
+  name: string;
+}
--- a/angular-viewer/adam-angular-demo/src/app/classes/linguistic-output.ts
+++ b/angular-viewer/adam-angular-demo/src/app/classes/linguistic-output.ts
@@ -1,3 +1,4 @@
+import { Affordances } from './affordances';
 import { Features } from './features';
 
 export class LinguisticOutput {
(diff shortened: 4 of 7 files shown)


angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
```diff
@@ -26,6 +27,20 @@ export class ObjectResultsComponent implements OnChanges {
...
+    const affordances: Affordances[] = [];
+    if (entry.affordances) {
+      entry.affordances.forEach((element) => {
+        const element_split = element.split("_")
+        let feat = {name: element};
+        if (element_split.length == 4) {
+          feat = {
+            name: `can be ${element_split[0]} in "${element_split.slice(1).join(" ")}"`
+          };
+        }
```
spigo900, line 39: The processing logic might not be strictly needed here given it's already there in the output generation from #1141. Maybe after that is merged it is worth testing to see if we need this.

spigo900 (approved):
Looks good to me.

angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
sidharth-sundar: Right, that's what I figured as well after looking at the code Jacob wrote up. I don't think it would hurt too much performance-wise to keep in, but we can still check after that is merged.

spigo900: I'm less concerned with performance than "is this going to make sense to a reader." If I were to find this logic in both places I'd probably be confused and suspect I was wrong about where the data was coming from, because why would it be processed like this twice?

The other option is to stick a comment in here saying _in effect_ "yes, we already did this on the backend, but just to be safe."

angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
```diff
@@ -26,6 +27,22 @@ export class ObjectResultsComponent implements OnChanges {
...
+    if (entry.affordances) {
+      entry.affordances.forEach((element) => {
+        const element_split = element.split('_');
+        let feat = { name: element };
+        if (element_split.length === 4) {
+          feat = {
+            name: `can be ${element_split[0]} in "${element_split
+              .slice(1)
+              .join(' ')}"`,
+          }; // Formatting the affordances is already handled in the back end, so this is just here to be safe
```
spigo900, line 40: Nit, but I would rather put this before the statement -- either the line before the `feat = {` or on the line before the `if` -- I slightly prefer the latter but either works.

spigo900 (changes requested):
Looks good, one small nitpick.

angular-viewer/adam-angular-demo/src/app/components/object-results/object-results.component.ts
sidharth-sundar: I think putting it on the line before the `if` would look good, so I can move it there

sidharth-sundar: Done. I also realized that I had a strict equality check before:
`if (element_split.length === 4) {`
which would fail for affordances of the form `slot1_slot1_eats` or in an affordance descriptor with >2 slots, so I fixed that issue as well.

spigo900: Sounds good.

spigo900 (approved):
Looks good to me. Just needs rebase/squashing/Travis. Let me know when you've done this and you're ready to merge and I'll merge it.

PR #1149, opened 2022-06-16
1147 generate downsampled objects

### LINKED ISSUE #1147: Generate downsampled objects curriculum
We want to generate a downsampled version of the objects curriculum. Specifically, we want two variants: One with k=2 images per object, and one with k=10 images per object. (The M5 objects curriculum has k=30 per object.) We want a script to do this downsampling given the curriculum.

Some further details. I want this script to sample at random from the 30 images we have for each object, rather than taking the first k. The downsampling script should take the seed as a parameter. For now, sample from only those images where stroke extraction succeeded. Sample without replacement. If it turns out we don't have enough "good" samples to sample `k` of them without replacement for some objects, log a warning and take as many samples as we have.

I think we want to do input/output on the processed curriculum format for convenience. If we used the raw format we'd have to update the reorganizing script to handle potentially not having 30 images per object. That's doable but sounds like an annoying tangent.

# First step
@sidharth-sundar, as a first step, please write a script to count, for each object type, the number of images where stroke extraction succeeded. The output should be a two-column table, left being the object and right being the success count, e.g. `| apple | 30 |` or `| window | 11 |`. You can post that here or use a Google Sheet, whichever is more convenient. I want to know so we know how whether we have 10 good samples for each object.

I am also thinking we may want to do a "relative sample size preserving" variation on this experiment. We know the distribution is not uniform, so this table would tell us how badly we're mangling the distribution. I think these counts (together with the GNN results) would help us decide if a distribution-preserving experiment is worthwhile.

sidharth-sundar:
Closes #1147

Added two scripts: `count_stroke_samples.py` which counts the number of objects in an input curriculum which have stroke graphs, and `downsample_curriculum.py`, which takes as an input a preexisting reorganized curriculum and copies up to `k` samples per object to create a new training curriculum.

### DIFF
--- a/scripts/count_stroke_samples.py
+++ b/scripts/count_stroke_samples.py
@@ -1,0 +1,37 @@
+import argparse
+from pathlib import Path
+
+import yaml
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Utility to count number of samples with successful stroke extraction"
+    )
+    parser.add_argument("--input-dir", type=Path, help="An input curriculum directory to read objects from.")
+    args = parser.parse_args()
+    input_dir: Path = args.input_dir
+    object_to_stroke_count: dict = dict()
+
+    for feature_file_path in sorted(input_dir.glob("situation*")):
+        print(f"Working on file: {feature_file_path}")
+        with open(feature_file_path / 'feature.yaml', encoding="utf-8") as feature_file:
+            feature_yaml = yaml.safe_load(feature_file)
+        with open(feature_file_path / 'description.yaml', encoding="utf-8") as description_file:
+            description_yaml = yaml.safe_load(description_file)
+            target_object = description_yaml['language']
+        if 'stroke_graph' in feature_yaml.get('objects')[0]:
+            # Meant to check if stroke graph is in decode, but not really necessary since preprocess filters objects
+            # without stroke graph
+            object_to_stroke_count[target_object] = object_to_stroke_count.setdefault(target_object, 0) + 1
+        else:
+            print('missing stroke for', target_object)
+
+    spaces: int = max(len(obj) for obj in object_to_stroke_count) # Total amount of space needed for object name
+    sorted_object_to_stroke_count = sorted(object_to_stroke_count.items(), key=lambda x: x[1], reverse=True)
+    for obj, stroke_count in sorted_object_to_stroke_count:
+        print(f'|\t{obj}' + (spaces - len(obj)) * ' ' + f'\t|\t{stroke_count}\t|')
+
+
+if __name__ == "__main__":
+    main()
--- a/scripts/downsample_curriculum.py
+++ b/scripts/downsample_curriculum.py
@@ -1,0 +1,108 @@
+import argparse
+import logging
+from pathlib import Path
+from shutil import copytree, rmtree
+from typing import Mapping, Sequence
+from random import sample, seed
+
+import yaml
+
+situation_num: int = 0
+
+
+def get_situation_num(situation_dir: Path) -> int:
+    """Given a Path for a situation directory, return the associated number"""
+    return int(situation_dir.name.split('_')[-1])
+
+
+def copy_directory(start_path: Path, end_path: Path) -> None:
+    """Function to recursively copy directory from start_path to end_path"""
+    copytree(str(start_path), str(end_path))
+
+
+def copy_all_directories(output_dir: Path, dirs_to_copy: Sequence[Path]) -> None:
+    """Move all directories in dirs_to_move to their corresponding location in the target directory"""
+    global situation_num
+    sorted_dirs = sorted(dirs_to_copy, key=get_situation_num)
+    for directory in sorted_dirs:
+        copy_directory(directory, output_dir / f'situation_{situation_num}')
+        situation_num += 1
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Script to downsample preprocessed curricula"
+    )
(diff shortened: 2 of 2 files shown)


scripts/downsample_curriculum.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def copy_directory(start_path: Path, end_path: Path) -> None:
+    """Function to recursively copy directory from start_path to end_path"""
+    copytree(str(start_path), str(end_path))
+
+
+def copy_all_directories(output_dir: Path, dirs_to_move: Sequence[Path]):
+    """Move all directories in dirs_to_move to their corresponding location in the target directory"""
+    global situation_num
```
spigo900, line 26: Is there a reason `situation_num` has to be global here? It doesn't seem to be used anywhere else.

scripts/downsample_curriculum.py
```diff
@@ -0,0 +1,110 @@
...
+    """Given a Path for a situation directory, return the associated number"""
+    return int(str(situation_string).split('_')[-1])
+
+
+def copy_directory(start_path: Path, end_path: Path) -> None:
+    """Function to recursively copy directory from start_path to end_path"""
+    copytree(str(start_path), str(end_path))
+
+
+def copy_all_directories(output_dir: Path, dirs_to_move: Sequence[Path]):
```
spigo900, line 24: Slight name change:

```suggestion
def copy_all_directories(output_dir: Path, dirs_to_copy: Sequence[Path]):
```

scripts/downsample_curriculum.py
```diff
@@ -0,0 +1,110 @@
...
+        logging.info(f"Now processing {obj}")
+        if num_samples > len(potential_samples):
+            logging.warning(f"Attempting to retrieve more samples than exist for {obj}. Default to using all samples")
+            copy_all_directories(output_dir, potential_samples)
+            selected_sample_count += len(potential_samples)
+
+        else:
+            selected_samples: Sequence[Path] = sample(potential_samples, k=num_samples)
+            copy_all_directories(output_dir, selected_samples)
+            selected_sample_count += num_samples
```
spigo900, line 100: Since the two endling lines are virtually identical between the two and that's unlikely to change, I'd prefer to factor them out of the if/else and do `selected_samples = potential_samples` in the `num_samples > len(potential_samples)` case. Then you can just always do:
```python
copy_all_directories(output_dir, selected_samples)
selected_sample_count += len(selected_samples)
```
This makes the code a little easier to follow since it's explicit that both branches have this same logic.

spigo900 (changes requested):
Looks good overall. A few style/clarity changes.

scripts/count_stroke_samples.py
```diff
@@ -0,0 +1,37 @@
...
+        print(f"Working on file: {feature_file_path}")
+        with open(feature_file_path / 'feature.yaml', encoding="utf-8") as feature_file:
+            feature_yaml = yaml.safe_load(feature_file)
+        with open(feature_file_path / 'description.yaml', encoding="utf-8") as description_file:
+            description_yaml = yaml.safe_load(description_file)
+            target_object = description_yaml['language']
+        if 'stroke_graph' in feature_yaml.get('objects')[0]:
+            # Meant to check if stroke graph is in decode, but not really necessary since preprocess filters objects
+            # without stroke graph
+            object_to_stroke_count[target_object] = object_to_stroke_count.setdefault(target_object, 0) + 1
```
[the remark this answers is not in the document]
spigo900: ETA: nope, I was wrong, in the raw curriculum the feature file is absent when there are no strokes.

Anyway, no change needed. I think this is fine.

scripts/count_stroke_samples.py
```diff
@@ -0,0 +1,37 @@
...
+        else:
+            print('missing stroke for', target_object)
+
+    spaces: int = max(len(obj) for obj in object_to_stroke_count) + 4
```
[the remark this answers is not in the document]
sidharth-sundar: Ah, I was using the 4 as padding for the output format, since there are approximately 4 spaces in a tab in most contexts, but I think it works better/looks better if we combine the suggestion to alter the `+ 4` and to add the tab before the second `|`:

`spaces: int = max(len(obj) for obj in object_to_stroke_count) # Total amount of space needed for object name`
`sorted_object_to_stroke_count = sorted(object_to_stroke_count.items(), key=lambda x: x[1], reverse=True)`
`for obj, stroke_count in sorted_object_to_stroke_count:`
        `print(f'|\t{obj}' + (spaces - len(obj)) * ' ' + f'\t|\t{stroke_count}\t|')`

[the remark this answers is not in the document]
spigo900: That works too.

spigo900 (changes requested):
Just the one nit about path methods.

scripts/downsample_curriculum.py
```diff
@@ -0,0 +1,108 @@
...
+            raise FileExistsError(str(output_dir))
+
+    if num_samples < 0:
+        logging.warning("Attempting to retrieve a negative number of samples")
+        raise ValueError(num_samples)
+
+    if random_seed:
+        seed(random_seed)
+
+    object_to_target_directories: Mapping[str, Sequence[Path]] = dict()
```
spigo900, line 79: I think this should be called `object_to_potential_samples` for clarity. `target` sounds like we're copying to them.

spigo900 (changes requested):
One last nit, otherwise I think this looks good.

spigo900 (approved):
Looks good to me.

PR #1150, opened 2022-06-17
Add script to generate unknown object sample

### LINKED ISSUE #1148: Generate unknown objects rehearsal curriculum
We want to generate a version of the objects curriculum to do an "unknown objects rehearsal." Specifically, we want to generate a version of the M5 objects train curriculum where we drop all samples of the 5 objects that @sidharth-sundar picked. We want a script to generate this alternative train curriculum given the M5 objects curriculum.

The 5 objects were:

* `apple`
* `ball`
* `cube_block`
* `chair`
* `window` (the more unique object)

As with #1147, I think for convenience we want this script to accept input/generate output in the processed curriculum format.

sidharth-sundar:
Closes #1148

Add script generate_unknown_objects.py, which parses a reorganized curriculum and removes all objects specified as "unknown objects".


scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,105 @@
...
+    """Given a Path for a situation directory, return the associated number"""
+    return int(str(situation_string).split('_')[-1])
+
+
+def copy_directory(start_path: Path, end_path: Path) -> None:
+    """Function to recursively copy directory from start_path to end_path"""
+    copytree(str(start_path), str(end_path))
+
+
+def copy_all_directories(output_dir: Path, dirs_to_move: Sequence[Path]):
```
spigo900, line 23:
```suggestion
def copy_all_directories(output_dir: Path, dirs_to_copy: Sequence[Path]):
```

scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Mapping, Sequence
+
+import yaml
+
+situation_num: int = 0
+
+
+def get_situation_num(situation_string: Path) -> int:
+    """Given a Path for a situation directory, return the associated number"""
+    return int(str(situation_string).split('_')[-1])
```
spigo900, line 15: Same comment as on the other PR -- rename and prefer path.name.

scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,105 @@
...
+        situation_num += 1
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Utility to filter out specified objects from curriculum"
+    )
+    parser.add_argument('--input-dir', type=Path, help="Directory with curriculum training samples")
+    parser.add_argument('--output-dir', type=Path, help="Directory to write remaining samples to")
+    parser.add_argument("--curriculum-name", required=False, type=str, help="Name of training curriculum")
```
spigo900, line 38: I don't think it's necessary to specify "training" here. It doesn't have to be. It's just that that's all we're using it for right now.

spigo900 (changes requested):
Looks good, just a few small comments.

scripts/generate_unknown_objects.py
spigo900: I think it is confusing to specify `training` curriculum here, because there's no reason that the output must be used for training. I'd rather call it `Name of output curriculum`.

sidharth-sundar: Ok, I can update that.

scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,105 @@
...
+
+    input_dir: Path = args.input_dir
+    output_dir: Path = args.output_dir
+    force_overwrite: bool = args.force
+    curriculum_name: str = args.curriculum_name
+    unknown_objects: Sequence[str] = args.unknown_objects
+    if not curriculum_name:
+        curriculum_name = output_dir.name
+
+    if not isdir(str(input_dir)):
```
spigo900, line 65: Didn't notice this before, but here you can use `input_dir.exists()`. If you really want to check if it's a directory you can separately check `input_dir.is_dir()`.

scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,105 @@
...
+        logging.warning("Input directory does not exist")
+        raise FileExistsError(str(input_dir))
+
+    if isdir(str(output_dir)):
```
spigo900, line 69: Similar comment as above, although here maybe you do want to check if it's a directory based on the warning.

spigo900 (changes requested):
Looks good, some nits about using path methods.

scripts/generate_unknown_objects.py
spigo900: Good to drop the import.

scripts/generate_unknown_objects.py
sidharth-sundar: Right, it's currently set up to create an output directory if one doesn't already exist, and it looks like `Path.is_dir()` is `False` if the directory doesn't exist. Also realized here I raised a `FileExistsError` instead of a `FileNotFoundError`, so I fixed that.

scripts/generate_unknown_objects.py
```diff
@@ -0,0 +1,104 @@
...
+        raise FileNotFoundError(str(input_dir))
+
+    if output_dir.is_dir():
+        if force_overwrite:
+            rmtree(str(output_dir))
+        else:
+            logging.warning("There already exists a directory in the target location")
+            raise FileExistsError(str(output_dir))
+
+    object_to_target_directories: Mapping[str, Sequence[Path]] = dict()
```
spigo900, line 75: Ah, just noticed this should be `object_to_sample_directories`. `target` sounds like we're copying from but we're copying to.

spigo900 (changes requested):

scripts/generate_unknown_objects.py
sidharth-sundar: Ok, I've changed the variable name.

spigo900 (approved):
Looks good to me.

Clean up the history, rebase, then let me know when you're ready to merge.

PR #1151, opened 2022-06-17
GNN module

spigo900:
This seems to be working. I'm planning to clean up the history and probably the code more the next working day.

I've tried to keep the code mostly as Sheng wrote it although I've renamed some variables and of course the data loading logic is entirely different.

For some reason PyCharm thinks`torch` and `numpy` are both unused in these scripts... :shrug:


spigo900:
This should be ready for review now. I've cleaned up the history. I've also blackened the sections of the two scripts that I changed. I left the rest as-is.

I haven't set up the Makefile to run checks on this directory given the different style of the original code. I did manually run some checks. `flake8` and `pylint` are mostly happy except for some quirks preserved from the original scripts. Mypy is grumpy about `ndarray` not having type parameters -- it looks like the second should be a dtype but I haven't yet found an explanation for the first parameter.

lichtefeld (changes requested):
I'm assuming the places in the code where strings are used as what looks like comments is carry-over from an iPython file? I don't know if its worth the time to adjust those to comments or not given they really shouldn't impact runtime.

A few small requests!

adam_preprocessing/shape_stroke_graph_inference.py
```diff
@@ -0,0 +1,179 @@
...
+    logging.info("Done loading data.")
+
+    device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+    # data = from_pickle('./data.pkl')
+    # train_coords = data['train_coords']
+    # train_adj = data['train_adj']
+    # train_label = data['train_label']
+    # test_coords = data['test_coords']
+    # test_adj = data['test_adj']
+    # test_label = data['test_label']
```
sidharth-sundar, line 70: If these lines are no longer being used, we can delete the dead code.

adam_preprocessing/utils.py
```diff
@@ -0,0 +1,254 @@
...
+                distance_map += np.square(stroke1_2).repeat(n, 1) + np.square(stroke2_2.T).repeat(n,
+                                                                                                  0) - 2 * np.dot(
+                    stroke1_2, stroke2_2.T)
+
+                distance_map += 1e-13
+                distance_map = np.sqrt(distance_map)
+                distance_map = distance_map.flatten()
+
+                e[(ee[0][jj], ee[1][jj])] = distance_map
+                # e[(ee[0][jj], ee[1][jj])] = np.concatenate([distance_map, angle])
```
sidharth-sundar, line 252: Dead code

adam_preprocessing/utils.py
```diff
@@ -0,0 +1,254 @@
...
+        super(LinearModel, self).__init__()
+        self.model = model
+        self.linear = nn.Linear(in_features=out, out_features=target)
+        torch.nn.init.eye_(self.linear.weight)
+        self.feature = None
+
+    def forward(self, g, h, e):
+        x = self.model(g, h, e)
+        self.feature = x
+        # x = nn.ReLU()(x)
```
sidharth-sundar, line 178: Dead code

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -0,0 +1,311 @@
...
+                      'w') as file:
+                documents = yaml.dump(data, file)
+                file.close()
+        return data
+
+
+def main():
+    for i in range(3):
+        for j in range(5):
+            S = Stroke_Extraction(base_path='/home/scheng53/Desktop/darpa_3d/adam_single_objv0', obj_type='test_small_single_mug', obj_view=str(i), obj_id=str(j))
```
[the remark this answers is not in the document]
spigo900: At least I assume we don't -- but perhaps it's actually there somewhere

sidharth-sundar (changes requested):
Looks good for the most part, the thing I was going to comment on was the README but @lichtefeld already took care of that. The other thing I noticed was some dead code, so I've pointed it out wherever I found it.

adam_preprocessing/README.md
```diff
@@ -0,0 +1,61 @@
+# ADAM visual preprocessing integrations
+This is the code for ADAM's visual preprocessing integrations. For the moment we have only integrated the object stroke GNN. This is based on ASU's code [here][asu_gnn].
+
+[asu_gnn]: https://github.com/ASU-APG/adam-stage/tree/main/processing
+
+# Setup
+1. Create and activate a Python 3.10 Anaconda environment (or your favorite other means of creating a virtual environment): `conda create --name adam python=3.7`
```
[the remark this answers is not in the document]
spigo900: Hmm, I swear I fixed this but seems not. I pushed a fix.

adam_preprocessing/shape_stroke_extraction.py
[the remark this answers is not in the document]
spigo900: True. I can parameterize it.

spigo900:
I've tested and it seems the stroke extraction code is working. Results are inconsistent with Sheng's, however I've reproduced that inconsistency using the original code so this inconsistency is not due to any of the changes I've made. So I think this code is ready for review.

I am separately checking to see if I get consistent results when using the same Matlab version Sheng used. That shouldn't affect this pull request however.

lichtefeld (approved):
I've given the files a look over, It looks like licenses are appropriately copied and Sheng's been given credit where applicable. I'm not going to do any deeper review on this unless @spigo900 feels there's specific files I should look at.

adam_preprocessing/predict.sh
```diff
@@ -0,0 +1,29 @@
...
+  python shape_stroke_graph_inference.py --help
+  exit 1
+else
+  model_path=$1
+  input_curriculum_dir=$2
+  output_curriculum_dir=$3
+  shift 3
+fi
+
+python shape_stroke_graph_inference.py "$model_path" "$input_curriculum_dir" --save_outputs_to "$output_curriculum_dir" "$@"
```
sidharth-sundar, line 29: Out of curiosity, after shifting over by 3, what is the `"$@"` used for here? Earlier, you check to make sure there are exactly 3 inputs to the script (`if [[ "$#" -ne 3 ]]`), so even if there are further arguments provided, the script would just terminate instead of adding the extra arguments to the end. It also looks like in `shape_stroke_graph_inference.py`, there are only 3 specified command line arguments anyway, so it makes sense to me that you only check for 3 in the script, just that the `"$@` wouldn't really do anything as is.

sidharth-sundar (approved):

adam_preprocessing/predict.sh
spigo900: The `"$@"` is as you thought not actually doing anything here. I think it came over out of habit or because I copied the setup from a Slurm script for another project where I use `-lt`.

I think it's good to have it there as a reminder in case we decide to add more args. But you're right, it's a confusing thing to include. I've added comments make this explicit.

spigo900:
Travis seems to have lost track of this PR, so if no one objects by 11, I'm going to force-merge it. It passed pre-commit last time Travis ran and the only changes were to the README and to add some comments so it should still.

PR #1153, opened 2022-06-22
Bug fixes for observational affordance learner.

lichtefeld:
Adds a new affordance curriculum type from the m5 actions curriculum & fixes a few bugs from the observational affordance learner.

### DIFF
--- a/adam/learner/perception_graph_template.py
+++ b/adam/learner/perception_graph_template.py
@@ -278,7 +278,15 @@ class PerceptionGraphTemplate:
         """
         return PerceptionGraphTemplate(
             graph_pattern=self.graph_pattern.copy_replacing_nodes(current_to_new_node),
-            template_variable_to_pattern_node=self.template_variable_to_pattern_node,
+            template_variable_to_pattern_node=immutabledict(
+                (variable, current_to_new_node.get(node, node))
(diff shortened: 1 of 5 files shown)


adam/learner/perception_graph_template.py
```diff
@@ -278,7 +278,11 @@ def copy_replacing_nodes(
         """
         return PerceptionGraphTemplate(
             graph_pattern=self.graph_pattern.copy_replacing_nodes(current_to_new_node),
-            template_variable_to_pattern_node=self.template_variable_to_pattern_node,
+            template_variable_to_pattern_node=immutabledict(
+                (variable, node)
+                for variable, node in self.template_variable_to_pattern_node.items()
+                if node not in current_to_new_node
```
spigo900, line 284: So this omits replaced nodes from the template? Wouldn't we want to replace them?

spigo900 (commented):
Looks good. Left one comment about something that confused me.

adam/learner/perception_graph_template.py
lichtefeld: I don't think I can get away with an easy replacement as the node I'll replace too will change into a new semantic node when the object learners process and we'll be unable to perform the update to the affordance semantic node when that occurs.

Issue #1155, opened 2022-06-28
Online Decode Pipeline for Objects


lichtefeld:
The following is my current outline for the needed components that need to be implemented for a decode pipeline. First I've listed how I think the live demo should be structured for runtime and then below each needed component has a heading for further details.

## Demo Outline
The input is a PNG image, this needs to have some form of segmentation applied and then stroke preprocessing needs to occur. To prevent needing to make multiple command line calls in the live demo my thought in the script for the demo should just invoke a subprocess for preprocessing and wait for it to complete before continuing on to ADAM decode. Once stroke preprocessing has occurred the scene can just be loaded into ADAM and decoded as normal into a specific directory for live demos. Once the 'complete' (or similar message) has appeared in the terminal I'll just refresh the available scenes in the UI and select the newly decoded scene. 

### Demo Script

Ideally this demo runtime environment can load a pickled ADAM learner for decoding so we can train a learner offline and not need to run any training prior to the demo. This would allow us to potentially invoke the entire pipeline for each image to be processed but I think being able to load the learner configuration once and then pass in files to process would be better for a demo. This requires:

- [x] Validating the new learners can be pickled and unpickled successfully (In theory working, but untested)
- [x] Implementing the entry point to load a learner and handle file path inputs from the command line
  - [x] Basic validation for invalid data
  - [x] Ability to invoke all preprocessing either completely in python or split across python/matlab
  - [x] Run of ADAM decode | Bonus: Ability to change pretrained learner
  - [x] Bonus if this is driven as a server with API calls by the UI... but this is truly a bonus goal without file upload through the UI.
- [x] A configuration with the UI that can detect new available data for display

### Preprocessing
The preprocessing work is in progress for integration by @spigo900. An ideal goal is a single python script entry point for preprocessing an individual image rather than an entire curriculum at once. 

#### Segmentation
@spigo900 Noted that the current preprocessing requires a segmentation file. We won't have that from just a PNG so how could we acquire that in the easiest way? Alternatively, can stroke extract run directly on the PNG images with success.

### Decode
We'll need to have a trained ADAM learner for live decode. It may be nice to have models trained with different amounts of data to investigate any reasonable differences in decode output but mostly I think we should use the model with essentially all available training data produced used for decode. 

### Testing
I plan to take ~3-5 images of a chair & table from my apartment for some basic testing images. I can look through the other objects we have to see what else I could easily acquire an image of. This will hopefully allow us to test this implementation well before the live demo. 

[...]

PR #1161, opened 2022-07-26
Frontend live updates

sidharth-sundar:
Closes the live image display portion of #1159 

There are 2 major changes present in this update:

- Add functionality to send image files from directory exterior to angular working directory:
  - Rather than just providing image paths, this encodes the image in a base64 string in the flask back end and sends the embedded image object in response to an HTTP GET request. 
  - Then the image is decoded and displayed in the front end.
  - Also means assets do not need to be cloned/duplicated into the Angular working directory at build time
 
- Add polling service to display changes in back end
   - The app stores a trace of the available data when the app is initiated
   - Every 10 seconds, the app sends a GET request to the back end to see if any new data is present (can be manually adjusted)
   - Once the user submits a request with a specific learner, training curriculum, and testing curriculum, those values are stored in the Angular context
   - Rather than checking if there are ANY changes, we restrict our view to the curriculum the user is currently observing.

I can make some additional improvements around these features, such as presenting the notification of new data more cleanly/more gracefully handling errors with image retrieval.

### DIFF
--- a/adam/flask_backend/app.py
+++ b/adam/flask_backend/app.py
@@ -20,6 +20,7 @@ from adam.paths import (
     TESTING_CURRICULUM_DIR,
     is_relative_to,
 )
+from utils import get_image_data
 
 app = Flask(__name__)
 cors = CORS(app)
@@ -140,6 +141,67 @@ def get_scene() -> Any:
         #     "language_with_scene": "a chair",
         # }
     }
+
+
+@app.route("/api/get_image", methods=["GET"])
+def get_image() -> Any:
+    if not request.args:
+        abort(HTTPStatus.BAD_REQUEST)
+
+    learner = request.args.get("learner", default="")
+    training_curriculum = request.args.get("training_curriculum", default="")
+    testing_curriculum = request.args.get("testing_curriculum", default="")
+    scene_number = int(request.args.get("scene_number", "")) - 1
+    situation_dir = f"situation_{scene_number}"
+    image_file = request.args.get("image_file", default="")
+    if (
+        not learner
+        or not training_curriculum
+        or not testing_curriculum
+        or not situation_dir
+    ):
+        abort(HTTPStatus.BAD_REQUEST)
+
+    if any(
+        "/" in arg
+        for arg in (
+            learner,
+            training_curriculum,
+            testing_curriculum,
+            situation_dir,
+            image_file,
+        )
+    ):
+        abort(HTTPStatus.BAD_REQUEST)
+    image_path = (
+        LEARNERS_DIR
+        / learner
+        / EXPERIMENTS_DIR_NAME
+        / training_curriculum
+        / EXPERIMENTS_TESTING_DIR_NAME
+        / testing_curriculum
+        / situation_dir
+        / image_file
+    ).resolve()
+    if not is_relative_to(image_path, DATA_DIR):
+        return {"message": "Directory out of range"}, HTTPStatus.BAD_REQUEST
+
+    if (
+        learner not in app.config["LEARNER_DIR_CONTENTS"]
+        or training_curriculum not in app.config["LEARNER_DIR_CONTENTS"][learner]
+        or testing_curriculum
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum]
+        or situation_dir
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum][
+            testing_curriculum
+        ]
+        or image_file
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum][
+            testing_curriculum
+        ][situation_dir]
+    ):
+        return {"message": "Invalid file"}, HTTPStatus.BAD_REQUEST
+    return {"image_data": get_image_data(image_path)}
 
 
 if __name__ == "__main__":
--- a/adam/flask_backend/app.py
+++ b/adam/flask_backend/app.py
@@ -1,6 +1,5 @@
 from glob import glob
 from http import HTTPStatus
-from os.path import relpath
 from typing import Any
 
 import yaml
@@ -119,20 +118,21 @@ def get_scene() -> Any:
         "test_curriculum": testing_curriculum,
         "scene_number": scene_number,
         "scene_images": [
-            relpath(path, DATA_DIR)
+            get_image_data(path)
             for path in sorted(
                 chain(
                     glob(f"{experiment_dir}/rgb__[0-9]*.png"),
+                    glob(f"{experiment_dir}/rgb_[0-9]*.png"),
                     glob(f"{experiment_dir}/id_rgb_[0-9]*.png"),
                 )
             )
         ],
         "object_strokes": [
-            relpath(path, DATA_DIR)
+            get_image_data(path)
             for path in sorted(glob(f"{experiment_dir}/stroke_[0-9]*_[0-9]*.png"))
         ],
         "stroke_graph": [
-            relpath(path, DATA_DIR)
+            get_image_data(path)
             for path in sorted(glob(f"{experiment_dir}/stroke_graph_*.png"))
         ],
         "post_learning": post_learn,
--- a/adam/flask_backend/app.py
+++ b/adam/flask_backend/app.py
@@ -1,5 +1,6 @@
 from glob import glob
 from http import HTTPStatus
+from os.path import relpath
 from typing import Any
 
 import yaml
@@ -9,6 +10,7 @@ from flask import Flask, abort, request
 from flask_cors import CORS
 
 from adam.paths import (
+    DATA_DIR,
     LEARNERS_DIR,
     SITUATION_DIR_NAME,
     POST_LEARN_FILE_NAME,
@@ -16,6 +18,7 @@ from adam.paths import (
     EXPERIMENTS_TESTING_DIR_NAME,
     TRAINING_CURRICULUM_DIR,
     TESTING_CURRICULUM_DIR,
+    is_relative_to,
 )
 
 app = Flask(__name__)
@@ -79,10 +82,12 @@ def get_scene() -> Any:
     scene_number = int(request.args.get("scene_number", "")) - 1
     if (
         not learner
-        and not training_curriculum
-        and not testing_curriculum
-        and not scene_number
+        or not training_curriculum
+        or not testing_curriculum
+        or not scene_number
     ):
+        abort(HTTPStatus.BAD_REQUEST)
+    if any("/" in arg for arg in (learner, training_curriculum, testing_curriculum)):
         abort(HTTPStatus.BAD_REQUEST)
 
     experiment_dir = (
@@ -93,7 +98,9 @@ def get_scene() -> Any:
         / EXPERIMENTS_TESTING_DIR_NAME
         / testing_curriculum
         / SITUATION_DIR_NAME.format(num=scene_number)
-    )
+    ).resolve()
+    if not is_relative_to(experiment_dir, LEARNERS_DIR):
+        return {"message": "Directory out of range"}, HTTPStatus.BAD_REQUEST
     if not experiment_dir.exists():
         return {"message": "Selected configuration does not exist"}
 
@@ -110,7 +117,7 @@ def get_scene() -> Any:
         "test_curriculum": testing_curriculum,
         "scene_number": scene_number,
         "scene_images": [
-            path.rsplit("data", maxsplit=1)[-1]
+            relpath(path, DATA_DIR)
             for path in sorted(
                 chain(
                     glob(f"{experiment_dir}/rgb__[0-9]*.png"),
@@ -119,11 +126,11 @@ def get_scene() -> Any:
             )
         ],
         "object_strokes": [
-            path.rsplit("data", maxsplit=1)[-1]
+            relpath(path, DATA_DIR)
             for path in sorted(glob(f"{experiment_dir}/stroke_[0-9]*_[0-9]*.png"))
         ],
         "stroke_graph": [
-            path.rsplit("data", maxsplit=1)[-1]
+            relpath(path, DATA_DIR)
             for path in sorted(glob(f"{experiment_dir}/stroke_graph_*.png"))
         ],
         "post_learning": post_learn,
--- a/adam/flask_backend/app.py
+++ b/adam/flask_backend/app.py
@@ -20,12 +20,13 @@ from adam.paths import (
     TESTING_CURRICULUM_DIR,
     is_relative_to,
 )
-from utils import get_image_data
+from utils import get_image_data, retrieve_relevant_files
 
 app = Flask(__name__)
 cors = CORS(app)
 app.config["CORS_HEADERS"] = "Content-Type"
 app.config["JSON_SORT_KEYS"] = False
+app.config["LEARNER_DIR_CONTENTS"] = {}
 
 
 @app.route("/api/learners", methods=["GET"])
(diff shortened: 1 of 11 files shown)


spigo900 (changes requested):
This mostly looks good to me, just a few change requests. My main comments on the Python code are security concerns about paths. I see you've noticed one of those. There are a few nits. The TypeScript code looks good to me.

adam/flask_backend/app.py
```diff
@@ -135,7 +138,104 @@ def get_scene() -> Any:
...
+        abort(HTTPStatus.BAD_REQUEST)
+
+    curriculum_dir = (
+        LEARNERS_DIR
+        / learner
+        / EXPERIMENTS_DIR_NAME
+        / training_curriculum
+        / EXPERIMENTS_TESTING_DIR_NAME
+        / testing_curriculum
+    )
```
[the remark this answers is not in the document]
sidharth-sundar: I believe on most filesystems on unix-based devices, `/` is prohibited in file names, but there are no additional restrictions, so `.` can still appear. It's not advised and definitely bad practice, but I think we can be slightly looser on it. We can enforce that `/` does not appear in any of the input parameters, and then we can check that the resulting curriculum_dir is a relative file path to the LEARNERS_DIR using `curriculum_dir.is_relative_to(LEARNERS_DIR).

adam/flask_backend/app.py
```diff
@@ -135,7 +138,104 @@ def get_scene() -> Any:
     }
 
 
+@app.route("/api/get_image", methods=["GET"])
+def get_image() -> Any:
+    if not request.args:
+        abort(HTTPStatus.BAD_REQUEST)
+
+    image_path = request.args.get("image_path", default="")
```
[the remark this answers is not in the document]
sidharth-sundar: Ah, I was looking for some way to enforce that the resolved path is inside our target directory, but I didn't know that pathlib had built in functionality for it

adam/flask_backend/app.py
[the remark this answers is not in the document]
spigo900: But we can enforce no "." because they're not just filenames. They're filenames from a very limited set of possible names specifically defined by ADAM.

adam/flask_backend/app.py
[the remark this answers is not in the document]
spigo900: Makes sense.

adam/flask_backend/app.py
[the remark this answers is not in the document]
sidharth-sundar: Ah right, because the curriculum names are restriced based on what we manually allow ADAM to process. I thought for some reason the output directories could be separately configured in the parameter file.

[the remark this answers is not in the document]
spigo900: I see, that would make sense.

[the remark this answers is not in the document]
spigo900: In fact you're right, it is allowed to place the directories anywhere. I'm sorry. So I agree, forbid `/`s, permit `.`s as long as it resolves to the data directory.

adam/flask_backend/app.py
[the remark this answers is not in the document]
sidharth-sundar: Well we have a bit of a hiccup. `is_relative_to` is implemented in Python 3.9. We could still do some string processing and see if the `DATA_DIR` absolute path is entirely contained in the `DATA_DIR / image_path` absolute path, but it's a bit messier.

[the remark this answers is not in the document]
spigo900: Hmm, that is unfortunate. I think something like this should work as a shim, though this probably needs more checking:

```
def is_relative_to(path: Path, base_path: Path) -> bool:
    """
    Return whether `path` is relative to `base path`.

    Shim for Py3.9's `.is_relative_to()`.
    """
    try:
        path.relative_to(base_path)
        return True
    else:
        return False
```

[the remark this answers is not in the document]
spigo900: Ah whoops, that should be `except ValueError`, not `else`. Was thinking of if-statements. 🙃 

Anyway, this is [more or less the actual definition](https://github.com/python/cpython/blob/f118661a18f9e0cc692bfec226f38c26f5fe0558/Lib/pathlib.py#L824-L831) -- they allow variable arguments but that's not needed here. So I think this is good enough for a shim.

adam/flask_backend/app.py
```diff
@@ -135,7 +143,116 @@ def get_scene() -> Any:
...
+                    f"New decode available for situation {int(situation_dir.name.rsplit('_')[-1]) + 1}"
+                )
+                ref_experiments[situation_dir.name] = set(
+                    file.name for file in situation_dir.iterdir()
+                )
+            else:
+                for file in situation_dir.iterdir():
+                    if file.name not in ref_experiments[situation_dir.name]:
+                        update_message.append(
+                            f"Updated information available for {int(situation_dir.name.rsplit('_')[-1]) + 1}: {file.name}"
```
spigo900, line 213: Missing a word here:
```suggestion
                            f"Updated information available for situation {int(situation_dir.name.rsplit('_')[-1]) + 1}: {file.name}"
```

spigo900 (changes requested):
Thanks for the changes. And thanks for patching get_scene() -- I hadn't thought of that having issues, but it obviously did. Just a typo this round.

adam/flask_backend/app.py
```diff
@@ -135,7 +143,116 @@ def get_scene() -> Any:
...
+@app.route("/api/get_image", methods=["GET"])
+def get_image() -> Any:
+    if not request.args:
+        abort(HTTPStatus.BAD_REQUEST)
+
+    image_path = request.args.get("image_path", default="").strip("/")
+    image_file = DATA_DIR.joinpath(image_path).resolve()
+    if not is_relative_to(image_file, DATA_DIR):
+        abort(HTTPStatus.BAD_REQUEST)
+    print(image_file)
```
spigo900, line 155: Debugging?

sidharth-sundar: Debugging indeed - good catch!

adam/flask_backend/app.py
```diff
@@ -135,7 +143,116 @@ def get_scene() -> Any:
...
+                ref_experiments[situation_dir.name] = set(
+                    file.name for file in situation_dir.iterdir()
+                )
+            else:
+                for file in situation_dir.iterdir():
+                    if file.name not in ref_experiments[situation_dir.name]:
+                        update_message.append(
+                            f"Updated information available for {int(situation_dir.name.rsplit('_')[-1]) + 1}: {file.name}"
+                        )
+                        ref_experiments[situation_dir.name].add(file.name)
```
sidharth-sundar: A quick note here. Currently, the polling function can return one of 2 options depending on the new information available. First, if there is an entirely new directory available, the function alerts the user that a new situation decode is available. Second, if a preexisting situation directory contains new files, the user is alerted to each individual new file available. This doesn't occur too often, but when it does, there is massive information overload (consider an action experiment test case, for instance, which has several camera views/time points where images are provided). I was considering one of four workarounds to improve visibility/clarity. 

1. Rather than provide new information, specify only the situation that is updated. 
2. Add a utility function to truncate the return info (not preferred because if multiple new decodes are available, some necessary information would be dropped)
3. With more entries, decrease the font size of the displayed results
4. Restrict the files considered based on our naming convention. The front end only cares about a handful of the files available, specifically
   - The id_rgb images (or just rgb if no id_rgb is available)
   - The stroke image
   - The stroke graph
   - The object decode

Thoughts? @spigo900 @lichtefeld

adam/flask_backend/app.py
```diff
@@ -93,7 +99,9 @@ def get_scene() -> Any:
         / EXPERIMENTS_TESTING_DIR_NAME
         / testing_curriculum
         / SITUATION_DIR_NAME.format(num=scene_number)
-    )
+    ).resolve()
+    if not is_relative_to(experiment_dir, LEARNERS_DIR):
+        return {"message": "Directory out of range"}
```
sidharth-sundar, line 104: When the user tries to retrieve data in an out-of-range directory, should we always err out with `HTTPStatus.BAD_REQUEST`? Currently, this response is inconsistent with the response in `get_image()

spigo900: I think the different responses make sense. In this case we have more specific/restricted parameters so that we know the user is requesting info on a particular situation. With `get_image()` the user gives an arbitrary path within `DATA_DIR`, so it is harder to give a useful error message. It would make sense to return an error code as well rather than 200 OK -- [I think](https://flask.palletsprojects.com/en/1.1.x/quickstart/#about-responses) the syntax would be:

```python
return {"message": "Directory out of range"}, HTTPStatus.BAD_REQUEST
```

sidharth-sundar:
Two structural comments I wanted to bring up while I remembered them
1. @spigo900 mentioned that we can be stricter with how we have users retrieve files, more in line with the newly implemented `poll_changes()` function or the `get_scene()` function. One way we can do this is by setting app-specific environment variables on initiation. When `get_all_learners()`, `get_all_train_curriculum()`, and `get_all_test_curriculum()` are initially invoked, we can maintain a list of learners and curricula, and can later enforce that the user-provided learners/curricula are among the selected ones.
2. I initially implemented `get_image` as an API call because it made sense to me for image retrieval to take place independently; however, with the current approach, selecting a scene generates 1 GET request for retrieving image paths, and 1 GET request per individual image. We could instead wrap all of this functionality into a single request, where the `get_scene` call in Flask invokes `get_image` within the back end itself.

adam/flask_backend/app.py
sidharth-sundar: Sounds good. The general structure for finding the associated files is present in `get_scene()`, so it's a fairly intuitive addition. On that note, however, I had a question on the way the RGB images are returned:
```py
"scene_images": [
            path.rsplit("data/", maxsplit=1)[-1]
            for path in sorted(
                chain(
                    glob(f"{experiment_dir}/rgb__[0-9]*.png"),
                    glob(f"{experiment_dir}/id_rgb_[0-9]*.png"),
                )
            )
        ],
```
This is the return value for RGB image paths in `get_scene()`. Currently, all the curricula I've worked with use the `id_rgb_[0-9]*` syntax, but I haven't seen any which contain `rgb__[0-9]*`. There are `rgb_[0-9]*` files available, but do we want to display those? Also, was there some case in the past when we used the `rgb__` naming convention?

spigo900: Yes, we do want to use `rgb_[0-9]*` files. They should be the same, name aside. I think that the `rgb__` convention was used previously because I know I've seen live demos where the ID-free images were shown.

adam/flask_backend/app.py
```diff
@@ -110,7 +118,7 @@ def get_scene() -> Any:
         "test_curriculum": testing_curriculum,
         "scene_number": scene_number,
         "scene_images": [
-            path.rsplit("data", maxsplit=1)[-1]
+            path.rsplit("data/", maxsplit=1)[-1]
```
sidharth-sundar, line 121: Currently we split on the word `data` being present, but that was posing some problems for me (it left the `/` after data which made python interpret resulting paths as absolute). However, this could pose an issue if either the train or test curriculum directories contain `data` as part of the name.

I think we could circumvent this with two approaches

1. Only return the file name. Since we have the situation number, training path, and testing path already present in the front end, we can reconstruct the file path.
2. Return the object image information directly with `get_image` here, rather than the image path

spigo900: So one [could also use](https://docs.python.org/3.7/library/os.path.html#os.path.relpath) `os.path.relpath(path, str(DATA_DIR))`. If I'm reading correctly `experiment_dir` should be an absolute path and so should the glob results, so this should work.

sidharth-sundar: That does work!

adam/flask_backend/app.py
sidharth-sundar: Ok, I've added functionality to check (4) above. Even when limiting only to those relevant files, displaying all the available updates still presented far too much information. While looking for any shorter ways to implement (1), I learned that there's a hack you can do in HTML where you can set the title of an element and it displays when you hover over the corresponding element. I limited the alert to the user to only display the situation label, and now it functions similarly to what (1) describes. There is a more formal way of implementing a hover display, but I'm not very familiar with it.

spigo900:
@sidharth-sundar Re: structural issues:

1. I think it makes sense to be more restrictive here unless @lichtefeld objects. Is this a place where we can reuse `LEARNER_DIR_CONTENTS`?
2. I think it makes sense to bundle the images into the initial retrieval -- send everything we've got -- and also have the separate API call so the UI can fetch additional images as they come in.

adam/flask_backend/app.py
spigo900: So it shows the label and then the tooltip lists the image names or relative paths? That sounds fine. Thank you for the creative thinking.

adam/flask_backend/app.py
```diff
@@ -135,7 +147,133 @@ def get_scene() -> Any:
...
+    if not is_relative_to(curriculum_dir, LEARNERS_DIR):
+        return {"message": "Directory out of range"}, HTTPStatus.BAD_REQUEST
+
+    if (
+        app.config["LEARNER_DIR_CONTENTS"].get(learner) is None
+        or app.config["LEARNER_DIR_CONTENTS"][learner].get(training_curriculum) is None
+        or app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum].get(
+            testing_curriculum
+        )
+        is None
```
spigo900, line 195: Nit: Since we're not using the result of `.get()` here, It would be slightly clearer to write this using `not in`:

```python
learner not in app.config["LEARNER_DIR_CONTENTS"]
or training_curriculum not in app.config["LEARNER_DIR_CONTENTS"][learner]
or testing_curriculum not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum]
```

sidharth-sundar:
@spigo900 Re: Re: structural issues:
1. Yes, we could reuse it here. 
2. Bundling images into retrieval is easy to implement, so I can finish that up today. Just one additional comment.
> and also have the separate API call so the UI can fetch additional images as they come in.

Currently the UI is set up to just alert the user if new data is available, and the user would then have to manually retrieve said data/resubmit the request for data. In the original discussion in #1155 we proposed automatically fetching additional images as a possibility, but it wasn't in scope for what I initially planned to work on. However, I think with the approach I took to retrieve data (polling the back end), I can feasibly implement auto-fetching without too much more effort.

spigo900:
@sidharth-sundar Re: fetching as they come in, I meant that more as "it's convenient for future work to leave in the API call." If you feel like adding auto fetching, I think it makes sense to spend up to an hour on that. If it would take more than that, I think it makes more sense to prioritize GNN improvements.

angular-viewer/adam-angular-demo/src/app/components/image-output/image-output.component.ts
```diff
@@ -38,36 +38,33 @@ export class ImageOutputComponent implements OnChanges {
...
+    this.sceneStrokeGraphArray = changes.stroke_graph_images.currentValue;
     this.isImg = true;
   }
 
   onClickCarousel(selection: string): void {
     this.carouselSelection = selection;
   }
 
-  getImage(image_path: string, arr: Array<string>): void {
+  getImage(
```
spigo900, line 51: Did we have a reason to keep this function? I forget. It doesn't seem to be in use.

spigo900 (changes requested):
Looks good, just a few small issues. The main one is and vs. or in `get_image()`.

adam/flask_backend/app.py
```diff
@@ -135,7 +144,168 @@ def get_scene() -> Any:
...
+    training_curriculum = request.args.get("training_curriculum", default="")
+    testing_curriculum = request.args.get("testing_curriculum", default="")
+    scene_number = int(request.args.get("scene_number", "")) - 1
+    situation_dir = f"situation_{scene_number}"
+    image_file = request.args.get("image_file", default="")
+    if (
+        not learner
+        and not training_curriculum
+        and not testing_curriculum
+        and not situation_dir
```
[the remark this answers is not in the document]
sidharth-sundar: I think they should be "ors" as well, but I based it on the preexisting code for the `get_scene` API call which used "ands".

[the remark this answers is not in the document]
spigo900: That makes sense. In that case I think it makes sense to fix both since this PR makes some changes to `get_scene` anyway.

angular-viewer/adam-angular-demo/src/app/components/image-output/image-output.component.ts
sidharth-sundar: Currently it's not in use, so I can remove it

spigo900: I think that makes sense.

adam/flask_backend/app.py
```diff
@@ -135,7 +144,168 @@ def get_scene() -> Any:
...
+        or testing_curriculum
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum]
+        or situation_dir
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum][
+            testing_curriculum
+        ]
+        or image_file
+        not in app.config["LEARNER_DIR_CONTENTS"][learner][training_curriculum][
+            testing_curriculum
+        ][situation_dir]
```
spigo900, line 202: Hm, I just realized technically this part is probably a bit too restrictive. In the absence of the frontend, any requests for images generated after the server starts would be rejected. But I'm okay with that because in practice we would only use this backend together with the frontend, and if we ended up needing this to work in another context it's a pretty easy fix.

spigo900 (approved):
Looks good to me. Just needs history cleanup.

adam/flask_backend/app.py
sidharth-sundar: Right, currently the `LEARNER_DIR_CONTENTS` only updates if `poll_queues` is invoked but if we want it to function independently of that function we can either drop the check or build the full file path and check if the file exists.

spigo900:
I would like to see the pure-cleanup changes to `get_scene()` split out of c7778e847db1317ef066b707050f3dc50f01d855 (change to relpath) and 88742f3893da77f801db2a5c84ee6e07c4c6a6ae (arg validation changes), and ideally combined into one commit that comes before the "new feature" changes. Otherwise this looks good, thanks for the cleanup.

sidharth-sundar:
@spigo900 Re: cleanup changes
Done! I think there's one other thing which _could_ be moved to the initial commit, which is displaying images with the format `rgb_#`. However, I think it makes sense to leave it where it currently is, coupled with the larger change in `get_scene` which now returns image data rather than image paths.

spigo900:
@sidharth-sundar That sounds fine. Are you ready to merge then?

sidharth-sundar:
@spigo900 Yes, it's ready to be merged

PR #1162, opened 2022-07-27
Stroke extraction: Fix crash when seg. image is larger than 256x256

spigo900:
I originally just fixed the shape of `img_tmp` but as far as i can tell we never use it, so we can delete it to fix the crash.


lichtefeld (approved):
lgtm!

PR #1163, opened 2022-08-08
Fix dropped requests for situation_0 scenes

sidharth-sundar:
In the changes in #1161, we added logic to abort the `get_scene` and `get_image` API calls when any individual component of the call was missing, but this logic was faulty and aborted whenever we request contents of a directory with scene number 0 since we detect 0 to be `false`. This PR patches that error and adds logic to abort if an invalid (negative) scene number is requested.


spigo900 (approved):
Looks good to me.

PR #1164, opened 2022-08-08
Remove similarity panel if no data is provided

sidharth-sundar:
Fixes a bug in the front end where the panel at the bottom of the page displaying object similarities/differences lingers when submitting new requests for scenes with no differences present. We now check for cases where there are no differences, and then explicitly clear/hide the difference panel.


lichtefeld (approved):
Thanks! I figured it was something like just not having a reset function for that panel :)

Can you just rebase this on master before merging?

Issue #1157, opened 2022-06-28
Unknown objects rehearsal experiment


spigo900:
We're interested in checking what the object learner does when it decodes unknown objects. We want to run a "rehearsal" experiment here.

As part of this task, you'll need to:

1. Run the unknown objects script on the M5 objects train curriculum, removing the list of objects from earlier (from your look at object contrasts and the discussion). That is:
    1. apple
    2. ball
    3. chair
    4. cube block
    5. window
2. Train an object GNN module (#1110 / #1151) on the rehearsal curriculum.
3. Evaluate that GNN on train and on test. It's worth recording train accuracy for completeness though it doesn't tell us much. The main reason to run inference on train is to produce the decode files ADAM needs (the `feature.yaml` s). Also record the test accuracy. I'd expect this to be very similar to ADAM's test accuracy.
4. Run ADAM with the subset learner over the resulting curriculum / decode. Parameters would be similar to the M5 objects curriculum except for the train curriculum.
5. Since it's ~easy to do (we already have observers set up to do it), it's probably worth collecting/comparing per-object accuracy and qualitative outcomes results vs. the `m5_objects_v0_with_mugs` baseline results.
6. The main interesting thing to look at here are the per-object results. I think this means:
    1. A table of ADAM's train vs. test accuracies in the baseline (`m5_objects_v0_with_mugs`) case and for "unknown objects." (Mostly for completeness.)
    6. The confusion matrices for ADAM's output on train and on test. (details on confusion matrices: see #1156.)
    7. Maybe separate matrices for the GNN?
    8. Separately, a plot of only those rows of the test-time confusion matrix (or matrices) that shows only the objects removed from train (the "unknown objects").
    9. As with #1156, remember to avoid red/green contrasts in plots due to colorblindness.
    10. A writeup containing these figures plus discussion focusing on the unknown objects outcomes.
        1. e.g. we expect the GNN to confuse unknown objects with ones it saw during training, how weird vs. "reasonable" are the confusions it makes. Is it confusing apple/ball with orange, or with book? etc.
        2. This probably requires manually looking at the stroke and stroke graph images for the unknown objects in the test set.

PR #1165, opened 2022-08-16
1156 downsampling experiment

### LINKED ISSUE #1156: Downsampling experiment
We're interested in checking how well the object learner works when it has fewer samples to learn from. We want to run experiments where it has fewer images or samples per object to learn from.

As part of this task, you'll need to:

1. Run the downsampling script on the M5 objects train curriculum, downsampling to `k=2` and `k=10` images/samples per object. So there would be two curriculum, say `m6_objects_downsampled_2pertype` and `m6_objects_downsampled_10pertype`.
2. Train one object GNN module (#1110 / #1151) on each of the downsampled curricula. So, two GNNs trained.
3. Evaluate both the GNNs on train and on test. It's worth recording train accuracy for completeness though it doesn't tell us much. The main reason to run inference on train is to produce the decode files ADAM needs (the `feature.yaml` s). Also record the test accuracies. I'd expect this to be very similar to ADAM's test accuracy in general.
4. Run ADAM with the subset learner over the resulting curricula & decodes. Parameters would be similar to the M5 objects curriculum except for the train curriculum.
5. Collect/compare accuracy results between `k=2`, `k=10`, and baseline (`m5_objects_v0_with_mugs`). It will probably be interesting here to look at and analyze the per-object results as well as the overall accuracy. I think this means:
    1. Plotting the three GNN test accuracy values, probably as a bar chart.
    2. Plotting the three GNN train accuracy values.
    3. Separately plot the ADAM train/test accuracies? Or only test if that is easier. (I don't think we do a final pass through train so evaluating on train may be more annoying than it's worth.)
    4. Check that you don't red and green bars within a chart due to colorblindness. I think the default palette goes blue->orange->gray, so this shouldn't be a problem, but worth checking.
    5. Plotting the six GNN confusion matrices: One for each of the [(three models) X (two splits, train vs. test)].
        1. If you're not familiar with confusion matrices, this would be a table with 20 rows corresponding to "actual object type", and 20 columns to "the GNN's output prediction/what kind of object the GNN thought it was." Each cell is a number saying how many times the GNN thought that, for example, a chair was an apple vs. a chair vs. a desk, ...
        2. It might be worth making confusion matrices for ADAM separately, though because ADAM outputs multiple labels per sample that gets weirder.
    6. A writeup containing these images and some discussion of how sample size seems to affect (or not affect!) learning.
        1. This might also include a table showing the overall accuracy results by sample size and split.
        3. Generally I expect overall accuracy to suffer with fewer train samples. This is something to look out for. If accuracy goes up with fewer train samples, there is probably a bug somewhere. Assuming accuracy dues suffer the result is probably worth pointing out in one sentence but not worth discussing much.
        4. I think the main topic of interest is likely to be object accuracy for different types of object. For example maybe it turns out sample size affects windows more than balls.

sidharth-sundar:
Closes #1156 and #1157 

Performed both within the same branch because I needed similar scripts for both for visualizing GNN/ADAM performance between different experiments.

There are 4 important features added here:

1. Script to generate confusion matrix for GNN outputs given an input curriculum
2. Script to generate confusion matrix for ADAM outputs given an input curriculum
3. Script to generate confusion matrix for top GNN outputs per object given an input curriculum
4. Added functionality to
   - Retrieve top `k` GNN decodes per object from `shape_stroke_graph_inference.py`
   - Add each of the top `k` decodes as a node in ADAM and handle accordingly.

Aside from those, most of the new additions here are just adding parameter files which I used the preexisting downsample/unknown object generation scripts to create the curricula for.

### DIFF
--- a/adam_preprocessing/shape_stroke_graph_inference.py
+++ b/adam_preprocessing/shape_stroke_graph_inference.py
@@ -4,6 +4,7 @@ from argparse import ArgumentParser
 from copy import deepcopy
 import logging
 from pathlib import Path
+from typing import Sequence
 
 import torch
 import torch.nn as nn
@@ -22,12 +23,19 @@ from MPNN import MPNN, MPNN_Linear
 from utils import accuracy, get_stroke_data, LinearModel, load_data, STRING_OBJECT_LABELS
 
 
-def update_features_yaml(features_yaml, *, predicted_object: str):
+def update_features_yaml(features_yaml, *, predicted_objects: Sequence[str]):
     # jac: not terribly efficient to do a full deepcopy, but who cares, this should be a small dict
     # anyway... also this is probably much less expensive than the GNN inference itself.
     result = deepcopy(features_yaml)
     for object_ in result["objects"]:
-        object_["stroke_graph"]["concept_name"] = predicted_object
+        if len(predicted_objects) == 1:
+            object_["stroke_graph"]["concept_name"] = predicted_objects[0]
+            if "concept_names" in object_["stroke_graph"]:
+                del object_["stroke_graph"]["concept_names"]
+        else:
+            object_["stroke_graph"]["concept_names"] = predicted_objects
+            if "concept_name" in object_["stroke_graph"]:
+                del object_["stroke_graph"]["concept_name"]
     return result
 
 
@@ -49,6 +57,12 @@ def main():
         default=None,
         help="Directory where we should write the feature outputs to. Outputs are structured as if "
         "the output path is a curriculum directory.",
+    )
+    parser.add_argument(
+        "--top_k",
+        type=int,
+        default=1,
+        help="Top k decodes to retrieve from the GNN",
     )
     args = parser.parse_args()
     logging.basicConfig(level=logging.INFO)
@@ -129,7 +143,7 @@ def main():
 
         assert outputs.size(0) == curriculum_params["num_dirs"]
 
-        predicted_label_ints = outputs.argmax(dim=1)
+        _, predicted_label_ints = outputs.topk(args.top_k)
         n_saved = 0
         for situation_num in range(curriculum_params["num_dirs"]):
             input_situation_dir = args.curriculum_path / f"situation_{situation_num}"
@@ -143,8 +157,8 @@ def main():
 
                 updated_features = update_features_yaml(
                     features,
-                    predicted_object=STRING_OBJECT_LABELS[
-                        predicted_label_ints[situation_num]
+                    predicted_object=[STRING_OBJECT_LABELS[
+                        predicted_label_ints[situation_num][i]] for i in range(args.top_k)
                     ],
                 )
 
--- a/adam_preprocessing/shape_stroke_graph_inference.py
+++ b/adam_preprocessing/shape_stroke_graph_inference.py
@@ -137,7 +137,7 @@ def main():
             if len(feature_yamls) == 1:
                 # Load features, update them, then save
                 with open(
-                    input_situation_dir / feature_yamls[0], encoding="utf-8"
+                    feature_yamls[0], encoding="utf-8"
                 ) as feature_yaml_in:
                     features = yaml.safe_load(feature_yaml_in)
 
--- a/scripts/GNN_topk_analysis.py
+++ b/scripts/GNN_topk_analysis.py
@@ -1,0 +1,88 @@
+import argparse
+import logging
+from os import makedirs
+from pathlib import Path
+from shutil import rmtree, copytree
+from typing import Mapping, Sequence
+
+import matplotlib.pyplot as plt
+import pandas as pd
+import seaborn as sn
+import yaml
+
+def parse_in_domain_objects(input_dir: Path) -> Sequence:
+    in_domain_objects = set()
+    for feature_file_path in sorted(input_dir.glob("situation*")):
+        with open(feature_file_path / 'description.yaml', encoding="utf-8") as description_file:
+            description_yaml = yaml.safe_load(description_file)
+            object = description_yaml['language'].rsplit()[-1]
+            in_domain_objects.add(object)
+    return sorted(in_domain_objects)
+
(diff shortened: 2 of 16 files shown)


scripts/GNN_vs_expected.py
```diff
@@ -0,0 +1,84 @@
...
+            in_domain_objects.add(object)
+    return sorted(in_domain_objects)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Get confusion matrix for GNN decode"
+    )
+    parser.add_argument('--input-dir', type=Path, help="Directory to extract GNN decode to")
+    parser.add_argument('--output-dir', type=Path, help="File to write output .csv to")
```
spigo900, line 28: Argument help text should be updated.

scripts/GNN_vs_expected.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if output_dir.is_dir():
+        if not force_overwrite:
+            logging.warning("There already exists a directory in the target location")
+            raise FileExistsError(str(output_dir))
+    else:
+        makedirs(output_dir)
+
+    in_domain_objects = parse_in_domain_objects(input_dir)
+    arr = [[0 for i in range(len(in_domain_objects))] for i in range(len(in_domain_objects))]
```
spigo900, line 56: This variable could use a better name, say `confusion_matrix`.

scripts/adam_decode_vs_expected.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if output_dir.is_dir():
+        if not force_overwrite:
+            logging.warning("There already exists a directory in the target location")
+            raise FileExistsError(str(output_dir))
+    else:
+        makedirs(output_dir)
+
+    in_domain_objects=parse_in_domain_objects(input_dir)
+    arr = [[0 for i in range(len(in_domain_objects) + 1)] for i in range(len(in_domain_objects))]
```
spigo900, line 56: Could use a better name.

scripts/GNN_topk_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+    if output_dir.is_dir():
+        if not force_overwrite:
+            logging.warning("There already exists a directory in the target location")
+            raise FileExistsError(str(output_dir))
+    else:
+        makedirs(output_dir)
+
+    in_domain_objects = parse_in_domain_objects(input_dir)
+    object_counts = {in_domain_object: 0 for in_domain_object in in_domain_objects}
+    arr = [[0 for i in range(len(in_domain_objects))] for i in range(len(in_domain_objects))]
```
spigo900, line 57: This variable could use a better name, say `confusion_matrix`.

scripts/GNN_topk_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+            in_domain_objects.add(object)
+    return sorted(in_domain_objects)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Get confusion matrix for GNN decode"
+    )
+    parser.add_argument('--input-dir', type=Path, help="Directory to extract GNN decode to")
+    parser.add_argument('--output-dir', type=Path, help="File to write output .csv to")
```
spigo900, line 28: Argument help text should be updated.

scripts/adam_decode_vs_expected.py
```diff
@@ -0,0 +1,92 @@
...
+            in_domain_objects.add(object)
+    return sorted(in_domain_objects)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Get confusion matrix for GNN decode"
+    )
+    parser.add_argument('--input-dir', type=Path, help="Directory to extract adam decodes from")
+    parser.add_argument('--output-dir', type=Path, help="File to write output .csv to")
```
spigo900, line 28: `--output-dir` help text should be updated.

spigo900 (changes requested):
Looks good overall, just a few style nits. (And maybe one other small thing? Apparently I left 8 comments but I only found 7 looking back. 🤷‍♂️)

adam_preprocessing/shape_stroke_graph_inference.py
```diff
@@ -22,12 +23,19 @@
...
 
-def update_features_yaml(features_yaml, *, predicted_object: str):
+def update_features_yaml(features_yaml, *, predicted_object: Sequence[str]):
     # jac: not terribly efficient to do a full deepcopy, but who cares, this should be a small dict
     # anyway... also this is probably much less expensive than the GNN inference itself.
     result = deepcopy(features_yaml)
     for object_ in result["objects"]:
-        object_["stroke_graph"]["concept_name"] = predicted_object
+        if len(predicted_object) == 1:
+            object_["stroke_graph"]["concept_name"] = predicted_object
```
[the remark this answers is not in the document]
sidharth-sundar: It should indeed! Thanks for catching that. Since the multiple decode experiment was the last thing I worked on, I never ran the script with only 1 decode per object (the default behavior)

scripts/GNN_topk_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+    if output_dir.is_dir():
+        if not force_overwrite:
+            logging.warning("There already exists a directory in the target location")
+            raise FileExistsError(str(output_dir))
+    else:
+        makedirs(output_dir)
+
+    in_domain_objects = parse_in_domain_objects(input_dir)
+    object_counts = {in_domain_object: 0 for in_domain_object in in_domain_objects}
+    GNN_rank_matrix = [[0 for _ in range(len(in_domain_objects))] for _ in range(len(in_domain_objects))]
```
spigo900, line 57: This is better, thank you. Nit: case. I think it's better to stick with the conventional lowercase here.

spigo900 (approved):
One nitpick. Otherwise, looks good and just needs history cleanup. Thank you.

PR #1171, opened 2022-09-09
Optional param defaults to None

boyleconnor:
[Optional parameters](https://docs.python.org/3/library/typing.html#typing.Optional) should default to `None`. Otherwise this function has to be explicitly called with `None` as the value for `dir_num`.


spigo900 (approved):
Looks good to me.

lichtefeld (approved):

boyleconnor:
I don't have permission to merge

PR #1174, opened 2022-09-12
Readme fixes

### LINKED ISSUE #1172: README suggests different Python script than sbatch file
The [preprocessing README](https://github.com/isi-vista/adam/tree/master/adam_preprocessing#stroke-extraction) says to run the `adam_preprocess/shape_stroke_graph_learner.py` script, while the next section (which is implied to be calling the same python script) calls `extract_strokes.sh`, which in turn calls `shape_stroke_extraction.py`

spigo900:
Closes #1172. Also addresses the other README issues we discussed last team meeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ You can check out our papers to read more about the system and to cite our work:
 
    ```
    adam_root: PATH_TO_WORKING_COPY_OF_THIS_REPO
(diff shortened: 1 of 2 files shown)


lichtefeld (changes requested):
@spigo900 To fix CI you'll need to go run `make precommit` from within the `angular` directory so the style formatter for the `README` file process. Unfortunately, it's not currently linked to our outer precommit script.

lichtefeld (approved):
lgtm!

PR #1176, opened 2022-09-15
In README: switch working directories

boyleconnor:
In my experience, you need to switch your working directory to `adam_preprocessing` in order to allow the `ske.m` file to be available to the MATLAB backend. Not sure if there's some other way to achieve this that I'm missing.


spigo900:
Thanks for finding this. I'm inclined to fix the code here rather than the README -- the working directory dependency seems like trouble.

PR #1177, opened 2022-09-15
Add absolute dirs to MATLAB path to make stroke extraction independent of working dir

spigo900:
Alternative to #1176 where we fix the code instead of the README.

@boyleconnor Let me know if this fixes the problem on your end.

### DIFF
--- a/adam_preprocessing/shape_stroke_extraction.py
+++ b/adam_preprocessing/shape_stroke_extraction.py
@@ -236,10 +236,8 @@ class Stroke_Extraction:
         - E is a k x 2 sequence of "graph edges"
         """
         eng = matlab.engine.start_matlab()
-        s = eng.genpath("./lightspeed")
-        eng.addpath(s, nargout=0)
-        s = eng.genpath("./BPL")
-        eng.addpath(s, nargout=0)
+        file_dir = os.path.dirname(__file__)
+        eng.addpath(eng.genpath(file_dir), nargout=0)
         out = eng.ske(self.path, nargout=2)
         eng.close()
         return out


adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -236,9 +236,12 @@ def stroke_extraction_from_matlab(self):
...
         """
         eng = matlab.engine.start_matlab()
-        s = eng.genpath("./lightspeed")
+        file_dir = os.path.dirname(__file__)
+        s = eng.genpath(file_dir)
         eng.addpath(s, nargout=0)
-        s = eng.genpath("./BPL")
+        s = eng.genpath(os.path.join(file_dir, "lightspeed"))
+        eng.addpath(s, nargout=0)
+        s = eng.genpath(os.path.join(file_dir, "BPL"))
```
boyleconnor, line 244: Checked that it works mostly LGTM. One suggestion, inline the `s` variables:

```
        eng.addpath(eng.genpath(file_dir), nargout=0)
        eng.addpath(eng.genpath(os.path.join(file_dir, "lightspeed")), nargout=0)
        eng.addpath(eng.genpath(os.path.join(file_dir, "BPL")), nargout=0)
```

boyleconnor (approved):

adam_preprocessing/shape_stroke_extraction.py
spigo900: Good idea, thanks. I also deleted the second `addpath()` calls since it looks like they're redundant ([genpath()](https://www.mathworks.com/help/matlab/ref/genpath.html) includes subdirectories).

PR #1179, opened 2022-09-20
Predict diverse objects in scenarios

boyleconnor:
This is still very messy and incomplete (I haven't touched the training script; the `num_objs` variable turned out to be unnecessary, will factor out soon) but I wanted to check in sooner rather than later that this is the desired change (of recognizing multiple objects in preprocessing) for later adding spatial relations to ADAM.

EDIT: This PR is now ready for review

### DIFF
--- a/adam_preprocessing/utils.py
+++ b/adam_preprocessing/utils.py
@@ -38,6 +38,30 @@ STRING_OBJECT_LABELS = [
 def get_stroke_data(
     curriculum_path: Path, train_or_test: str, *, dir_num: Optional[int] = None, int_curriculum_labels: bool = True, multi_object: bool = False
 ) -> Tuple[Sequence[np.ndarray], Sequence[np.ndarray], Sequence[int]]:
+    """Load data on strokes from each scenario feature file in each scenario
+       dir in curriculum.
+
+    Params:
+        curriculum_path: path containing curriculum
+        train_or_test: whether the curriculum is for training or testing
+        dir_num: if specified, only extract strokes from the situation with this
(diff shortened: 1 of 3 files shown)


adam_preprocessing/utils.py
```diff
@@ -71,50 +80,68 @@ def get_stroke_data(
...
+                for object_ in features["objects"]:
+                    coords = np.asarray(object_["stroke_graph"]["strokes_normalized_coordinates"])
+                    curriculum_coords.append(coords)
+                    adj = np.asarray(object_["stroke_graph"]["adjacency_matrix"])
+                    curriculum_adjs.append(adj)
+                    if int_curriculum_labels:
+                        integer_label, _ = label_from_object_language_tuple(
+                            language_tuple, situation_num
+                        )
+                        curriculum_labels.append(integer_label)
```
spigo900, line 111: I'm assuming this is WIP code since it's so close to the multi-object code. Just noting that the `int_curriculum_labels` should go outside the loop.

spigo900 (commented):
Yes, this looks right to me. Thank you for checking in early.

adam_preprocessing/utils.py
boyleconnor: Sorry, I missed this when you posted it 2 days ago. I believe I have fixed it in the meanwhile; there is only one set of `(coords, adj, label)`extracted per feature file when in single-object mode

lichtefeld (approved):
This looks fine to me, @spigo900 should probably also review in case he has any comments

spigo900 (changes requested):
A few minor comments, overall looks good to me.

PR #1183, opened 2022-09-23
Add spatial/size features to preprocessing

boyleconnor:


adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -502,6 +530,15 @@ def get_strokes(self):
                         strokes_normalized_coordinates=reduced_strokes_norm.transpose(
                             0, 2, 1
                         ).tolist(),
+                        size=dict(
+                            # FIXME: calling .item() causes crash (it says they
+                            #  are type `float`), but if they aren't cast to
+                            #  `float`, they get serialized as complex objects,
+                            #  not primitive floats.
```
spigo900, line 537: 🤔

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -38,6 +38,9 @@
 from utils import label_from_object_language_tuple
 
 
+GAMMA = 0.05
```
spigo900, line 41: I'd like to rename this something more informative, say `TOUCHING_REL_DISTANCE_THRESHOLD`.

spigo900 (changes requested):
This looks good, thank you.

The features look right, but I think one's missing. Referring back to spatial relations I think the only one I don't see here is [relative sizes](https://docs.google.com/document/d/10ShN9AYLZ52OWBCdo1tIbYsGxM6o1aIWP7XP_D25sM4/edit#heading=h.ahfg91k29ywp). That would be another dict like `relative_distance` giving features `other_{foo}_bigger` and `other_{foo}_ratio` where the ratio is `{foo} / other_{foo}` (this divided by other) and `foo` can be any of `x` ,`y`, or `area`.

adam_preprocessing/shape_stroke_extraction.py
spigo900: to clarify, not a problem with your code -- I am just wondering to myself why this happens

boyleconnor:
Thanks for the review. To be clear, when you said:

> That would be another dict like `relative_distance`...

I think you meant to write `relative_size`. Correct?

spigo900:
Yes, it would be called `relative_size`, but I meant to write `relative_distance` -- I was comparing the dict to the one for `relative_distance`.

boyleconnor:
@spigo900 I just realized that the way you specified relative distances in the document is different from what I did. The document says to normalize by the product of both objects, not just the object whose feature-set we're in. Do you want me to change my implementation to match the document's specification?

spigo900:
@boyleconnor Ah, I forgot about that. Yes, please update that.

boyleconnor:
I still think it's a bit strange to base the `touching` relationship on the center-to-center distance of two objects. It helps that it is based on relative distance (so it incorporates the objects' own sizes), but it's still not the way I'd do it.

spigo900:
@boyleconnor What would you do instead?

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -510,6 +544,35 @@ def get_strokes(self):
...
+            for j in range(len(data)):
+                if i == j:
+                    continue
+                else:
+                    other_object_name = data[j]['object_name']
+                    for axis in ('x', 'y'):
+                        # Relative size is a discrete ('bigger_than') relation
+                        if data[i]['stroke_graph']['size'][axis] > \
+                                data[j]['stroke_graph']['size'][axis]:
+                            data[i]['relative_size'][axis+'_bigger_than'].append(other_object_name)
```
spigo900, line 560: It would be slightly simpler for `{axis}_bigger_than` to be a list of booleans. Then you don't need the `if`.

I think an f-string would make the relation name clearer: `f"{axis}_bigger_than"` vs. `axis+'_bigger_than_'`.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -510,6 +544,35 @@ def get_strokes(self):
...
+
+                        # Normalize relative distances by the relevant axis
+                        # sizes of both objects
+                        object_axis_size = data[i]['stroke_graph']['size'][axis]
+                        other_object_axis_size = data[j]['stroke_graph']['size'][axis]
+                        data[i]['relative_distance'][other_object_name][axis+'_offset'] /= (object_axis_size * other_object_axis_size)
```
spigo900, line 566: f-string again.

Also, because this is an assignment you don't need the parens for grouping. Except for the new walrus operator `:=`, Python assignments aren't expressions in the sense that `x = expr` can't be treated as having a value (not even `None`) so there's no ambiguity in writing `x /= y * z` -- it has to mean `x = x / (y * z)`.

boyleconnor:
@spigo900 I didn't have any one concrete method in mind when I wrote that, but I suppose you could look at the individual control points of the strokes (since they're outlines, they should actually represent the visual edges of the objects). Since the strokes are reduced to 10 control points each, even a brute force comparison would be at most a couple thousand comparisons per object pair.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -502,6 +527,15 @@ def get_strokes(self):
                         strokes_normalized_coordinates=reduced_strokes_norm.transpose(
                             0, 2, 1
                         ).tolist(),
+                        size=dict(
```
spigo900, line 530: I think it probably makes things easier to put this at the outer level -- so at the same level as `relative_distance` instead of being nested inside `stroke_graph`.

spigo900 (changes requested):
Some comments re: the new loop.

spigo900:
@boyleconnor Want to implement that method and compare running times on the full curriculum? I'm not opposed if it's fast enough and it's a fair point the number of comparisons is low.

adam_preprocessing/shape_stroke_extraction.py
boyleconnor: Personally, I find it a bit easier to read at a glance when there are parentheses, event though `/=` is an assignment operator with no value. I can change it if you think it is detracting from the readability

spigo900: No problem then.

boyleconnor:
@spigo900 I've been running stroke extraction on my laptop, which is not very scientific, but it should at least get us a rough estimate of the difference in run time. This is the time for running stroke extraction on `m5_objects_v0_with_mugs` with the pointwise touching calculations:

```
7640.89s user 1260.65s system 234% cpu 1:03:14.52 total
```

I am currently running it without the pointwise touching calculations to see the difference in time. I'd guess it will take around 45 minutes.

boyleconnor:
Here is the time for the run with no pairwise calculations (same dataset, also on my laptop):

```
7673.44s user 1276.39s system 301% cpu 49:30.83 total
```

spigo900:
@boyleconnor Good enough for me. If we find it's too slow we can easily fall back on the dumber version, or else something slightly smarter based on the bounding boxes. I'll re-review this later today.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -546,41 +563,52 @@ def get_strokes(self):
...
-            data[i]['relative_size'] = relative_size
-
-        self.data = data
+                    object_area = objects[i]['size']['area']
+                    other_object_area = objects[j]['size']['area']
+                    objects[i]['relative_distance'][other_object_name]['euclidean'] /= (object_area * other_object_area)
+            objects[i]['relative_size'] = relative_size
+
+        # Determine touching relations based on closest-point distance
+        touching = set()
```
spigo900, line 593: I'd prefer to use a list for `touching` here. Sets aren't deterministically ordered and I expect this to result in inconsistent serializations. Meanwhile, we don't actually need any set operations -- the only operations used on `touching` are (1) append, and (2) YAML serialization.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -111,6 +111,20 @@ def kp2stroke(strokeinfo):
...
+    euclidean (absolute) distance of their closest two points.
+    """
+    object_one_points = np.concatenate(object_one['stroke_graph']['original_reduced_strokes'], axis=0)
+    object_two_points = np.concatenate(object_two['stroke_graph']['original_reduced_strokes'], axis=0)
+    for point_one in object_one_points:
+        for point_two in object_two_points:
+            distance = np.sum((point_one - point_two) ** 2) ** 0.5
+            if distance < TOUCHING_DISTANCE_THRESHOLD:
+                return True
+    return False
```
spigo900, line 125: IIRC NumPy performs better when using its whole-array operations instead of Python loops. If so we might speed this up by switching to a loop-free version. This doesn't handle shapes correctly, but something like this:

```python
return np.any(
  np.sum(
    (object_one_points[:, None] - object_two_points) ** 2
  ) ** 0.5 < TOUCHING_DISTANCE_THRESHOLD
)
```

spigo900 (changes requested):
Looks good. Two ideas for possible changes.

adam_preprocessing/shape_stroke_extraction.py
spigo900: Testing with some arrays in the REPL, this sum should work (I think) but we'll need `axis=-1`. I'm assuming the concatenated arrays have shape `(10 * n_strokes, 3)`. Toy example:

```python
import numpy as np
x = np.array(
    [
        [0, 0, 1],
        [1, 0, 0],
    ]
)
y = np.array(
    [
        [0, 0, 2],
        [5, 0, 0],
        [0, 9, 0],
    ]
)
distances = np.sum((x[:, None] - y) ** 2, axis=-1) ** 0.5
```

boyleconnor: This looks good to me (with the `axis=-1` for the `sum()` to just combine coordinates of course)!

I had thought about this, too, but I couldn't think of how to do it but I figured the cost in human time of looking up how to do this in NumPy wasn't worth the saved computer time. Since you wrote it up, I'll incorporate it.

spigo900: That's fair. I too worried that this might be too gross to be worthwhile, but it turned out easier than expected when I looked into it.

boyleconnor:
At commit b08024e, I ran the stroke extraction script on the m5 train curriculum (again on my ISI laptop) and it finished with the following time:

```
 7842.58s user 1292.84s system 301% cpu 50:30.49 total
```

It looks like the NumPy calculations were a big boost to speed.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -504,21 +553,63 @@ def get_strokes(self):
...
+            for j in range(len(objects)):
+                if i == j:
+                    continue
+                else:
+                    other_object_name = objects[j]['object_name']
+                    for axis in ('x', 'y'):
+                        # Relative size is a discrete ('bigger_than') relation
+                        if objects[i]['size'][axis] > \
+                                objects[j]['size'][axis]:
+                            relative_size[f'{axis}_bigger_than'].append(other_object_name)
```
spigo900, line 578: It would be slightly simpler for the list here, `relative_size[f'{axis}_bigger_than']`, to be a list of booleans. Then you don't need the `if`.

spigo900 (commented):
Looks good, just one thread left from earlier.

adam_preprocessing/shape_stroke_extraction.py
boyleconnor: @spigo900 do you mean something like this:

```yaml
  relative_size:
    x_bigger_than:
    - false
    y_bigger_than:
    - true
```

(generated using commit 6ccd5b4)

spigo900: Yes, if the scene has two objects.

spigo900 (approved):
Looks good to me.

PR #1184, opened 2022-09-26
Color segmentation

spigo900:


spigo900:
FYI: The implementation of `cluster_small_strokes()` is not very efficient, [particularly the way it iterates over stroke clusters](https://github.com/isi-vista/adam/pull/1184/commits/f16a446c931444b593f202a2e84a0fc42e21a454#diff-61e49aa5f7b89dca1bb3ff0efe17d75baff24cbbfb0137ab363dc8df03a35932R202-R203). If either of you have suggestions I'm open to it. I thought of using a heap but that doesn't seem to work because once I've chosen a merge target I need to remove that one as well (and only that one) -- unless I just pop until that stroke is gone and then continue, rebuilding the heap every loop. But I'm not sure that's better than sorting the list of clusters every loop.

Anyway, it works and seems fast enough. That is, it takes AFAICT about 1.5 hours to complete on the M5 objects training data when using the color-refined segmentations. So, a while but not unreasonably long.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -108,6 +110,309 @@ def kp2stroke(strokeinfo):
...
+    For strokes with just one point, we plot a circle instead.
+
+    Parameters:
+        ax: The axes to plot on.
+        strokes: The strokes to plot.
+    """
+    from matplotlib.rcsetup import cycler
+    import matplotlib.patheffects as patheffects
+    from itertools import cycle
+    colors = cycler(color=['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd', '#8c564b', '#e377c2', '#7f7f7f', '#bcbd22', '#17becf'])
```
lichtefeld, line 363: Nit: I'm assuming these are from a copy/paste operation? No need to move them just to modify this function though.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -108,6 +110,309 @@ def kp2stroke(strokeinfo):
...
+    Plot stroke graph on the given axes.
+
+    We try to position the vertices as close as possible to the mean of the stroke's points.
+
+    Parameters:
+        ax: The axes to plot on.
+        strokes: The strokes to plot.
+        adj: The adjacency matrix over stroke indices.
+    """
+    from itertools import cycle, islice
```
lichtefeld, line 392: nit: Same with these imports.

lichtefeld (approved):
Look goods. I don't have a better suggestion immediately for your loop so long as its fast-enough in practice for us I'm fine with it. Our curriculum sizes are relatively small :)

A few small nits about local imports I'm assuming they are from other code so no need to change them just for that.

adam_preprocessing/shape_stroke_extraction.py
spigo900: Good point. I sketched these while I was debugging and forgot to move the imports. I'll make the change before merging.

spigo900: Moved these to top level.

adam_preprocessing/shape_stroke_extraction.py
spigo900: Moved these to top level.

PR #1188, opened 2022-09-29
Pipeline script

spigo900:
Implements a script to queue all preprocessing steps from one script. The preprocessing steps are queued as separate Slurm jobs with dependencies on each other, set up to cancel immediately if one of their dependencies is canceled/fails. This script can also supports running each step iteratively and interactively by running the Slurm scripts in Bash.

I haven't yet tested this with color segmentation or instance segmentation. Once I have both working I may add segmentation server startup to this.

### DIFF
--- a/adam/experiment/run_pipeline.py
+++ b/adam/experiment/run_pipeline.py
@@ -1,0 +1,644 @@
+"""
+A script intended to run the entire M3 pipeline.
+
+Note that this script does not take care of starting the object segmentation server. You have to do
+that separately.
+
+Note also that this writes to then reads from a params file for ADAM. If you run this again with the
+same params file name while a job is running, you will overwrite the old parameters! So be careful.
+"""
+import atexit
+from datetime import datetime
+from io import StringIO
+from locale import getpreferredencoding
+import logging
+import os
+from os import getenv
+from pathlib import Path
+import shlex
+from shutil import copytree
+from subprocess import run, CalledProcessError
+from typing import List, Mapping, NewType, Optional, Sequence
+
+import yaml
+
+from vistautils.parameters_only_entrypoint import parameters_only_entry_point
+from vistautils.parameters import Parameters, ParameterError
+
+logger = logging.getLogger(__name__)
+
+
+SlurmJobID = NewType("SlurmJobID", str)
+Email = NewType("Email", str)
+MailType = NewType("MailType", str)
+
+slurm_jobs_scheduled: List[SlurmJobID] = []
+
+
+def stop_slurm_jobs():
+    """
+    Exit handler to clean up Slurm jobs.
+    """
+    for job_id in slurm_jobs_scheduled:
+        logger.info("Canceling previously scheduled job %s...", job_id)
+        cancel_command = ["scancel", str(job_id)]
+        print(" ".join(shlex.quote(part) for part in cancel_command))
+        run(cancel_command, check=False)
+
+
+def build_sbatch_command(
+    script_path: Path,
+    *,
+    dependencies: Sequence[SlurmJobID] = (),
+    job_name: Optional[str] = None,
+    log_dir: Optional[Path] = None,
+    email: Optional[Email] = None,
+    mail_types: Sequence[MailType] = (),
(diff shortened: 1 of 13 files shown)


spigo900:
I've since tested this with color segmentation and it works as far as I could tell. I did some work here to separate the segmentation and refinement scripts as well which means color and instance segmentation can run in parallel.

As for instance segmentation, I've added code for server startup and shutdown but still need to test it to confirm it works.

spigo900:
Testing is still in progress and should be done today assuming no more bugs to fix. Since the basic setup seems to work, I think this is ready for review now -- though I may push a few bug fixes between today and Monday.

spigo900:
Update: The pipeline finished, and as far as I can tell, everything works.

adam/experiment/run_pipeline.py
```diff
@@ -0,0 +1,646 @@
+"""
+A script intended to run the entire M3 pipeline.
```
lichtefeld, line 2: Is this supposed to be Phase 3 or milestone 3?

adam/experiment/run_pipeline.py
```diff
@@ -0,0 +1,646 @@
...
+
+
+def stop_slurm_jobs():
+    """
+    Exit handler to clean up Slurm jobs.
+    """
+    for job_id in slurm_jobs_scheduled:
+        logger.info("Canceling previously scheduled job %s...", job_id)
+        cancel_command = ["scancel", str(job_id)]
+        print(" ".join(shlex.quote(part) for part in cancel_command))
```
lichtefeld, line 45: Is this a debugging print statement or an intended one? It's unclear to me which its suppose to be.

adam/experiment/run_pipeline.py
```diff
@@ -0,0 +1,646 @@
+"""
+A script intended to run the entire M3 pipeline.
+
+Note that this script does not take care of starting the object segmentation server. You have to do
+that separately.
```
lichtefeld, line 5: Is this still accurate?

slurm/adam.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+#SBATCH --job-name=ADAM
+#SBATCH --account=borrowed
+#SBATCH --partition=scavenge
+#SBATCH --qos=scavenge
```
lichtefeld, line 5: Do we want to run on `scavenge`? Running on `ephemeral` gives us a guarantee of 1 hour minimum runtime which given we're only requesting an hour for this job seems reasonable?

slurm/segment.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+#SBATCH --job-name=adamSegment
+#SBATCH --account=borrowed
+#SBATCH --partition=scavenge
+#SBATCH --qos=scavenge
```
lichtefeld, line 5: Same note on partition here

lichtefeld (changes requested):
Looks good in general, I haven't tried to make a params file yet but it doesn't seem difficult. 

Just a few minor questions :)

adam/experiment/run_pipeline.py
spigo900: Phase 3. Whoops.

slurm/adam.sh
spigo900: I wasn't aware of the guarantee, but that makes sense.

adam/experiment/run_pipeline.py
spigo900: Actual one, if I recall correctly..

adam/experiment/run_pipeline.py
spigo900: Nope. 🙃

slurm/adam.sh
lichtefeld: It's detailed here in the [shared resources policy](https://github.com/isi-vista/saga-cluster/wiki/Adding-a-Project-to-SAGA:-Policy-(Work-in-Progress)#shared-resources)

lichtefeld (approved):
lgtm!

PR #1189, opened 2022-10-03
Fix GNN decode when some situations are missing stroke output

### LINKED ISSUE #1187: GNN decode doesn't work correctly when lacking strokes for a scene
The GNN decode script crashes in single-object mode if any scene is missing strokes. This is because it doesn't know how to align the model outputs with the situations when some situations have no outputs. (It looks like the decode script also shouldn't work in multi-object when  some situations lack strokes, though it may not crash.)

To fix this, we need to keep track of the `situation number ->* stroke data` relationship when loading stroke data. An easy way to do this would be to track a `situation_number ->* input_index` mapping in `get_stroke_data()` and add this to the return value, then use that in the decode script to build `predictions_by_object`. We could then get rid of the `outputs.size(0) == curriculum_params["num_dirs"]` assertion, instead checking the "we have enough outputs" condition using `zip_equal()` in the prediction update code. As a bonus I think this could also remove the need for special logic to save multi-object decodes.

spigo900:
Closes #1187. I've tested this briefly and it seems to fix the crash. I tested on the "M5 objects v0 with mugs" curriculum both single-object and multi-object. I also tested with the live demo, using both "good" examples and some examples where we know we won't get any strokes.

### DIFF
--- a/adam_preprocessing/shape_stroke_graph_inference.py
+++ b/adam_preprocessing/shape_stroke_graph_inference.py
@@ -168,8 +168,6 @@ def main():
                 args.curriculum_path / "info.yaml", encoding="utf=8"
             ) as curriculum_info_yaml:
                 curriculum_params = yaml.safe_load(curriculum_info_yaml)
-            if outputs is not None and not args.multi_object:
-                assert outputs.size(0) == curriculum_params["num_dirs"]
 
         n_saved = 0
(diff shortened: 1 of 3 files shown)


lichtefeld (approved):
Looks fine to me. I didn't test this locally but I trust you've tested the fix.

PR #1194, opened 2022-10-06
Update segmentation server install to fix missing requirement

lichtefeld:


spigo900 (approved):
Looks good to me. Can confirm this fixes the problem I ran into yesterday.

Issue #70, opened 2019-09-04
Need better representation of relative size than biggerThan and smallerThan


rgabbard:
Since of course size can vary along a number of dimensions (e.g. by length? by volume? what if something is longer in one dimension but not another?)

rgabbard:
Not for M3.

LichMaster98:
It's worth noting that for Axes we introduced `MuchBiggerThan` and `MuchSmallerThan` which we could also apply to the relationships between objects _if_ we can justify the relationship. It might be reasonable to assert that a `house` is `MuchBiggerThan` a `ball` for example in normal cases.

PR #1198, opened 2022-10-17
Add new nodes to ADAM perception graph

boyleconnor:
I defined and added three new top-level nodes to the ADAM perception graph, `size`, `relative_size`, and `relative_distance`. I also changed one of the feature extraction steps of ADAM to output a `dict`/`object` for the `relative_size` feature instead of the `list`/`array` (which was harder to read as I originally coded it; you had to infer the relevant `object_name` instead of it being stored explicitly).

This PR is definitely not done yet, as it doesn't handle the `touching` feature, but I want to get eyes on this ASAP.

I also wanted to hear from @lichtefeld and/or @spigo900 whether there's anything downstream of this that I need to change to make sure that these new nodes in the perception graph are getting ingested properly.


adam/perception/perception_graph_nodes.py
```diff
@@ -98,6 +98,41 @@ def dot_label(self):
...
+@attrs(slots=True, frozen=True, eq=False)
+class RelativeSize(GraphNode):
+    x_bigger_than: float = attrib(validator=instance_of(bool))
+    y_bigger_than: float = attrib(validator=instance_of(bool))
+
+    def dot_label(self) -> str:
+        return (
+            f"RelativeSize(x_bigger_than={self.x_bigger_than},"
+            f" y_bigger_than={self.y_bigger_than})"
+        )
```
lichtefeld, line 120: Rather than a new node for `RelativeSize` we should consider re-using the size representation from Phase 1/2. Essentially that was just an edge for `BiggerThan(Object_X, Object_Y)` where it was asserted that Object_X was the larger object. If we reuse the other representation, we won't need to build new predicates (and given my understanding is the information contained is the same it should be fine? @spigo900 Do you agree?)

adam/perception/perception_graph_nodes.py
```diff
@@ -98,6 +98,41 @@ def dot_label(self):
         )
 
 
+@attrs(slots=True, frozen=True, eq=False)
+class Size(GraphNode):
+    x: float = attrib(validator=instance_of(float))
+    y: float = attrib(validator=instance_of(float))
+    area: float = attrib(validator=instance_of(float))
```
lichtefeld, line 105: Is X/Y roughly length, width? The latter may be more clear as variable names.

adam/perception/perception_graph_nodes.py
```diff
@@ -98,6 +98,41 @@ def dot_label(self):
...
+            f"RelativeSize(x_bigger_than={self.x_bigger_than},"
+            f" y_bigger_than={self.y_bigger_than})"
+        )
+
+
+@attrs(slots=True, frozen=True, eq=False)
+class RelativeDistance(GraphNode):
+    x_offset: float = attrib(validator=instance_of(float))
+    y_offset: float = attrib(validator=instance_of(float))
+    euclidean: float = attrib(validator=instance_of(float))
```
lichtefeld, line 127: Can you give me an example of what this data looks like?

lichtefeld (commented):
Each new `GraphNode` will need a matching predicates (defined in here: [perception_graph_predicates.py](https://github.com/isi-vista/adam/blob/master/adam/perception/perception_graph_predicates.py) and translation between the two (in a large `elif` chain) defined here: [perception_graph.py#L970](https://github.com/isi-vista/adam/blob/master/adam/perception/perception_graph.py#L970)

adam/perception/perception_graph_nodes.py
spigo900: width, height? I'd agree

adam/perception/perception_graph_nodes.py
spigo900: @lichtefeld Yes, I agree, I think this makes sense.

adam/perception/perception_graph_nodes.py
```diff
@@ -98,6 +98,41 @@ def dot_label(self):
...
+
+    def dot_label(self) -> str:
+        return (
+            f"RelativeSize(x_bigger_than={self.x_bigger_than},"
+            f" y_bigger_than={self.y_bigger_than})"
+        )
+
+
+@attrs(slots=True, frozen=True, eq=False)
+class RelativeDistance(GraphNode):
```
spigo900, line 124: A docstring might be useful here to explain what we are storing in this class.

spigo900 (commented):
This looks good. I agree with Jacob's comments. As far as things downstream of this changing, besides the need for predicates, the translation code in `_VisualPerceptionFrameTranslation` will have to change. It will need to translate the relative size relations (in the way Jacob mentioned) and it will need to add the relative distances and `Touching` information to the graph.

adam/perception/perception_graph_nodes.py
boyleconnor: The relative size features that we decided to extract in the pre-processing phase refer specifically to the relative widths and heights of the bounding boxes around each object. From what I'm reading, it doesn't look like the `biggerThan` and `smallerThan` edges support this (see [this issue](https://github.com/isi-vista/adam/issues/70))

spigo900: You're right, I forgot about that. In that case this representation makes sense to me.

boyleconnor: I had elected to represent the other relative feature-set (`relative_distance`) as a mapping of object IDs to nodes; should I switch to representing it with edges (instead of nodes) as well? I also realized we never use the absolute `distance` feature in the perception graph--it just gets generated by pre-processing and then never used by ADAM.

adam/perception/perception_graph_nodes.py
boyleconnor: Should I change this in pre-processing as well?

adam/perception/perception_graph_nodes.py
spigo900: The mapping setup is fine for this intermediate representation. In the graph relative distance should be a node plus a pair of edges, e.g. for a toy car on a table, something like `table --(in-region)--> relative_distance(x_offset: 0, y_offset=5, ...) --(reference-object)-->toy_car`. This is similar to how `Region` works. This is what i meant about adding the relative distance information to the graph.

adam/perception/perception_graph_nodes.py
spigo900: That's probably a good idea for consistency's sake.

adam/perception/perception_graph_nodes.py
lichtefeld: I'm fine with `relative_distance` being a node, it should have connections to the two referenced objects as edges in the graph then. I'm forgetting the exact edge name but its something like `reference object`? Our current `relative_distance` is just the shortest line between the two objects correct? We don't break it out in to the component X/Y axes do we? No is fine, but if we are separating on the per-axis basis then each of the `relative_distance` relationships should get separate nodes with a label indicating if its an `X` or `Y` axis relative distance.

With respect to the absolute `distance` value. It should probably be encoded so it can be presented as output but I don't see how an absolute value is useful for learning concepts, even if its important in understanding and analyzing the specific instance.

adam/perception/perception_graph_nodes.py
lichtefeld: No objections to changing in pre-processing but no strong preference on the rename there.

adam/perception/perception_graph_nodes.py
spigo900: @lichtefeld `reference object` sounds right. I believe we are calculating the axis distances as well as straight-line. I agree it is probably simplest to break it out by axis/label -- x or y -- especially since then you can just use a continuous node to represent it.

I agree we probably don't need the straight-line distance, no.

adam/perception/perception_graph_nodes.py
boyleconnor: If the centroid of object 1 is at `(1,2)` and the centroid of object 2 is at `(2, 6)` (and there are no other objects), then object 1 will have the following `relative_distance`:

```json
{
  object2: {
    x_offset: 1 // = 2 - 1
    y_offset: 4 // = 6 - 4
    euclidean: 4.1231 // = (1^2 + 4^2) ^ 0.5
  }
}
```

and object 2 will have the following `relative_distance`:

```json
{
  object1: {
    x_offset: -1 // = 1 - 2
    y_offset: 4 // = 4 - 6
    euclidean: 4.1231 // = (1^2 + 4^2) ^ 0.5
  }
}
```

adam/perception/visual_perception.py
```diff
@@ -165,25 +195,42 @@ def from_mapping(
...
+                    other_object_id = object_name_to_id(other_object_name)
+                    relative_properties[other_object_id] = []
+                    for label, value in relative_distance.items():
+                        relative_properties[other_object_id].append(ContinuousNode(label=label, value=value, weight=1.0))
+            if cluster_map["relative_size"] is not None:
+                for other_object_name, relative_size in cluster_map["relative_size"].items():
+                    other_object_id = object_name_to_id(other_object_name)
+                    for label, value in relative_size.items():
+                        # Have to convert value to `str` to fit CategoricalNode
+                        relative_properties[other_object_id].append(CategoricalNode(label=label, value=str(value), weight=1.0))
```
spigo900, line 210: The labels here will be `x_greater_than` and `y_greater_than`, correct?

adam/perception/visual_perception.py
```diff
@@ -165,25 +195,42 @@ def from_mapping(
                             weight=strokes_map["confidence_score"],
                         )
                     )
+            relative_properties = {}
+            if cluster_map["relative_distance"] is not None:
+                for other_object_name, relative_distance in cluster_map["relative_distance"].items():
+                    other_object_id = object_name_to_id(other_object_name)
+                    relative_properties[other_object_id] = []
+                    for label, value in relative_distance.items():
+                        relative_properties[other_object_id].append(ContinuousNode(label=label, value=value, weight=1.0))
```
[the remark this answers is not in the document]
boyleconnor: The labels here will be `euclidean_distance`, `x_offset`, and `y_offset`. Would you prefer that I put these into a named constant (similar to the [size features](https://github.com/isi-vista/adam/blob/907afb228ffdc6df58fd2d4f68db450f244b897d/adam/perception/visual_perception.py#L35))?

adam/perception/visual_perception.py
boyleconnor: The labels here will be `height_greater_than` and `width_greater_than`. Would you prefer that I put these into a named constant (similar to the [size features](https://github.com/isi-vista/adam/blob/907afb228ffdc6df58fd2d4f68db450f244b897d/adam/perception/visual_perception.py#L35))

adam/perception/visual_perception.py
```diff
@@ -58,6 +64,11 @@ class ClusterPerception:
         validator=deep_iterable(instance_of((GraphNode, OntologyNode))),
         converter=_to_immutableset,
     )
+    relative_properties: ImmutableDict[str, ImmutableSet[GraphNode]] = attrib(
+        validator=deep_mapping(
+            instance_of(str), deep_iterable(instance_of(GraphNode))),
+        converter=lambda x: _to_immutabledict({k: _to_immutableset(v) for (k, v) in x.items()}),
```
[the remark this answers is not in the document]
boyleconnor: I changed `_to_immutabledict()` to `immutabledict()` and `_to_immutableset()` to `immutableset()`; it did not cause a `TypeError`.

I had originally made used the `_to_type()` functions for consistency with the other properties' `converter` parameters (see, e.g. the [one for `strokes`](https://github.com/isi-vista/adam/blob/907afb228ffdc6df58fd2d4f68db450f244b897d/adam/perception/visual_perception.py#L52)). Is there a reason to prefer one over the other? If I change these, should I change the other ones as well?

adam/perception/visual_perception.py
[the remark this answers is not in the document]
spigo900: Yes, that might be helpful for readability.

adam/perception/visual_perception.py
spigo900: Yes, that might be helpful for readability.

boyleconnor:
I should mention that I'm getting a bunch of MyPy issues that I don't think are related to my code:

<details>
<summary>Non-relevant MyPy errors</summary>

```
adam/experiment/log_experiment.py:547: error: Argument "attribute_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:548: error: Argument "relation_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:549: error: Argument "action_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:550: error: Argument "functional_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[FunctionalLearner]"; expected "FunctionalLearner"
adam/experiment/log_experiment.py:553: error: Argument "generics_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[SimpleGenericsLearner]"; expected "SimpleGenericsLearner"
adam/experiment/log_experiment.py:556: error: Argument "plural_learner" to "SymbolicIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:606: error: Argument "attribute_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:607: error: Argument "relation_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:608: error: Argument "action_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:609: error: Argument "functional_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[FunctionalLearner]"; expected "FunctionalLearner"
adam/experiment/log_experiment.py:612: error: Argument "generics_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[SimpleGenericsLearner]"; expected "SimpleGenericsLearner"
adam/experiment/log_experiment.py:615: error: Argument "plural_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:617: error: Argument "affordance_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[TemplateLearner]"; expected "TemplateLearner"
adam/experiment/log_experiment.py:618: error: Argument "mapping_affordance_learner" to "SimulatedIntegratedTemplateLearner" has incompatible type "Optional[MappingAffordanceLearner]"; expected "MappingAffordanceLearner"
```

</details>

Not sure what to do about these

spigo900:
@boyleconnor Do you get those MyPy errors when running `make mypy`? I'm not able to reproduce with that command locally.

spigo900:
Aha, it happens when I upgrade to `0.982`. `0.942` (my original version) and `0.950` (locked) are too old to get the error. It looks like those `log_experiment` warnings are spurious, because the arguments it's complaining about should have optional type and are being assigned from an optional type, but I'm not sure why they're happening.

lichtefeld:
@boyleconnor Just `#type: ignore` them. mypy is improved beyond some of ADAM's original code base and sometimes our usage of inheritance still causes mypy issues. Given the time sink it would probably be to correctly update everything across the library its not worth it.

adam/perception/visual_perception.py
```diff
@@ -165,25 +212,52 @@ def from_mapping(
                             weight=strokes_map["confidence_score"],
                         )
                     )
+            relative_properties: Dict[str, List[GraphNode]] = {}
+            if cluster_map["relative_distance"] is not None:
+                for other_object_name, relative_distance in cluster_map[
+                    "relative_distance"
+                ].items():
+                    other_object_id = object_name_to_id(other_object_name)
+                    relative_properties[other_object_id] = []
```
spigo900, line 221: Minor: I would like to work on the readability here. In the current flow we have to assume in the `relative_size` code that every object has been assigned a list value in the dict already. This makes sense if I think about it, but I'd rather not have to.

One solution might be to separate out this initialization step. Another might be to use `dict.setdefault()` in both loops.

spigo900 (approved):
Looks good to me, though one minor readability issue.

adam/perception/visual_perception.py
boyleconnor: @spigo900 I fixed it by switching `relative_properties` to a `defaultdict(list)`. Hope that works; I'm gonna try to merge today

spigo900: `defaultdict` also works. Probably a good idea to convert it back when storing it in the `ClusterPerception` so that typos downstream of this code don't get hidden.

spigo900: Looks good to me. Thank you.

PR #1206, opened 2022-11-04
Stop merging objects

### LINKED ISSUE #1205: Prevent merging of touching objects
Because of how our stroke extraction process works (by edge detection), touching objects in an image are currently ~guaranteed to have their strokes mingle and get treated as one mega-object. That means we cannot successfully recognize touching relationships (a specific type of the spatial relation features we are currently investigating) without addressing this problem.

To fix that, one simple hack is to separate out each unique mask in the segmentation image and treat it as its own input to the stroke extraction code (feed into Matlab and process results). We then concatenate the outputs for each such unique mask. This ensures that objects that are distinct in the segmentation image result in distinct stroke extraction objects. It also runs into some problems, namely:

1. STEGO has a bunch of weird artifacts that will also get recognized as objects. So, for example, we'll end up with an object for "that triangle-shaped hallucination in the corner of the background" and "the shadow of the ball we're looking at." This is probably OK for the spatial relations experiment, but may have weird effects for other things.
2. Any time one object's mask surrounds another's, we are going to end up with two copies of that object's strokes, so probably two copies of that object, which is probably going to have weird results. At the moment the main problem is the background, which we may be able to hack out in some cases (via #1204), although this hack relies heavily on the special case no-background-noise nature of our curricula.
3. Probably something else.

This should be a flag because such behavior is not always appropriate. It's not appropriate when training the objects GNN with e.g. STEGO, which may produce multiple masks for just one object. It is probably counterproductive to train the GNN to classify "each weird shape STEGO recognized in the surface of a ball" as being "ball". For similar reasons it's also counterproductive to separate on each mask when using the color-refined segmentation, *unless* we retain the original instance segmentation and use that for separating out parts of the combined mask image. But that's complicated.

boyleconnor:
Fixes #1205.

Having just requested review, I realize I need to duplicate these changes in `adam_preprocessing/single_file_extraction.py`. That said, I think it is far enough along in `shape_stroke_extraction.py` to make review worth it

### DIFF
--- a/adam_preprocessing/shape_stroke_extraction.py
+++ b/adam_preprocessing/shape_stroke_extraction.py
@@ -508,7 +508,8 @@ class Stroke_Extraction:
         rgb_img_path: str,
         stroke_img_save_path: str,
         stroke_graph_img_save_path: str,
-        features_save_path: str,
+        output_dir: str,
+        process_masks_independently: bool = False,
         should_merge_small_strokes: bool = False,
         debug_vis: bool = False,
         debug_matlab_stroke_img_save_path: str,
@@ -529,7 +530,9 @@ class Stroke_Extraction:
         self.debug_matlab_stroke_img_save_path = debug_matlab_stroke_img_save_path
         self.stroke_img_save_path = stroke_img_save_path
         self.stroke_graph_img_save_path = stroke_graph_img_save_path
-        self.features_save_path = features_save_path
+        self.output_dir = output_dir
+        self.features_save_path = os.path.join(output_dir, "feature.yaml")
+        self.process_masks_independently = process_masks_independently
 
     @staticmethod
     def from_m5_objects_curriculum_base_path(
@@ -765,6 +768,24 @@ class Stroke_Extraction:
         # objects for those locations are (considered) distinct. I think that is the convention
         # here.
         img_seg = cv2.imread(self.path)
+        all_seg_pixels = np.reshape(img_seg, (-1, img_seg.shape[-1]))
+        all_unique_colors = np.unique(all_seg_pixels, axis=0)
+
+        # Ignore black (background) pixels
+        unique_colors = all_unique_colors[
+            np.where((all_unique_colors != [0, 0, 0]).any(axis=-1))
+        ]
+        # Array of boolean masks (one for each color), shape: (n_masks, height, width)
+        masks = np.all(
+            img_seg[np.newaxis, ...] ==
+            unique_colors[:, np.newaxis, np.newaxis, :], axis=-1
+        )
+        for i in range(unique_colors.shape[0]):
+            color = unique_colors[i, np.newaxis]
+            int_mask = masks[i, :, :, np.newaxis].astype(int)
+            color_mask = int_mask * color
+            cv2.imwrite(os.path.join(self.output_dir, f"{os.path.basename(self.path)}_mask_{i}.png"), color_mask)
+
         img_seg = cv2.cvtColor(img_seg, cv2.COLOR_BGR2GRAY)
         # jac: obj_area is a tuple of two equal-length arrays xs, ys such that for all i,
         # img_seg[xs[i], ys[i]] != 0.
@@ -1001,6 +1022,11 @@ def main():
         dest="tolerate_errors",
         help="If passed, will halt immediately when extraction fails for a given situation."
     )
+    parser.add_argument(
+        "--process-masks-independently",
+        action="store_true",
+        help="If passed, will not merge adjacent masks of different colors."
+    )
     args = parser.parse_args()
 
     logging.basicConfig(level=logging.INFO)
@@ -1050,14 +1076,15 @@ def main():
                 debug_matlab_stroke_img_save_path=str(output_situation_dir / "matlab_stroke_0.png"),
                 stroke_img_save_path=str(output_situation_dir / "stroke_0.png"),
                 stroke_graph_img_save_path=str(output_situation_dir / "stroke_graph_0.png"),
-                features_save_path=str(output_situation_dir / "feature.yaml"),
+                output_dir=output_situation_dir,
                 should_merge_small_strokes=args.merge_small_strokes,
-                obj_type=string_label,
+                obj_type="dummy",
                 # TODO create issue: in the reorganized curriculum format, we don't store the
                 #  information about the object view/ID in an authoritative global-per-sample way such
                 #  that we could retrieve it here in the stroke extraction code.
                 obj_view="-1",
                 obj_id="-1",
+                process_masks_independently=args.process_masks_independently,
             )
             S.get_strokes()
         except ValueError as e:
--- a/adam_preprocessing/shape_stroke_extraction.py
+++ b/adam_preprocessing/shape_stroke_extraction.py
@@ -10,7 +10,7 @@ import logging
 import cv2
 import numpy as np
 from pathlib import Path
-from typing import List, Optional, Sequence, Tuple
+from typing import List, Optional, Sequence, Tuple, NamedTuple
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +42,19 @@ from utils import label_from_object_language_tuple
 
 
 TOUCHING_DISTANCE_THRESHOLD = 50.0
+
+
+class ExtractionResult(NamedTuple):
+    """
+    Stroke extraction results for one segmentation file.
+
+    File may or may not contain more than 1 mask.
+    """
+    num_obj: int
+    reduced_strokes: np.ndarray
+    stroke_obj_ids: np.ndarray
+    adj: np.ndarray
+    colors: np.ndarray
 
 
 def vectorized_bspline_coeff(vi, vs):
@@ -516,6 +529,7 @@ class Stroke_Extraction:
         vis: bool = True,
         save_output: bool = True,
     ):
+        self.segmentation_paths: List[str] = []
         self.obj_view = obj_view
         self.obj_type = obj_type
         self.obj_id = obj_id
@@ -579,7 +593,7 @@ class Stroke_Extraction:
             ),
         )
 
-    def stroke_extraction_from_matlab(self):
+    def stroke_extraction_from_matlab(self, segmentation_path: str):
         """
         Get the raw stroke extraction results from Matlab using the BPL code.
 
@@ -594,11 +608,11 @@ class Stroke_Extraction:
         eng = matlab.engine.start_matlab()
         file_dir = os.path.dirname(__file__)
         eng.addpath(eng.genpath(file_dir), nargout=0)
-        out = eng.ske(self.path, nargout=2)
+        out = eng.ske(segmentation_path, nargout=2)
         eng.close()
         return out
 
-    def remove_strokes(self):
+    def remove_strokes(self, segmentation_path: Path) -> Tuple[int, np.ndarray, np.ndarray, np.ndarray]:
         """
         Process Matlab stroke extraction results into labels.
 
@@ -610,14 +624,14 @@ class Stroke_Extraction:
         - determine the number of objects in the scene as the number of components
         - label each stroke according to which object it is a part of
 
-        The outputs are stored in:
+        The outputs are returned as:
 
-        - self.num_obj
-        - self.strokes
-        - self.adj
-        - self.labels
+        - num_objs
+        - strokes
+        - stroke_obj_ids
+        - adj
         """
-        out_s, out_e = self.stroke_extraction_from_matlab()
+        out_s, out_e = self.stroke_extraction_from_matlab(str(segmentation_path))
         strokes = []
         removed_ind = []
         adj = kp2stroke(np.array(out_e))
@@ -657,7 +671,7 @@ class Stroke_Extraction:
                         adj[i, j] = 1.0
                         adj[j, i] = 1.0
         # jac: hack to get things running
-        self.num_obj = num_obj = min(n_components, len(adj))
+        num_obj = min(n_components, len(adj))
         # If we have at least two strokes and two objects, see if we can break up the connected components further to
         # get more fine-grained objects.
         #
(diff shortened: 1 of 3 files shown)


adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -738,6 +741,24 @@ def get_colors(self):
...
+        ]
+        # Array of boolean masks (one for each color), shape: (n_masks, height, width)
+        masks = np.all(
+            img_seg[np.newaxis, ...] ==
+            unique_colors[:, np.newaxis, np.newaxis, :], axis=-1
+        )
+        for i in range(unique_colors.shape[0]):
+            color = unique_colors[i, np.newaxis]
+            int_mask = masks[i, :, :, np.newaxis].astype(int)
+            color_mask = int_mask * color
```
spigo900, line 759: Could do `img_seg[masks[i, :, :, np.newaxis]]`? Does that work?

scripts/generate_curriculum_from_images.py
```diff
@@ -0,0 +1,55 @@
...
+        description = image_file.name[:-len('.jpg')]
+
+        # Make situation directory
+        situation_dir = args.output_dir / f"situation_{situations_count}"
+        situation_dir.mkdir()
+        situations_count += 1
+
+        # Save description
+        with open(situation_dir / "description.yaml", "w") as description_file:
+            dump({'language': description}, description_file, Dumper=Dumper)
```
spigo900, line 40: Why do we need to specify a `Dumper` here?

scripts/generate_curriculum_from_images.py
```diff
@@ -0,0 +1,55 @@
...
+        "--output_dir", type=Path, default="data/multi_obj")
+    args = parser.parse_args()
+
+    args.output_dir.mkdir(parents=True)
+
+    situations_count = 0
+    images_dir: Path = args.images_dir
+    for image_file in images_dir.glob('*.jpg'):
+        input_image = Image.open(image_file)
+        description = image_file.name[:-len('.jpg')]
```
spigo900, line 31: You can use `image_file.stem` here (if I recall the attribute correctly) to get the path's base name without the extension/suffix.

scripts/generate_curriculum_from_images.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        "Create new curriculum from directory full of objects. Take language "
+        "descriptions from filenames. "
+    )
+    parser.add_argument("images_dir", type=Path)
+    parser.add_argument(
+        "--output_dir", type=Path, default="data/multi_obj")
```
spigo900, line 22: This is fine for testing, but note that when you create the actual curricula for your experiment they must go under `data/curriculum/train` and `data/curriculum/test` respectively.

spigo900 (commented):
What you have so far looks good. Just a few nits.

boyleconnor:
As a check-in (just for the mask separation), how does this look:

Original RGB:
rgb_0

STEGO output (not fixed):
semantic_0

2 of the masks (there are 5 total, 2 are shadows and 1 is the background)
semantic_0 png_mask_2

semantic_0 png_mask_4?

spigo900:
@boyleconnor Those images look good to me. Re: names, now that I think of it we probably want to use [`os.path.splitext()`](https://docs.python.org/3.7/library/os.path.html#os.path.splitext) to remove the .png before adding `_mask_{i}.png` but that's minor.

adam_preprocessing/shape_stroke_extraction.py
spigo900: Huh, that's annoying. I swear there should be an indexing way to do this but it's not too worth fixing. I'm okay with this as-is.

Not necessary to change, but I wanted to think about how to do it -- here's an idea. I think we don't need `masks` at all. We can do `color_mask = np.where(img_seg == unique_colors[i], img_seg, [0, 0, 0])`. I tested this in a Python console and it seems to work with a two-mask test image.

boyleconnor: Yep, this works!

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -567,7 +567,18 @@ def stroke_extraction_from_matlab(self):
...
+            for segmentation_filename in self.distinct_mask_filenames:
+                path = os.path.join(self.output_dir, segmentation_filename)
+                s: List[matlab.double]
+                e: matlab.double
+                s, e = eng.ske(path, nargout=2)
+                s_out.extend(s)
+                e_out.append(e)
+            out = (s_out, eng.cat(1, *e_out))
+        else:
+            out = eng.ske(self.path, nargout=2)
```
boyleconnor, line 581: @spigo900 is this roughly what I should be doing here? Extracting all the strokes separately from each masks and then putting them all together into one big array?

Also, it's weird to me that one of the outputs of `ske()` is a `list[matlab.double]` while the other output is just a `matlab.double` (I probably just don't know enough about MATLAB/its Python API to see why this makes sense)

spigo900: No, this won't do what we want. The stroke graphs will start separate but get glued together during the post-processing. The post-processing step adds edges to the stroke graph between strokes that are sufficiently close to each other.

What we want to do is to run, for each image, the whole pipeline up until the part where we convert to YAML. This produces some outputs that get fed into the "convert to YAML" step. I think we want to concatenate those outputs and then feed them to "convert to YAML." I think this should be relatively straightforward though a bit annoying to do. I'm thinking you'll want to actually concatenate them since that keeps the code the same. Some thoughts on concatenating the five implicit inputs to "convert to YAML," that is `num_obj`, `stroke_obj_ids`, `reduced_stroke`, `adj`, and `color`:

1. `num_obj` you can just sum, of course.
2. `stroke_obj_ids` is a NumPy array and needs special treatment. Before concatenating `prev_stroke_obj_ids` and `new_stroke_obj_ids` you'll want to increment `new_stroke_obj_ids` by `prev_num_objs` (aka `np.max(prev_stroke_obj_ids) + 1`?) to avoid accidentally gluing together different objects. I expect stroke object IDs to start at 0 and increase to `num_objs - 1`, with each mask's stroke object IDs starting at 0 by default such that you have to manually increment them before concatenating.
3. `reduced_strokes` you can just straight up concatenate, I think. It's a NumPy array.
4. You'll probably want to concatenate the adjacency matrices along the diagonal. That way you don't have to change the existing logic. I am not sure the easiest/most efficient way to do this in NumPy. Worst case you can allocate a new adjacency matrix using `cat_adj = np.zeros()` and use `cat_adj[:self.num_objs, :][:, :self.num_objs] = old_adj; cat_adj[:self.num_objs, :][:, :self.num_objs] = new_adj`. Lots of allocations and uses lots of memory but it should work OK for us.
5. `colors` is 2D of shape `[num_obj, 5]`. I think you can just concatenate on the first axis.

I don't understand the MATLAB API that well myself, so not sure why it's a list of MATLAB doubles.

boyleconnor:
I'm getting **a lot** of this warning: `UserWarning: Graph is not fully connected, spectral embedding may not work as expected.` I was worried this was due to the way I concatenate the stroke adjacency matrix, which of course leaves the strokes unconnected to one another. But it turns out the only call to `SpectralClustering.fit()` is before the stroke adjacency matrices are concatenated together (i.e. during the extract-from-each-file step). I'm not sure if this is something to worry about, but the stroke extraction doesn't look very good and I'm looking into why

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -673,61 +689,68 @@ def strokes2controlpoints(self):
...
+                s_c = []
+                for i in range(len(reduced_obj)):
+                    p = plt.plot(reduced_obj[i][:, 1], reduced_obj[i][:, 0], linewidth=3)
+                    s_c.append(p[0].get_color())
+                    plt.scatter(reduced_obj[i][:, 1], reduced_obj[i][:, 0], c="r")
+                stroke_colors.append(s_c)
+                plt.gca().invert_yaxis()
+                plt.gca().set_aspect("equal", adjustable="box")
+                plt.axis("off")
+            stroke_colors_list.append(stroke_colors)
```
boyleconnor, line 721: I think there is something going wrong with this; the stroke output image (`stroke_0.png`) looks right when there are just a handful of masks per file, but it looks totally wrong when there are lots of masks in an image. I don't know what a lot of this pyplot stuff is doing but I'm wondering if some of these methods should be called only once. Also not sure how much effort this is worth fixing.

boyleconnor: Here's an example of the bad stroke plotting:

<details><summary>rgb_0.png</summary>

rgb_0

</details>

<details><summary>semantic_0.png</summary>

semantic_0

</details>

<details><summary>stroke_0.png</summary>

stroke_0

</details>

<details><summary>stroke_graph_0.png</summary>

stroke_graph_0

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -927,6 +979,38 @@ def get_strokes(self):
         # FIXME: This return value is never used
         return objects
 
+    def combine_extractions(self, extractions: Sequence[ExtractionResult]):
+        num_obj = sum(e.num_obj for e in extractions)
+        prev_num_strokes = 0
+        prev_num_obj = 0
```
spigo900, line 985: `prev` sounds like it should be equal to the previous value, rather than the sum of all previous values. Maybe we want something like `num_strokes_seen` and `num_objs_seen`?

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -996,16 +1085,16 @@ def main():
...
                 situation_description = yaml.safe_load(situation_description_file)
             language_tuple = tuple(situation_description["language"].split(" "))
-            _integer_label, string_label = label_from_object_language_tuple(
-                language_tuple, situation_num
-            )
+            # _integer_label, string_label = label_from_object_language_tuple(
+            #     language_tuple, situation_num
+            # )
         else:
             string_label = "dummy"
```
spigo900, line 1099: Might as well delete this dead code.

adam_preprocessing/singlefile_stroke_extraction.py
```diff
@@ -32,6 +32,11 @@ def main():
         default="dummy",
         help="String label to use."
     )
+    parser.add_argument(
+        "--process-masks-independently",
+        action="store_true",
+        help="If passed, will not merge adjacent masks of different colors."
+    )
```
spigo900, line 39: This arg is never passed into the stroke extraction code?

Incidentally, thanks for updating this file.

spigo900 (changes requested):
Processing looks good to me. Some notes about fixing the plotting. The rest are minor comments.

spigo900:
Re: `Spectral`, I don't think it affects the stroke outputs. I've seen it a few times on the M5 objects curriculum as well. As far as I can tell the spectral clustering is only used as a way of regrouping strokes into objects, a different way from using the connected components of the graph. I think it's downstream of the bad strokes.

I expect the bad strokes are mostly a function of STEGO just picking up lots of spurious distinct object masks when run on real images. Turning stroke merging on might improve the stroke outputs a bit by fixing some of the disconnects and removing some of the bare patches. I expect it would at least decrease the number of warnings -- it looks like there are lots of disconnected strokes in the visualization, like you might expect given the number of masks.

adam_preprocessing/shape_stroke_extraction.py
boyleconnor: I've worked on this and here is the output for `situation_1`

<details><summary>rgb_0.png</summary>

rgb_0

</details>

<details><summary>semantic_0.png</summary>

semantic_0

</details>

<details><summary>stroke_0.png</summary>

stroke_0

</details>

<details><summary>stroke_graph_0.png</summary>

stroke_graph_0

</details>

These look right to me (minus the fact that the original extracted strokes are bad due to the background not being black). How does it look to you?

spigo900: Much better! Yes, this looks right to me.

adam_preprocessing/shape_stroke_extraction.py
spigo900: To be clear, I think you can delete this whole `if`/`else` -- we don't actually need the string label so there's no point in loading it.

adam_preprocessing/shape_stroke_extraction.py
```diff
@@ -564,11 +581,11 @@ def stroke_extraction_from_matlab(self):
         eng = matlab.engine.start_matlab()
         file_dir = os.path.dirname(__file__)
         eng.addpath(eng.genpath(file_dir), nargout=0)
-        out = eng.ske(self.path, nargout=2)
+        out = eng.ske(segmentation_path, nargout=2)
         eng.close()
         return out
 
-    def remove_strokes(self):
+    def remove_strokes(self, segmentation_path: str) -> Tuple[int, np.ndarray, np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
spigo900: Thanks!

spigo900 (changes requested):
Looks good to me. There's just the `string_label` `if`/`else` to delete.

boyleconnor:
@spigo900 I think that piece of code is useful for extracting the `concept_name` (I believe it's used for GNN training) from the `description.yaml`. What do you think with it now?

spigo900:
@boyleconnor No, the GNN training repeats this process to extract a description of each situation. It shouldn't use the `concept_name` because that's where the GNN decode script stores the GNN output. If it looked for the label there, it would be easy to accidentally train the GNN on a previous GNN's labels rather than the ground truth. We actually want to use a dummy concept name here.

spigo900:
Specifically, the GNN code uses `get_stroke_data()` which includes [this equivalent block of code](https://github.com/isi-vista/adam/blob/a20846da8a552d8066eee08002c0da4164b1b673/adam_preprocessing/utils.py#L86-L97). [This](https://github.com/isi-vista/adam/blob/a20846da8a552d8066eee08002c0da4164b1b673/adam_preprocessing/utils.py#L168-L172) is the tuple->label conversion code.

boyleconnor:
@spigo900 there was a lot of extraneous code I ended up deleting in [9d97a40](https://github.com/isi-vista/adam/pull/1206/commits/9d97a40bcd04de4ffe5516c0fc982870c5d01aa6). Let me know if that's alright

spigo900:
@boyleconnor Looks okay to me.

spigo900 (approved):
Looks good to me.

PR #1209, opened 2022-11-14
Fix crash when learner hypothesis and perception have continuous nodes

spigo900:
The crash happened because we were actually updating the new hypothesis created from the current perception, rather than updating the old hypothesis.


spigo900:
Of course once the tests are passing one still has to remember to make `black` happy with the formatting. 🙃 Oh well. Fixed that and cleaned up the history. CI should pass soon enough I hope.

lichtefeld (approved):
The joys of CI :) -- These changes look good I'll approve so you can merge when CI completes.

PR #1213, opened 2022-11-15
Calculate situation-level GNN test accuracy

### LINKED ISSUE #1212: Calculate GNN's situation-level test accuracy
To make it easier to relate ADAM accuracy to GNN accuracy it would be nice to calculate the GNN's situation-level test accuracy as opposed to only calculating the object-level test accuracy. It would be good to have those numbers for the report, especially to explain the one weird color segmentation result where ADAM test accuracy is higher than the GNN's object-level test accuracy (but lower than the GNN's situation-level test accuracy).

spigo900:
Closes #1212.

### DIFF
--- a/adam_preprocessing/shape_stroke_graph_learner.py
+++ b/adam_preprocessing/shape_stroke_graph_learner.py
@@ -164,7 +164,7 @@ def main():
             )[0]
         )
         situation_test_acc = get_situation_accuracy(
-            nn.LogSoftmax(dim=1)(
+            (
                 model(
                     test_adjacency_matrices,
                     test_node_features,
@@ -200,7 +200,7 @@ def main():
         )[0]
     )
     situation_test_acc = get_situation_accuracy(
-        nn.LogSoftmax(dim=1)(
+        (
             model(
                 test_adjacency_matrices,
                 test_node_features,
--- a/adam_preprocessing/shape_stroke_graph_learner.py
+++ b/adam_preprocessing/shape_stroke_graph_learner.py
@@ -184,6 +184,33 @@ def main():
             )
         )
     logging.info("Best test acc is {}".format(best_acc))
+    test_acc = Variable(
+        evaluation(
+            (
+                nn.LogSoftmax(dim=1)(
+                    model(
+                        test_adjacency_matrices,
+                        test_node_features,
+                        test_edge_features,
+                    )
+                )
+            ).data,
+            test_label.data,
+            topk=(1,),
+        )[0]
+    )
+    situation_test_acc = get_situation_accuracy(
+        nn.LogSoftmax(dim=1)(
+            model(
+                test_adjacency_matrices,
+                test_node_features,
+                test_edge_features,
+            )
+        ).data,
+        test_label.data,
+        test_situation_number_to_object_indices,
+    )
+    logging.info("Final test acc is {}; situation-level is {}".format(test_acc, situation_test_acc))
     logging.info(f"Saving model state dict to {args.save_model_to}")
     torch.save(model.state_dict(), args.save_model_to)
     logging.info("Model saved.")
--- a/adam_preprocessing/shape_stroke_graph_learner.py
+++ b/adam_preprocessing/shape_stroke_graph_learner.py
@@ -18,7 +18,9 @@ import matplotlib.pyplot as plt
 import glob
 import pickle as pickle
 from MPNN import MPNN, MPNN_Linear
-from utils import accuracy, get_stroke_data, LinearModel, load_data, STRING_OBJECT_LABELS
+from utils import (
+    accuracy, get_situation_accuracy, get_stroke_data, LinearModel, load_data, STRING_OBJECT_LABELS
+)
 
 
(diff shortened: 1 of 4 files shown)


adam_preprocessing/utils.py
```diff
@@ -231,6 +231,52 @@ def accuracy(output, target, topk=(1,)):
...
+    # jac: Shouldn't be necessary, given we already forced pred to be .type_as(target), but it was
+    # in Sheng's accuracy code and I didn't care to experiment and find out.
+    target = target.type_as(pred)
+
+    n_situations = 0
+    n_correct = 0
+    for situation, object_indices in situation_number_to_object_indices.items():
+        n_situations += 1
+        if object_indices:
+            if max(object_indices) < pred.shape[0]:
```
boyleconnor, line 266: I'm not sure why this case is singled out as one to watch out for and provide a clear error message for

adam_preprocessing/utils.py
```diff
@@ -231,6 +231,52 @@ def accuracy(output, target, topk=(1,)):
     return res
 
 
+def get_situation_accuracy(
+    output: torch.Tensor,
+    target: torch.Tensor,
+    situation_number_to_object_indices: Mapping[int, Sequence[int]],
+) -> float:
+    """
+    Computes the situation-wise acurracy@k for the specified values of k.
```
boyleconnor, line 240: This docstring makes it sound like the function returns the accuracy of a particular situation, i.e. `situation_k`. I would not mention the variable `k` in this docstring and instead just say that the function calculates situation-wise accuracy across all situations

adam_preprocessing/utils.py
```diff
@@ -231,6 +231,52 @@ def accuracy(output, target, topk=(1,)):
...
+
+    n_situations = 0
+    n_correct = 0
+    for situation, object_indices in situation_number_to_object_indices.items():
+        n_situations += 1
+        if object_indices:
+            if max(object_indices) < pred.shape[0]:
+                situation_preds = pred[object_indices]
+                n_correct += torch.any(
+                    situation_preds == target[object_indices].expand_as(situation_preds)
```
boyleconnor, line 269: Do I understand correctly that this considers an entire situation "correctly predicted" if any object in the situation is predicted with the correct label? If so, I think that should be stated in the docstring.

adam_preprocessing/shape_stroke_graph_learner.py
```diff
@@ -161,16 +163,54 @@ def main():
                 topk=(1,),
             )[0]
         )
+        situation_test_acc = get_situation_accuracy(
+            nn.LogSoftmax(dim=1)(
```
boyleconnor, line 167: I don't think there's any reason to take the softmax before inputting to `get_situation_accuracy()`. It should work correctly directly on the logits

adam_preprocessing/shape_stroke_graph_learner.py
```diff
@@ -161,16 +163,54 @@ def main():
...
+                        test_edge_features,
+                    )
+                )
+            ).data,
+            test_label.data,
+            topk=(1,),
+        )[0]
+    )
+    situation_test_acc = get_situation_accuracy(
+        nn.LogSoftmax(dim=1)(
```
boyleconnor, line 203: Same as my above comment on softmax

boyleconnor (changes requested):

adam_preprocessing/shape_stroke_graph_learner.py
spigo900: Fair enough.

adam_preprocessing/utils.py
spigo900: Correct.

adam_preprocessing/utils.py
spigo900: My bad, that came from the original accuracy function's docstring. I'll remove it.

adam_preprocessing/utils.py
spigo900: I suppose it's not strictly necessary here.

adam_preprocessing/shape_stroke_graph_learner.py
spigo900: Removed the softmax.

adam_preprocessing/shape_stroke_graph_learner.py
spigo900: Removed the softmax.

adam_preprocessing/utils.py
spigo900: Added this to the docstring.

adam_preprocessing/utils.py
spigo900: Removed the `@k` part from the docstring.

adam_preprocessing/utils.py
spigo900: Removed the special error message.

lichtefeld (approved):
looks good to me

boyleconnor (approved):

PR #1216, opened 2022-11-18
Fix bug where color refining doesn't change segmentation

### LINKED ISSUE #1215: Color-refinement script doesn't refine segmentation at all
This is a very silly bug... it's a case of passing the arguments in reverse order. Oops. At least it's an easy fix.

spigo900:
Closes #1215.


lichtefeld (approved):
Looks good. Silly bug :)

PR #1224, opened 2022-11-28
Calibrate touching threshold

boyleconnor:


boyleconnor:
Here is the current multi-object curriculum (incl. black backgrounds)

[multi_obj_curriculum.zip](https://github.com/isi-vista/adam/files/10106311/multi_obj_curriculum.zip)

boyleconnor:
It's still not very scientific but I have tried playing with a few examples and `5.0` seems like a much better touching threshold than `50.0`.

More importantly, I made a significant bugfix. The mask extraction code was extracting anywhere there was an exact match along any of the `(r, g, b)` values; instead it should be (and I think now is) extracting based on exact matches across all of the `(r, g, b)` values. I have also added a minimum threshold of 20 pixels, below which a mask will be dropped. This helps remove a lot of trivially small masks.

spigo900 (approved):
Looks good to me.

PR #1226, opened 2022-11-29
Account for no objects detected in stroke extraction code

lichtefeld:
Account for no objects detected in the stroke extraction code so we don't make an image of size 0.


spigo900 (approved):
lgtm

PR #1227, opened 2022-11-29
Prevent crash when attempting to combine empty stroke extraction result

spigo900:
Fixes another crash along the lines of #1226.


lichtefeld (approved):
🔨🐛

